PR #46, opened 2016-02-21
Lw/json convertible

LoganWright:


LoganWright:
@tannernelson I think this is good to go, let me know how you feel about the way data is architected and I'll make a quick update to the README.

Basic access should be:

```
let name = request.data["name"]?.string
let age = request.data["age"]?.int
// ... etc
```

PR #68, opened 2016-02-25
Events and Subscribers

mpclarkson:
This PR proposes a similar event-subscriber system as Laravel. This will allow users to 'hook' into system events, such as errors, or `Fluent` model events, such as `pre-save`, `pre-remove`, as well as allowing users to create their own events.

Users would register subscribers when the application boots that respond to specific events. Note that the `targets` do not subscribe to events and should not need to be deregistered. It is the job of each subscriber to check if it should respond to a particular event. Subscribers are able to define a `priority` that determines the order in which they are processed.

An example is below.

```swift

//A user model
struct User {
    let name: String
}

//Example events fired by Fluent
enum FluentEvents: String, EventType {
    case PreSave, PostSave, PreUpdate, PostUpdate, PreDestroy, PostDestroy...
}

//User's custom defined 
enum MyEvents: String, EventType {
    case CustomEvent...
}

//We create subscribers

let high = Subscriber(FluentEvents.PreSave, priority: 100) { data in    
    guard let user = data as? User else { return }
    print("High Priority - \(user.name) just registered. Do something")
}

let low = Subscriber(FluentEvents.PreSave) { data in
    guard let user = data as? User else { return }
    print("Standard Priority - \(user.name) registered. Stuff has already been done.")
}

let other = Subscriber(MyEvents.CustomEvent) { data in
    print("Custom event - \(data)")
}

//We add the subscribers in main.swift
Subscriber.add([low, high, other])

//Fluent fires an event
let event = Event(FluentEvents.PreSave data: user)
Event.fire(event)

//Subscribers 'high' and 'low' respond to this specific event (in that order).
// The custom subscriber does not respond.

```

### DIFF
--- a/Sources/Event.swift
+++ b/Sources/Event.swift
@@ -35,8 +35,10 @@ public class Event<T, U where T: EventType, U: Any> {
             }
             .sort { $0.priority > $1.priority }
         
-        subscribed.forEach { subscriber in
-            subscriber.handle(event.data)
+        for subscriber in subscribed {
+            if !subscriber.handle(event.data) {
+                break
+            }
         }
     }
     
@@ -46,7 +48,7 @@ public class Event<T, U where T: EventType, U: Any> {
     /* Initializer
      
      - parameter type: The EventType used to identify susbcribers that should respond to this event
-     - parameter: data: Any - the data passed to subscribers
+     - parameter: data: Any - the data passed to subscribers to respond to
      */
     public required init(_ type: T, data: U) {
         self.type = type
@@ -56,13 +58,14 @@ public class Event<T, U where T: EventType, U: Any> {
 
 //MARK: Subscriber
 
-public typealias EventHandler = (Any) -> ()
+public typealias Continue = Bool
+public typealias EventHandler = (Any) -> (Continue)
 
 /*
  Subcribers must implement this protocol
  */
 public protocol Subscribable {
-    func handle(data: Any)
+    func handle(data: Any) -> Continue
     var priority: Int { get }
     var type: EventType { get }
 }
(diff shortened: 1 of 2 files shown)


mpclarkson:
There is more we can do with this, e.g. allowing subscribers to stop event propagation.

tannernelson:
Looks cool. Is there something similar to this in Laravel or other frameworks?

mpclarkson:
Yes, most as far as I am aware, some examples: 

- Django 'signals' https://docs.djangoproject.com/es/1.9/topics/signals/
- Laravel 'events': https://laravel.com/docs/5.1/events
- Symfony 'events': http://symfony.com/doc/current/components/event_dispatcher/introduction.html

Matthew Clarkson

tannernelson:
Awesome. If you can include some tests and documentation markdown we can merge this.

mpclarkson:
Cool, I’m going to refactor a few things too. 

The main thing is I want to move the subscriber registration to the `Application` class so the initialisation is consistent  with `providers` and `middleware`. I also want to set upper and lower limits to the priority and include some debug logging.

mpclarkson:
Moving this to a new branch as I have Xcode project conflicts.

tannernelson:
Did you see the message in #release in Vapor Slack? You might want to reclone Vapor and add the files back in. Folder structure changed.

mpclarkson:
Yeah I did. It's a good change and is what obviously caused the conflicts. I've opened a new branch on my fork with the files. I've moved the `Subscriber` to a protocol (no class) and have moved the registration to the Application:

```swift
app.subscribers = [
//Subscribers
]
```

It's documented but I haven't done the tests yet - I won't be able to get to them until tonight my time, as I have "real" work to do now. :(

tannernelson:
d'oh "real work" >_<

well i promise we won't be changing the folders around again so take your time haha.

PR #77, opened 2016-02-26
Environment Support

shnhrrsn:
### Default Detection

By default, Vapor will first check for an env cli argument, and then default to "local" if there isn't one present.  _NOTE: Detection should default to production, not local, in release builds but I was unable to figure out how to detect release builds via SPM building._

##### Specifying an environment at runtime:

```bash
./.build/debug/App --env=test
```

### Overriding Detection

To override detection, can set an environment detection handler on Application:

```swift
let app = Application()
app.detectEnvironmentHandler = { env in
   #if os(Linux)
      if env == "test" {
         return .Test
      } else {
         return .Production
      }
   #else
      return .Development
   #endif
}
```

### Simple

For a simple comparison, you can use the environment var on Application an do:

```swift
if app.environment == .Production {
   print("In Production")
}
```

### Advanced

For more advanced querying, Application has an inEnvironment() method which can take any number of arguments and will return true if any of them match:

```swift
if app.inEnvironment(.Production, .Test) {
   print("In Production or Test")
} else {
   print("Somewhere else: \(app.environment)")
}
```


tannernelson:
I like this, but I'm wondering why `detectEnvironmentHandler` is needed?

tannernelson:
Additionally, I think `Environment` should be an enum. 

```swift
enum Environment {
  case Production
  case Debug
  case Custom(String)
}
```

shnhrrsn:
I added detectEnvironmentHandler so the app could override default detection behavior.

As an example, in Laravel apps I have nginx/fpm configured to provide a "APP_ENV" environment var and then I have this in my apps:

```php
$app->detectEnvironment(function() {
	return isset($_SERVER['APP_ENV']) ? $_SERVER['APP_ENV'] : (stripos(PHP_OS, 'darwin') === false ? 'production' : 'local');
});
```

tannernelson:
Ah, okay cool. Do you agree with the enum change?

shnhrrsn:
Yeah, I keep forgetting Swift can do that.. I'll make the change

Sources/Vapor/Core/Application.swift
```diff
@@ -39,6 +39,31 @@ public class Application {
...
+		let environment = self.bootEnvironment()
+		self.detectedEnvironment = environment
+		return environment
+	}
+
+	/**
+		Optional handler to be called when detecting the
+		current environment.
+	*/
+	public var detectEnvironmentHandler: ((String) -> String)?
```
tannernelson: I would make this return `Environment`

tannernelson: Also why is the handler passed a String? Couldn't it be `() -> Environment` ?

shnhrrsn:
@tannernelson: detectEnvironmentHandler is being passed the string so it's aware of the value passed in via `--env=` -- we could get rid of that though.

tannernelson:
Ahh, okay that makes sense. I like it.

tannernelson:
Is QA super common? Haven't seen that before.

shnhrrsn:
Suppose it depends on the size of the company/project, I work for CBSi and QA envs are very common here.

Granted anyone who would use/need a QA environment, Vapor isn't mature enough to use.  I just thought I'd cover our bases since it costs us nothing to add various predefined environments vs effort to use custom ones.

mpclarkson:
We use envs all the time too. 

Matthew Clarkson

tannernelson:
I think let's limit what we have to support to Production/Test/Development/Custom for now.

Once the merge conflict is fixed we can merge this.

ketzusaka:
FWIW I believe "QA" is a rather common environment, often referred to as "Staging", though.

tannernelson:
I resolved the merge conflict. Let's see if it passes travis.

shnhrrsn:
Alright all merged and up to date and it's finally passing.

PR #79, opened 2016-02-26
Integrated Jay as the new JSON library

czechboy0:
- works on both OS X and Linux
- I reused the existing convenience type Json and convert between it and Jay's internal type-safe structure
- tested both get and post on both platforms, added those to the VaporDev test executable
- this net removes over 600 LOC from the project

Happy to hear any feedback!

At the moment I didn't check in the folder Packages, but since the only supported way is building with `swift build`, we're not adding any extra work on the developers.

@LoganWright @tannernelson Please review.


Sources/VaporDev/main.swift
```diff
@@ -28,6 +52,6 @@ app.group("abort") {
     }
 }
 
-app.resource("resource", controller: MyController.self)
+//app.resource("resource", controller: MyController.self)
```
LoganWright: I think we should remove commented out code

czechboy0: Sure, I thought it was something that would be put back again. It didn't
compile bc there's no such class.

Sources/VaporDev/main.swift
```diff
@@ -10,6 +10,30 @@ app.get("test") { request in
...
+            "text": "unicorns",
+            "nested": ["one", 2, false]
+        ]
+    )
+}
+
+app.post("json") { request in
+    //parse a key inside the received json
+    guard let json = request.json else { return Response(error: "Invalid json") }
+    guard let count = json.objectValue?["unicorns"]?.int else { return Response(error: "No unicorn count provided") }
```
LoganWright: Here, you should be able to just do `json["unicorns"]?.int`

We also have the data property which is a bit more magical than the existing Json, and can do things like `request.data["unicorns"]?.int`.  I think this is good to use in examples because it coaches users towards the path of least resistance.  `data` compiles form encoding, json, and url query parameters all together

Sources/VaporDev/main.swift
```diff
@@ -10,6 +10,30 @@ app.get("test") { request in
...
+    return "Received \(count) unicorns"
+}
+
+app.post("json2") { request in
+    //parse a key inside the received json
+    guard let json = request.json else { return Response(error: "Invalid json") }
+    guard let count = json.objectValue?["unicorns"]?.int else { return Response(error: "No unicorn count provided") }
```
LoganWright: Same here, I think we should encourage `.data`

czechboy0: Yeah fixed both, just pushed it.

Sources/VaporDev/main.swift
czechboy0: But I still removed it :)

PR #83, opened 2016-02-27
Event Subscribers

mpclarkson:
Here's the updated PR for the event subscribers. It is similar to the Laravel implementation. Django, Symfony and other frameworks provide similar functionality.

This implementation allow users to register subscribers that hook into framework and custom events. Subscribers are processed in order of `priority` and propagation can be cancelled by returning `false` in the Subscriber `handle` method.

Methods are documented and tests are passing. When you merge this, I'll write a wiki page.

```swift
//main.swift

let app = Application()

app.subscribers = [
   UserRegistrationSubscriber(),
   UserLoginSubscriber(),
]

app.start()

```

```swift

enum UserEvents: String, EventType {
    case Login, Registration, Logout
}

class UserLoginSubscriber() {
   let priority = 50
   let type = UserEvents.Login

   func handle(data: Any) -> Continue {
    if let user = data as? User {
      //Do something with the user, e.g. save the event to the db
    }
    return true  //or false to cancel propagation
}

//Somewhere in the app, e.g. a controller or middleware
let event = Event(UserEvents.Login, data: user)
Fire.event(event)

```


LoganWright:
I think the backing data instead of:

` public var subscribers: [Subscriber]`

Should be:

`public var subscribers: [String : [Subscriber]]`.  Where the string key is the event identifier. Right now we're filtering on new notifications, but a hash table lookup is going to be much more efficient. 

The other thing that I think we really need to consider is the memory implications of this.  As written, there's two main assumptions that we could avoid making.

- Any object that is a subscriber is a singleton as well.  If its not a singleton, it becomes one by subscribing.
- No subscribers will be added or removed during the lifecycle of the app

I would prefer to structure things in a way that does not make these assumptions and allow more flexible event subscriptions, but if we're going to make them, I think we should protect the user.  ie: if they add to or remove from application's subscribers after booted, will have no effect.  We could warn the user about this.  

My main concern is the singleton issue with subscriber objects.  Singletons are prone to problems and inherently carry state from place to place where maybe its not wanted.

I think we can do a few things to make it much more flexible:
 
- Use notification centers instead of an application that requires subscribers from boot to death. Users can subscribe to and fire from this notification center.  A default could be provided, but users could also customize their own. This also decouples it from Application.
- Hold references in weak container or make it clear how to unsubscribe and that unsubscribing is necessary to deallocate an instance.  I'm partial to a weak container, ie: `subscribers: [String : [Weak<Subscriber>]]` because its automatic and lets the user manage their own memory.

These two additions don't require much effort and it would give users more control in how they utilize the notification system.  Even if the current implementation captures the desired use, everything here can be preserved while also allowing users to control the behavior.

Would love other thoughts on this: @tannernelson @mpclarkson @shnhrrsn

shnhrrsn:
@LoganWright I agree with everything you said

I'm also in favor of building ontop of notification center, I don't see why we should try to reinvent the wheel here.

In another Swift project of mine, I have a "Message" class that provides a simple API around NSNotificationCenter:

```swift
public class Message {
	private static var ordinal = 0
	private let notificationName: String

	public init() {
		self.notificationName = "Message-\(++self.dynamicType.ordinal)"
	}

	public func subscribe(listener: AnyObject, action: Selector, object: AnyObject? = nil) {
		NSNotificationCenter.subscribe(listener, action: action, message: self.notificationName, object: object)
	}

	public func unsubscribe(listener: AnyObject, object: AnyObject? = nil) {
		NSNotificationCenter.unsubscribe(listener, message: self.notificationName, object: object)
	}

	public func publish(object: AnyObject? = nil) {
		NSNotificationCenter.publish(self.notificationName, object: object)
	}
}
```

And then I do something like this:

```swift
public class User {
	public static let didAuthenticateMessage = Message()

	public final func authenticate(token: String) {
		self.token = token
		self.dynamicType.didAuthenticateMessage.publish(self)
	}
}

class SomeController {
	init() {
		User.didAuthenticateMessage.subscribe(self, action: Selector("userDidAuthenticate"))
	}
}
```

Not saying we copy this exactly, but just an example of how we can leverage notification center while still providing our own set of APIs that feel more at home in Vapor.

mpclarkson:
Thanks for the feedback guys. I'll look into the other approaches and update the PR.

mpclarkson:
I am closing this, as I won't be able to get back to Vapor stuff until later in the week (very busy at work). I'll resubmit later.

PR #98, opened 2016-03-02
Adds the ability to define Middleware that should be ran on a specific controller

ketzusaka:
Open to comments on this one. This particular feature is inspired by Rails before_filter and Laravel's Controller Middleware. It opens the doors for certain controllers to be behind a logged in user, while others open to everyone. This is a very common need in most web applications, and I'm sure there are plenty more usecases where you'd want a middleware to run on only a certain controller.

### DIFF
--- a/Sources/Vapor/Routing/Application+Route.swift
+++ b/Sources/Vapor/Routing/Application+Route.swift
@@ -53,7 +53,7 @@ extension Application {
      
         Note: You are responsible for pluralizing your endpoints.
     */
-    public final func resource(path: String, controller: ResourceController.Type) {
+    public final func resource<RoutedController: ResourceController>(path: String, controller: RoutedController.Type) {
         let last = "/:id"
         let shortPath = path.componentsSeparatedByString(".")
             .flatMap { component in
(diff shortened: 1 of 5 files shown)


Sources/Vapor/Routing/Application+Route.swift
```diff
@@ -64,13 +64,27 @@ extension Application {
...
         // ie: /users/:id
-        self.get(fullPath) { try controller.init().show($0) }
-        self.put(fullPath) { try controller.init().update($0) }
-        self.delete(fullPath) { try controller.init().destroy($0) }
+        self.add(.Get, path: fullPath, action: RoutedController.show)
+        self.add(.Put, path: fullPath, action: RoutedController.update)
+        self.add(.Delete, path: fullPath, action: RoutedController.destroy)
+    }
+
+    public final func add<RoutedController: Controller>(method: Request.Method, path: String, action: (RoutedController) -> (Request) throws -> ResponseConvertible) {
```
LoganWright: This seems like a good addition for the potential of additional custom controllers and it cleans up the declarations in post above.

Sources/Vapor/Routing/Application+Route.swift
```diff
@@ -64,13 +64,27 @@ extension Application {
...
+        var handler: Request.Handler = { request in
+            let controller = RoutedController()
+            let actionCall = action(controller)
+            return try actionCall(request).response()
+        }
+
+        for mw in RoutedController.middleware {
```
LoganWright: I think we should stick with the original scoped middleware implementation and not bind controllers to this. I think we can achieve the same effect working within the framework

ketzusaka:
Thanks. Going to close this and just add the add method. Didn't realize there was scoped grouping of middleware.

I do feel like there is value in a controller specifying middleware that it runs for itself, but I'll do that separately.

Sources/Vapor/Routing/Controller.swift
```diff
@@ -2,4 +2,12 @@ import Foundation
 
 public protocol Controller {
     init()
-}
\ No newline at end of file
+    static var middleware: [Middleware.Type] { get }
```
LoganWright: I think we should leave this one to the scoped middleware on application for consistency, ie:

```Swift
app.middleware([scopedMiddlewares]) {
    // handle and add scoped controller here
}
```

PR #124, opened 2016-03-10
Refactored to include support for user-instantiated controllers.

svanimpe:
Please note that I'm developing on Linux, so my tooling support is very limited. I tested both `resource` methods and both `add` methods using my example app and everything worked for me, but I recommend more and better testing :)


Sources/Vapor/Routing/Controller.swift
```diff
@@ -1,5 +1,24 @@
...
+ * `Controller`. Controls group related route logic into
+ * a single protocol that, by default, conforms to standard
+ * CRUD operations.
+ */
 public protocol Controller {
-    init()
-}
\ No newline at end of file
+    /// Display many instances
+	  func index(request: Request) throws -> ResponseConvertible
```
svanimpe: Oh OK. @LoganWright's comment on #112 moved them back to `Controller`, that's why. What would you like to be in `Controller` then, if `init` is moved to the general purpose `DefaultInitializable`?

ketzusaka: Just an empty protocol for now. I'm imaging a place to add protocol extensions to that work on both resource controllers and regular controllers.

svanimpe: What do you mean with 'regular controllers'? Just controllers that do routing but don't fit into the resource CRUD scheme? Or some other type of controller?

ketzusaka: Yep exactly. Or controllers that may partially implement the CRUD scheme; I don't want to always force people to implement all of those methods to be considered a controller.

svanimpe: I see. I really feel like we're talking about two different things here:
* Regular vs. resource controllers.
* Default initialized vs. user initialized controllers.
I'd very much prefer a design that addresses there separately:
* User initialized: supply a factory when registering.
* Default initialized: conform to `DefaultInitializable`.
* Regular: no requirements (for now), just contains `Route.Handlers`.
* Resource: conform to `ResourceController`.
This keeps things separate and leaves room for a `Controller` protocol (once it's clear what that should be).

ketzusaka: What do you mean "just contains `Route.Handlers`?

Also, I was digging through some use cases, and maybe `Controller` as an empty protocol isn't required. However, I would like a `app.add` that takes a generic conforming to `DefaultInitializable`, so any class can define an action.

svanimpe: > What do you mean "just contains Route.Handlers?

I assume a regular controller simply contains methods that handle requests? I was referring to the `Route.Handler` typealias (Request -> Response).

> Also, I was digging through some use cases, and maybe Controller as an empty protocol isn't required. However, I would like a app.add that takes a generic conforming to `DefaultInitializable`, so any class can define an action.

Yes, that's what I meant with 'conform to `DefaultInitializable`'.

I've made this changed in my forked repo. I hope they show up in this PR?

ketzusaka:
Awesome. Need to add the new files to Xcode though, if if you don't have access to a Mac we could do that. 

@tannernelson @LoganWright any input on this?

tannernelson:
Can you run me over the motivation for this feature?

ketzusaka:
@tannernelson Often people will want to provide dependencies into their classes. Since our system handles all of the instantiation currently, we wanted to provide a way to allow a factory to handle the instantiation.

For example, Under something like Fluent, the DB is basically static. If the user wants to use something that isn't like that, one way they would set it up is create the DB in the same place they instantiate an `Application`, and pass it into the controllers that would want to use it. Under Test, they may want to provide a different object. This helps decouple things within the application. 

Laravel allows you to do this in a less explicit way through type-hinting: https://laravel.com/docs/5.0/controllers#dependency-injection-and-controllers

tannernelson:
Okay, this seems like a good addition.

A few things:
- [] Is `Factory` a widely used Swift term? I don't think I've seen it before. 
- [] Merge master into this branch + ensure Travis passes
- [] Documentation in Markdown. You can paste it here and I'll add it to the Wiki.

ketzusaka:
@tannernelson Great :)

Regarding #1, it is a frequently mentioned word on the Swift API guidelines: https://swift.org/documentation/api-design-guidelines/

svanimpe:
@tannernelson Factories are a common design pattern (or meta-pattern, as it has several variations) in OO programming. It's just a fancy word for objects that are in charge of instantiating other objects, adding a layer of indirection to instantiation.

tannernelson:
Cool. I've just never seen a method with a `factory` param. It seems like they use the word `make` for methods and such. But they don't mention parameter labels. `makeWith:` might be better?

tannernelson:
screen shot 2016-03-14 at 10 11 30 pm

Interesting

tannernelson:
@svanimpe I've added you to the org so you can move this branch to `qutheory/vapor`.

ketzusaka:
@svanimpe Think we could make the final alterations soon? I'd love to see this get in =)

svanimpe:
@tannernelson How do I go about moving a branch? Never done that before.
@ketzusaka Do you mean adding the files to Xcode? Is this simply a matter of opening the include Xcode project and adding the new files to it?

ketzusaka:
@svanimpe I was referring to the argument label "controllerFactory" being swapped with "makeWith" to be more consistent with Swift 3 API guidelines.

For adding a new branch, add this repo to your local repo like so:

```
git add remote qutheory git@github.com:qutheory/vapor.git
```

Then push your branch like this:

```
git push qutheory master:svan/controller-factories
```

That will create a new branch on qutheory with the name "svan/controller-factories"

For Xcode, yes, just opening and adding the files to the appropriate groupings.

svanimpe:
@ketzusaka @tannernelson The `make` prefix is for function names, not parameters. `makeWith` doesn't sound right as a role name, in part because the function name `add` doesn't make it clear what is being made. That's part of why I opted for `controllerFactory`.

tannernelson:
I would opt for `makeControllerWith: ` over `controllerFactory:` just my opinion.

svanimpe:
Branch pushed as `svanimpe/controller-factories`!

tannernelson:
Moved to #140

PR #148, opened 2016-03-23
Lock Hummingbird with minor version

ketzusaka:
I'll be updating Hummingbird/Strand to 1.1 soon, which will be compatible with Swift 2.2 and the 03-16 dev snapshot. By locking on 1.0, Vapor will continue to work under the 03-01 snapshot, and when its ready for 03-16, it can update to 1.1.


LoganWright:
:+1:

PR #163, opened 2016-03-31
Swiftlint autocorrect

shnhrrsn:
Pending #162


Sources/Vapor/Hash/CryptoSwift.swift
```diff
@@ -177,11 +177,11 @@ extension Int {
     func bytes(totalBytes: Int = sizeof(Int)) -> [UInt8] {
         return arrayOfBytes(self, length: totalBytes)
     }
-    
+
     static func withBytes(bytes: ArraySlice<UInt8>) -> Int {
         return Int.withBytes(Array(bytes))
     }
-    
+
```
LoganWright: Why all the spacing commits?

LoganWright:
Looks good to me, I'd prefer to remove the whitespace additions, but I'm not sure it's worth it. I know Xcode will move them back if someone is in the file area and formats. 

Happy to have some enforced linting!!!

:shipit: from me

shnhrrsn:
@LoganWright once we turn on houndci (see #162) if anyone screws up, the hound bot will comment in the PR to let you know and the test will fail, so that should protect us from this moving forward.

Everyone should probably turn this on in Xcode though:

PR #186, opened 2016-04-20
New Snapshot - 04-12

LoganWright:
You can test w/ this dependency:

```Swift
.Package(url: "https://github.com/LoganWright/vapor.git", majorVersion: 0, minor: 6)
```


Sources/Vapor/Process/Process.swift
```diff
@@ -10,11 +12,69 @@ extension Process {
...
         for argument in arguments where argument.hasPrefix("--\(name)=") {
-            return argument.split("=").last
+            return argument.split(byString: "=").last
         }
         return nil
     }
-
 }
+
+/*
```
LoganWright: rm all of this for sure

Sources/Vapor/Core/Application.swift
```diff
@@ -195,7 +195,7 @@ public class Application {
         optionally runs on the supplied
         ip & port overrides
     */
-    public func start(ip ip: String? = nil, port: Int? = nil) {
+    public func start(ip: String? = nil, port: Int? = nil) {
```
LoganWright: I think we should ignore this one.

czechboy0:
I assume #190 is still present here, right?

PR #332, opened 2016-05-27
Tiny router improvement

Evertt:
#325


Sources/Generator/main.swift
```diff
@@ -35,9 +35,17 @@ struct Func: CustomStringConvertible {
...
         }
         
-        let paramsString = params
-            .map { param in "_ \(param)" }
-            .joined(separator: ", ")
+        let paramsString: String
+        
+        if let param = params.first where params.count == 1 && param.type == .Path {
+            paramsString = "_ \(param) = \"/\""
+        }
```
LoganWright: Let's move this a bit to stick w/ `} else {` style

LoganWright:
Stylistic adjustment and tests passing, otherwise this one looks good to me :shipit:

Sources/Generator/main.swift
Evertt: Done

PR #336, opened 2016-05-29
Make Hash initializer public

bygri:
I believe that the Hash class's initialiser is not made public because configuration of the Hash is expected to be performed by:

    app.hash.key = "secretkey"

However, in the initialiser of Application it is possible to set your own Hash instance, so it would appear to make sense to allow to be inited outside of the scope of the init of an Application.

There are also use cases for being able to have more than one Hash instance (with different keys) available at one time, or having an instance available without needing to create an Application along with it. I also can't think of any particularly good reasons for disallowing it, hence this pull request.

### DIFF
--- a/Sources/Vapor/Hash/Hash.swift
+++ b/Sources/Vapor/Hash/Hash.swift
@@ -24,6 +24,13 @@ public class Hash {
     public var driver: HashDriver = SHA2Hasher(variant: .sha256)
 
     /**
+        Initialize the Hash.
+    */
+    public init() {
+
+    }
+
+    /**
         Hashes a string using the `Hash` class's
         current `HashDriver` and `applicationString` salt.
 
--- a/Sources/Vapor/Hash/Hash.swift
+++ b/Sources/Vapor/Hash/Hash.swift
@@ -13,7 +13,7 @@ public class Hash {
         the same during the lifetime of your application, since
         changing it will result in mismatching hashes.
     */
-    public var key: String = ""
+    private let key: String
 
     /**
         Any class that conforms to the `HashDriver`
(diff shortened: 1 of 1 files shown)


LoganWright:
@bygri thanks for the pull request, contributions from the community are always awesome to see!

I understand now, what this issue was referring to https://github.com/qutheory/vapor/issues/195

These changes look good so far, glancing at the hash code, I think I have an issue w/ <a href="https://github.com/qutheory/vapor/pull/336/files#diff-dc33adb9450f9cb86dc1e5bf6831072aR16">this var</a>.

It's marked `var`, but based on its comments, it should absolutely be a `let` constant.

Would you mind also changing that, and making the initializer:

```
public init(key: String) {
    self.key = key
}
```

Unless there's objections from other members, once we have the updated `key` attribute, we'll get this merged in next release 👍

bygri:
Agreed. Just didn't want to push it too far! Based on your comments, I have gone ahead and pushed it too far. In my view, setting the hash Driver is similar to setting the key, and may as well be done at init only. Also, if the key and driver properties are constant, I don't see any benefit in keeping them public so I made them private lets.

bygri:
Making the initialiser require a key and keeping the key constant has a knock-on effect with Application() which needs to know the key at init so it can configure the Hash properly. This would require all users currently setting a hash key to alter their code to provide the key at Application.init(), which is not a change I want to make without some direction.

Can I just submit a PR making the initialiser public, which wouldn't affect any current code, and then subsequently put together a PR addressing your earlier comments for consideration?

LoganWright:
@bygri that sounds great to me 👍

PR #353, opened 2016-06-03
allow custom JSON serializers

LoganWright:


Sources/Vapor/JSON/JSON.swift
```diff
@@ -4,6 +4,11 @@ import PureJSON
 public typealias JSON = PureJSON.JSON
```
czechboy0: Shouldn't this be `C7.JSON`? Otherwise it won't be compatible with other JSON libs AFAIK.

LoganWright: @czechboy0 they should be the same thing, our fork was updated to use the same struct

PR #406, opened 2016-06-24
Controllers to use RESTFul verbs

wess:
Just changed the controller's default mapped methods to be that of RESTFul verbs. This will make it clearer for developers to know what type of request they are making.

### DIFF
--- a/Sources/Development/Controllers/UserController.swift
+++ b/Sources/Development/Controllers/UserController.swift
@@ -18,9 +18,9 @@ class UserController: Controller {
     /**
         Create a new instance.
      */
-    func store(request: Request) throws -> ResponseRepresentable {
+    func post(request: Request) throws -> ResponseRepresentable {
         return JSON([
-            "controller": "MyController.store"
+            "controller": "MyController.post"
         ])
     }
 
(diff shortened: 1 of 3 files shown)


LoganWright:
@qutheory/contributors How do people feel about this semantic change? It diverges from concepts established in frameworks like Rails, however, semantically, it's a bit more clear what's getting called. 

This makes things much closer to route groups, perhaps we could use a variant of route group to provide this behavior more efficiently.

jchannon:
As it should be! Also where is OPTIONS!?

LoganWright:
@jchannon the goal is to provide `options` by default somehow based on what methods are implemented

Casperhr:
On more advanced apis you will have several POST/PATCH/GET on same model. Fx patch for update is fine, but if you can pass files as multipart entity, it has to be POST

Therefor I vote for not naming the function the method by default

jchannon:
Not really sure what Casper is saying, I don't know vapor too well but a
controller should be able to handle multiple verbs

Casperhr:
That was my point, I don't like a pattern where you think the http method is equal the function name

wess:
Well if you look at other rest api frameworks, you will see this pattern for class based namespacing, if you want to do custom/multiple put, you still could and just route them. @Casperhr argument is fine, but the way it currently is, are you not still forcing them to your pattern/naming of functions?

kdawgwilk:
I am 👍 for this change because this was one of the first things that stuck out to me as being odd when I first created a controller

wess:
I need to fix the CI

tannernelson:
This should be included now.

Issue #409, opened 2016-06-28
Overflow when casting UInt32.max to Int...


dgsmith:
# Bugs

Trying to build on a BeagleBone Black with swift 3. All dependencies compile, but in `Linking .build/debug/Generator` I get the error: 
```
/home/debian/server/Packages/Vapor-0.11.1/Sources/Vapor/WebSockets/Model/Frame.swift:357:50: error: integer overflows when converted from 'UInt32' to 'Int'
                let key = UInt32(libc.random() % Int(UInt32.max))
                                                 ^
<unknown>:0: error: build had 1 command failures
error: exit(1): /home/debian/swift-3.0/usr/bin/swift-build-tool -f /home/debian/server/.build/debug.yaml
```

Shouldn't the line of code just be:
```swift
let key = UInt32(libc.random()) % UInt32.max
```

After changing that line of code, my project compiles, thanks!

LoganWright:
It appears the issue is that Int32 is the max Int on that system, so it's breaking the overflow which makes sense. Would you submit a PR for that patch, or I can add it in today

PR #414, opened 2016-06-30
Fixed overflow error for certain linux builds when casting UInt32.max to Int

dgsmith:
issue described in #409


LoganWright:
Looks good, once travis passes, we'll merge this in. Thanks for the contribution @dgsmith

dgsmith:
glad I could help!

PR #415, opened 2016-06-30
Fixed bug where Abort.notFound wasn't propagated to AbortMiddleware

timominous:
Replication steps:

1. Visit an unimplemented route.


LoganWright:
Looks good @timominous thanks for the addition! Once Travis finishes, we'll merge it 👍

PR #426, opened 2016-07-03
Fix for changes implemented in upcoming swift snapshot

damuellen:
This pull request takes care of the latest changes in the compiler (SE0060),
order of the arguments in Package.swift
And dropping the `NS` prefix in some Foundation types. (SE0086)
Merge when moving to a new swift snapshot.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -2,6 +2,26 @@ import PackageDescription
 
 let package = Package(
     name: "Vapor",
+        targets: [
+        Target(
+            name: "Vapor"
+        ),
+        Target(
+            name: "Development",
+            dependencies: [
+                .Target(name: "Vapor")
+            ]
+        ),
+        Target(
+            name: "Performance",
+            dependencies: [
+                .Target(name: "Vapor")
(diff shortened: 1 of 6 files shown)


LoganWright:
@damuellen thanks so much for the contribution! The current master branch, already has support for latest snapshots, the NS stuff is handled on typealiases because although Foundation is updated, the naming changes haven't made it to Linux yet.

You can see this in the [`Bridge.swift`](https://github.com/qutheory/vapor/blob/bb9ecfbf4dbe8fe9fc10ee120b59cd82732e60a0/Sources/Vapor/Utilities/Bridge.swift).

damuellen:
I am building Swift from source for armv7. So I am a little ahead of the last snapshot.

LoganWright:
@damuellen I see now, that's awesome! Thanks for doing that. We'll leave this up until next snapshot is released and merge right in 👍

damuellen:
I have deleted the fork the request was created from. So I will close this.

PR #438, opened 2016-07-07
Update MediaType.swift

sebastianheinig:


LoganWright:
@sebastianheinig good catch, thanks for the contribution 👍

PR #453, opened 2016-07-13
Simplified Package.swift

czechboy0:
This should work exactly the same way, there was no need to explicitly specify the `Generator` target. And dependencies of targets can be just strings, no need to manually create the `.Target` structs.


LoganWright:
This seems good to merge to me, it wants a rebase or merge, but if there's no conflicts, we could probably push this through.

PR #459, opened 2016-07-17
Swift will soon enforce the order of defaulted parameters

damuellen:
The order must than targets, dependencies, testDependencies, exclude

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -29,6 +29,7 @@ targets += [
 
 let package = Package(
     name: "Vapor",
+    targets: targets,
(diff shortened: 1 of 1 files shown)


LoganWright:
@damuellen we had one other package.swift PR that had been queued up, can you make sure those conflicts are resolved and we'll get this one in 👍

damuellen:
@LoganWright Did you mean #451 ?

LoganWright:
@damuellen I was referring to #453

PR #563, opened 2016-09-01
Added new up-to-date tutorial written on Septembers 1th 2016

MartinLasek:
It's actually a tutorial written and published on my own. I will continuously improve it, add new steps getting deeper very soon. My goal is to keep every single step just up-to-date as vapor evolves, so there's a tutorial out there you can rely on it will work right away :)! My idea is to provide tutorial-versioning, so you can switch between vapor versions on my site and according to the chosen one, the right steps are provided to get vapor run in that version (starting with version 16) :)


LoganWright:
Thanks @MartinLasek 👍

PR #592, opened 2016-09-11
Update PROJECTS.md

tadija:
I've added a sample Loyalty backend system made with Vapor and did some minor fixes for SwiftyBeaver (added it on top list and fixed demo URL).


LoganWright:
@tadija thanks, looks great!

PR #607, opened 2016-09-15
Implemented conditional requests, Fixes #539

### LINKED ISSUE #539: Caching and Conditional Requests
Currently, Vapor has no caching support/mechanism for HTTP Requests.

Even for static files ( images, logos, css, js ... ) Vapor is not retuning proper headers for browser to cache the data,

Current Response:

```
Connection:keep-alive
Content-Length:211183
Content-Type:application/javascript
Date:Tue, 23 Aug 2016 13:40:44 GMT
Server:Vapor 0.16.0
```

Missing headers:

```
Cache-Control: max-age=21600
Last-Modified: Wed, 01 Sep 2016 16:34:52 GMT
ETag: "3115cba6"
```

After implementing the missing headers/caching, browser will be able to send :

```
If-Modified-Since: Wed, 01 Sep 2016 16:34:52 GMT
If-None-Match: "3115cba6"
```

For static files Vapor should be able to process the request and provide proper result ( 304 when not modified ) itself.

Also, would be nice, to have helper functionality for custom request handlers, to do the same.

ogres:
Implemented conditional requests for static files using ETag ( https://github.com/vapor/vapor/issues/539 )

Added few tests to FileManager.

This commit handles requests to the Public Directory.

It checks for header "If-None-Match" and compares to the local files ETag, If ETags match, returns 304 not modified,

If header is not present or does not match local ETag, returns 200 with actual file's content and ETag header.

### DIFF
--- a/Sources/Vapor/Droplet/Droplet.swift
+++ b/Sources/Vapor/Droplet/Droplet.swift
@@ -269,7 +269,7 @@ public class Droplet {
 
         // default available middleware
         var am: [String: Middleware] = [
-            "file": FileMiddleware(workDir: workDir),
+            "file": FileMiddleware(publicDir: workDir + "Public"),
             "validation": ValidationMiddleware(),
             "date": DateMiddleware(),
             "type-safe": TypeSafeErrorMiddleware(),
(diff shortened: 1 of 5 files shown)


LoganWright:
Moving to #608

PR #636, opened 2016-10-06
ProtectMiddleware tests + better auth tests

adriencanterot:
Hi, 
On this PR I added a test for the protect middleware that lacked some.
Also, I reorganised the tests to be more modular and make test functionalities that would require auth easier.

Thanks for reviewing !

### DIFF
--- a/Tests/VaporTests/AuthTests.swift
+++ b/Tests/VaporTests/AuthTests.swift
@@ -57,12 +57,12 @@ class AuthTests: XCTestCase {
     
     func testUserLogin() throws {
         
-        let authenticated = try login(credentials: Identifier(id:1))
+        let authenticated = try login(credentials: Identifier(id: 1))
         XCTAssertEqual(authenticated.status, .ok)
         XCTAssertEqual(authenticated.data["name"]?.string, "John")
         XCTAssertNotNil(authenticated.cookies["vapor-auth"])
         
-        let notAuthenticated = try login(credentials: Identifier(id:2)) // user with id of 2 does not exist in the database
+        let notAuthenticated = try login(credentials: Identifier(id: 2)) // user with id of 2 does not exist in the database
         XCTAssertNotEqual(notAuthenticated.status, .ok)
         XCTAssertNil(notAuthenticated.cookies["vapor-auth"])
         
@@ -83,7 +83,7 @@ class AuthTests: XCTestCase {
         var protected = try droplet.respond(to: secure)
         XCTAssertEqual(protected.status, .methodNotAllowed) // thrown from ProtectMiddleware
         
-        let authenticated = try login(credentials:Identifier(id:1))
+        let authenticated = try login(credentials:Identifier(id: 1))
         secure.cookies = authenticated.cookies //set appropriate login cookie
         
         protected = try droplet.respond(to: secure)
@@ -106,7 +106,7 @@ class AuthTests: XCTestCase {
         
         let request = Request(method: .get, path: "logout")
         
-        let authenticated = try login(credentials: Identifier(id:1))
+        let authenticated = try login(credentials: Identifier(id: 1))
         request.cookies = authenticated.cookies //set appropriate login cookie
         
         let res = try droplet.respond(to: request)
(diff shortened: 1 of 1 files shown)


Tests/VaporTests/AuthTests.swift
```diff
@@ -24,53 +26,92 @@ class AuthTests: XCTestCase {
...
     
-    func testUserLoginAndOut() throws {
-        let request = Request(method: .get, path: "login")
+    /**
+        Logs in a basic user with provided credentials.
+        Can be used in any tests requiring auth.
+        Will provide a response with an auth cookie only
+        with an Identifier with an id of 1.
+    */
+
```
LoganWright: remove this newline here so comments are right up against the signature.

Tests/VaporTests/AuthTests.swift
```diff
@@ -24,53 +26,92 @@ class AuthTests: XCTestCase {
...
         
-        XCTAssertEqual(res.status, .ok)
-        XCTAssertEqual(res.data["name"]?.string, "John")
-        XCTAssertNotNil(res.cookies["vapor-auth"])
+        return try droplet.respond(to: login)
+    }
+    
+    func testUserLogin() throws {
+        
+        let authenticated = try login(credentials:Identifier(id:1))
```
LoganWright: let's put spaces between external parameters and the value `: Identifier(id: 1)`

Tests/VaporTests/AuthTests.swift
```diff
@@ -24,53 +26,92 @@ class AuthTests: XCTestCase {
...
+        XCTAssertEqual(authenticated.status, .ok)
+        XCTAssertEqual(authenticated.data["name"]?.string, "John")
+        XCTAssertNotNil(authenticated.cookies["vapor-auth"])
         
-        try testUserLogout(with: res) //try to logout with content of response (w/ cookies etc.)
+        let notAuthenticated = try login(credentials:Identifier(id:2)) // user with id of 2 does not exist in the database
```
LoganWright: same here, spaces on `:`

Tests/VaporTests/AuthTests.swift
```diff
@@ -24,53 +26,92 @@ class AuthTests: XCTestCase {
...
+                return try req.auth.user().converted(to: JSON.self)
             }
+        }
+        
+        let secure = Request(method: .get, path:"secure")
+
+        var protected = try droplet.respond(to: secure)
+        XCTAssertEqual(protected.status, .methodNotAllowed) // thrown from ProtectMiddleware
+        
+        let authenticated = try login(credentials:Identifier(id:1))
```
LoganWright: same here

Tests/VaporTests/AuthTests.swift
```diff
@@ -24,53 +26,92 @@ class AuthTests: XCTestCase {
...
+        Can be used in any tests requiring auth.
+        Will provide a response with an auth cookie only
+        with an Identifier with an id of 1.
+    */
+    
+    func login(credentials:Identifier) throws -> Response {
+        
+        //First : creates a dummy user with an Id of 1 if not exists
+        User.database = self.database
+        var user = try User.find(1)
```
LoganWright: do we need the user below? I think we might've been using it before, but something like this might be easier to remember at a glance:

```
if try User.find(1) == nil {
    let user = User(name: "John")
    try user.save()
}
```

Tests/VaporTests/AuthTests.swift
```diff
@@ -86,32 +127,21 @@ class AuthTests: XCTestCase {
...
-            throw Error.invalidCredentialType
-        }
-        
-        let user = try User.find(id.id)
+        let id = credentials as? Identifier
+        XCTAssertNotNil(id)
         
-        guard let u = user else {
-            throw Error.userNotFound
+        guard let user = try User.find(id!.id) else {
```
LoganWright: I've been trying to personally start avoiding the `!` in tests since they kill the suite before others can run which can be harder when doing  ci runs. For this one, we could do:

```swift
guard let id = id?.id, let user = try User.find(id) ...
```

adriencanterot:
Okay ! I changed a few things according to what you reviewed.
Thanks by the way, I understand it's important to keep style consistent :).

PR #642, opened 2016-10-10
Fixed jpg and jpeg mime type retrieval for FileMiddleware

sroebert:
The key for retrieving the `jpg` and `jpeg` mime type, was a combined comma separated list of `jpg, jpeg`. This obviously does not work when trying to use `jpg` or `jpeg` as a key. This commit solves that by creating separate entries for both `jpg` and `jpeg`.


Sources/Vapor/Content/MediaType.swift
```diff
@@ -489,7 +489,8 @@ let mediaTypes = [
     "g3": "image/g3fax", // G3 Fax Image Wikipedia: G3 Fax Image
     "gif": "image/gif", // Graphics Interchange Format Wikipedia: Graphics Interchange Format
     "ief": "image/ief", // Image Exchange Format RFC 1314
-    "jpeg, .jpg": "image/jpeg", // JPEG Image RFC 1314
+    "jpeg": "image/jpeg", // JPEG Image RFC 1314
```
LoganWright: Is there a situation where we might receive `"jpeg, .jpg"` as well as these other two? If no, then let's get this merged. Thanks for the additions!

sroebert: Currently there is none. I was thinking of optimising it for that, but it doesn't seem worth it, until there really is a need.

natebird:
This PR looks good to merge.

Sources/Vapor/Content/MediaType.swift
LoganWright: Thanks @sroebert this is getting merged in today. Appreciate the contributions.

PR #659, opened 2016-10-22
Avoid setting an empty string as a Set-Cookie header.

sroebert:
When no cookies were set, Vapor was still returning an empty `Set-Cookie` header. This resulted in problems when I was testing with Postman and I am assuming this is not wanted behaviour.

This pull request fixes this, by validating that the serialisation result of the cookies is not empty. In the case that it is empty, it simply removes the `Set-Cookie` header altogether.

### DIFF
--- a/Sources/Cookies/Request+Cookies.swift
+++ b/Sources/Cookies/Request+Cookies.swift
@@ -23,7 +23,14 @@ extension Request {
         }
         set(cookie) {
             storage["Cookie"] = cookie
-            headers["Cookie"] = cookie.serialize(for: .request)
+            
+            let cookieHeader = cookie.serialize(for: .request)
+            if !cookieHeader.isEmpty {
+                headers["Cookie"] = cookieHeader
+            }
+            else {
+                headers["Cookie"] = nil
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Sources/Cookies/Request+Cookies.swift
```diff
@@ -23,7 +23,14 @@ extension Request {
         }
         set(cookie) {
             storage["Cookie"] = cookie
-            headers["Cookie"] = cookie.serialize(for: .request)
+            
+            let cookieHeader = cookie.serialize(for: .request)
+            if !cookieHeader.isEmpty {
+                headers["Cookie"] = cookieHeader
+            }
```
LoganWright: can we bump the `else` statement up so that it's `} else {`

Sources/Cookies/Response+Cookies.swift
```diff
@@ -23,7 +23,14 @@ extension Response {
         }
         set(cookie) {
             storage["Set-Cookie"] = cookie
-            headers["Set-Cookie"] = cookie.serialize(for: .response)
+            
+            let cookieHeader = cookie.serialize(for: .response)
+            if !cookieHeader.isEmpty {
+                headers["Set-Cookie"] = cookieHeader
+            }
+            else {
```
LoganWright: Same here to keep style consistent `} else {`

sroebert: Will change it now, I'm used to the other way :) Might be nice to add SwiftLint or something to Vapor by the way, to catch these things.

PR #660, opened 2016-10-23
Updated AuthMiddleware to allow for a custom cookie name

sroebert:
The `AuthMiddleware` has an option to pass a cookie factory closure. The name of the cookie however, is fixed to a private string variable. This makes the cookie factory closure a bit difficult to use, as the name has to be hardcoded as the same private string in `AuthMiddleware`. 

This pull request overcomes this, by making the cookie name a property that can be passed when creating the `AuthMiddleware`.

### DIFF
--- a/Sources/Auth/Authentication/AuthMiddleware.swift
+++ b/Sources/Auth/Authentication/AuthMiddleware.swift
@@ -4,21 +4,24 @@ import Cookies
 import Foundation
 import Cache
 
-private let cookieName = "vapor-auth"
+private let defaultCookieName = "vapor-auth"
 private let cookieTimeout: TimeInterval = 7 * 24 * 60 * 60
 
 public class AuthMiddleware<U: User>: Middleware {
     private let turnstile: Turnstile
+    private let cookieName: String
     private let cookieFactory: CookieFactory
 
     public typealias CookieFactory = (String) -> Cookie
 
     public init(
         turnstile: Turnstile,
+        cookieName: String,
         makeCookie cookieFactory: CookieFactory?
     ) {
         self.turnstile = turnstile
-
+        
+        self.cookieName = cookieName
         self.cookieFactory = cookieFactory ?? { value in
             return Cookie(
                 name: cookieName,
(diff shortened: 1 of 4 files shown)


Sources/Auth/Authentication/AuthMiddleware.swift
```diff
@@ -4,21 +4,24 @@ import Cookies
...
 public class AuthMiddleware<U: User>: Middleware {
     private let turnstile: Turnstile
+    private let cookieName: String
     private let cookieFactory: CookieFactory
 
     public typealias CookieFactory = (String) -> Cookie
 
     public init(
         turnstile: Turnstile,
+        cookieName: String,
```
LoganWright: We'll need to implement this at the end of the argument list w/ a default argument in order to avoid breaking api.

Sources/Auth/Authentication/AuthMiddleware.swift
```diff
@@ -29,16 +32,24 @@ public class AuthMiddleware<U: User>: Middleware {
             )
         }
     }
+    
+    public convenience init(
```
LoganWright: I see you added this, so api should be preserved but I think it'd be easier to keep argument order the same and tack on the default as above. That also allows us to not mark this convenience which might cause some weirdness if people are getting fancy with it.

Tests/AuthTests/Utilities/AuthMiddlewareResponser.swift
```diff
@@ -0,0 +1,9 @@
+import HTTP
+import Turnstile
+
+class AuthMiddlewareResponser : Responder {
```
LoganWright: small thing, `:` up against `AuthMiddlewareResponder`

Sources/Auth/Authentication/AuthMiddleware.swift
sroebert: I agree this would have been better, however I always prefer having a closure argument at the end, as this allows for nicer usage of the method. This way you can still write:

```swift
let middleware = AuthMiddleware(user: User.self, cache: drop.cache, cookieName: authCookieName) { value in
  return Cookie(
    name: authCookieName,
    value: value,
    expires: Date().addingTimeInterval(24 * 60 * 60),
    secure: drop.environment == .production,
    httpOnly: true)
}
```

PR #664, opened 2016-10-25
Fix `Response had no 'Content-Type' header. when request files in Public folder

honghaoz:
#### Issue: 
When requesting a file in `/Public` folder, if this file is cached, the HTTP response had no 'Content-Type' header.` set.

#### Problem
FilleMiddle returns a `Response` without setting `Content-Type` header if the requested file has been cached.

#### Fix
This PR set `Content-Type` header regardless of the file is new or cached.

### DIFF
--- a/Sources/Vapor/File/FileMiddleware.swift
+++ b/Sources/Vapor/File/FileMiddleware.swift
@@ -40,7 +40,15 @@ public final class FileMiddleware: Middleware {
             // Generate ETag value, "HEX value of last modified date" + "-" + "file size"
             let fileETag = "\(modifiedAt.timeIntervalSince1970)-\(fileSize.intValue)"
             headers["ETag"] = fileETag
-            
+			
+			// Set Content-Type header based on the media type
+			if
+				let fileExtension = filePath.components(separatedBy: ".").last,
+				let type = mediaTypes[fileExtension]
+			{
+				headers["Content-Type"] = type
+			}
+			
             // Check if file has been cached already and return NotModified response if the etags match
             if fileETag == request.headers["If-None-Match"] {
                 return Response(status: .notModified, headers: headers, body: .data([]))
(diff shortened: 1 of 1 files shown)


LoganWright:
@honghaoz thank you for the great bug fix!

ogres:
Actually, the old implementation was correct here.

We do not have to set Content-Type for the .notModified, because we do not return any content.

Other web-servers ( nginx/apache ... ) do the same.

See https://tools.ietf.org/html/rfc7232#section-4.1

> The server generating a 304 response MUST generate any of the
   following header fields that would have been sent in a 200 (OK)
   response to the same request: Cache-Control, Content-Location, Date,
   ETag, Expires, and Vary.

>    Since the goal of a 304 response is to minimize information transfer
   when the recipient already has one or more cached representations, a
   sender SHOULD NOT generate representation metadata other than the
   above listed fields unless said metadata exists for the purpose of
   guiding cache updates (e.g., Last-Modified might be useful if the
   response does not have an ETag field).

LoganWright:
@ogres thanks, I moved this to #683

honghaoz:
Great, which means we need to stop logging warning of `Response had no 'Content-Type' header` if the status is `.notModified`

ogres:
I think its not just 304, but 301, 1xx ... responses where content is not expected or is empty.

https://tools.ietf.org/html/rfc7231#section-3.1.1.5 says, that there should be Content-Type for the responses with body, maybe we can ignore it for the responses without body?

>  A sender that generates a message containing a payload body SHOULD
   generate a Content-Type header field in that message unless the
   intended media type of the enclosed representation is unknown to the
   sender.

LoganWright:
@honghaoz @ogres and any others, if we have more comments on this, let's move them to #683 so they don't get lost in the shuffle 👍

Issue #691, opened 2016-11-06
Allow overwriting of slug name chosen for resources


kdawgwilk:
# Feature, Enhancement, or Optimization

# Overwrite slug name for resources

* Author(s): [Kaden Wilkinson](https://github.com/kdawgwilk)

## Introduction

I would like to choose what the slug name is used when creating resource routes so that I can extend those routes easier.

## Motivation

Here is a sample of what I am trying to accomplish:

```swift
let drop = Droplet()

let api: RouteGroup  = drop.grouped("api")
let v1: RouteGroup = api.grouped("v1")

// /users
v1.resource("users", UserController())

let user: RouteGroup = v1.grouped("users/:user_id")

// /users/:user_id/groups
user.resource("groups", GroupController())

let group: RouteGroup = v1.grouped("users/:user_id/groups/:group_id")

// /users/:user_id/groups/:group_id/tasks
group.resource("tasks", TaskController())

// /users/:user_id/groups/:group_id/rewards
group.resource("rewards", RewardController())
```

but I keep getting these warnings in the console:

```
[WARNING] Mismatched Slugs:
Attempted to overwrite :id with :user_id
Please use the same slug name for all routes on shared branch
```

## Proposed solution

Currently the slug name is hard coded in the function for creating a new resource.  I propose to move this to a default parameter so that it can be overridden if needed.

## Code snippets

This would allow you to then say something similar to to this:

```swift
let drop = Droplet()

let api: RouteGroup  = drop.grouped("api")
let v1: RouteGroup = api.grouped("v1")

// /users
v1.resource("users", ":user_id", UserController())

let user: RouteGroup = v1.grouped("users/:user_id")

// /users/:user_id/groups
user.resource("groups", ":group_id", GroupController())

let group: RouteGroup = v1.grouped("users/:user_id/groups/:group_id")

// /users/:user_id/groups/:group_id/tasks
group.resource("tasks", ":task_id", TaskController())

// /users/:user_id/groups/:group_id/rewards
group.resource("rewards", ":reward_id", RewardController())
```

## Impact

By using a default parameter it should not effect existing code but support greater flexibility.

## Alternatives considered

This was inspired as a workaround by my struggling with the previous issue #676

LoganWright:
@kdawgwilk this looks good, I think the trick here is that the controller would need to know about `task_id` somehow. Don't have a solution yet, we could consider passing it in or something perhaps, just laying it here so we know the issue.

kdawgwilk:
[...]

PR #693, opened 2016-11-07
Add support for overwriting slugName in resource

kdawgwilk:
This is related to the issue #691 and the solution I came up with with a little experimentation.  I don't see this as the final draft or the best solution but figured a PR would be the best place to get some feedback so others could benefit.

### DIFF
--- a/Sources/Vapor/Routing/Resource.swift
+++ b/Sources/Vapor/Routing/Resource.swift
@@ -45,12 +45,12 @@ public protocol ResourceRepresentable {
 }
 
 extension RouteBuilder where Value == Responder {
-    public func resource<Resource: ResourceRepresentable>(_ path: String, slugName: String = "id", _ resource: Resource) {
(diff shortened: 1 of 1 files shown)


kdawgwilk:
I really want to use an unlabeled parameter for the slugName but generics are struggling to resolve correctly when I try.  Also I think I am missing the default on the bottom resource function.

kdawgwilk:
Or maybe its better to just use the resource name and attach a `_id` to the end of it so it doesn't end up ambiguous?

LoganWright:
@kdawgwilk I think this is a good start, we'd also need to go into generated code so that the generated slug names there also don't interfere w/ here.

I'm trying to think of some type of alternate system, but don't have one immediately.

LoganWright:
@kdawgwilk sorry this got stale, I'm updating now, I think this is safe to merge.

@tannernelson can you take a quick look regarding whether this would interfere w/ generated code? I believe I was mistaken thinking it would, but would like your second look.

PR #733, opened 2016-12-01
fix vapor Accept-Language parsing

yudun1989:
the browser could accept multiple languages separated by ','  
just like:

Accept-Language: da, en-gb;q=0.8, en;q=0.7

I've choosed the first matched localization file in vapor locales to return

ref:
https://www.w3.org/International/questions/qa-lang-priorities


yudun1989:
duplicated....closed

PR #741, opened 2016-12-03
Add a default 'GET [resource]/new' and 'GET [resource]/edit' routes for ResourceRepresentable Controllers

### LINKED ISSUE #740: Add a default 'GET [resource]/new' route for ResourceRepresentable Controllers
# Add a default 'GET [resource]/new' route for ResourceRepresentable Controllers

* Author: [Gianluca Tranchedone](https://github.com/gtranchedone)

## Introduction

In order to have a consolidated way to display a HTML-based view to create a new entity for a Model object, a new `GET` route to `[resource]/new` could be automatically generated for declared `resources`.

## Motivation

As Swift and Vapor become more and more popular to build web apps there will be an increased demand for conventions and utilities to build HTML driven apps. Right now Vapor rightfully seems very much designed to build server-side apps to drive mobile apps and web apps that make heavy use of front-end technologies that mostly consume HTTP-based APIs.

The proposed change helps developers build web pages that display forms and/or other constructs to create a record, edit its values and delete the record (cancelling its creation) before the record is saved on a database.

An example of such flow is this very issue and its related Pull-Request as it's being created on github.com. An intermediate step between not having a record and having one committed to a database is often a necessity when working on HTML-based web apps, or at least it may offer a better user experience.

Additionally, this change would make Vapor `resources` match [Rails](https://github.com/rails/rails) resources. Besides this being a possible motivation on its own, an underlying motivation is that Swift and Vapor are likely going to be used, at least initially, by developers with Swift experience on other platforms. As on such platforms, many of the tools that help developers get their job done are built using [Ruby](http://ruby-lang.org) - see [CocoaPods](http://cocoapods.org) and [Fastlane](http://fastlane.tools), for example - they are also more likely to have developed web apps using Rails, or otherwise be familiar with it. In that case a matching behavior between the two web frameworks may help developers use Vapor.

If this feature was to be integrated into the main code base, it would also make sense to match the conventional method names for the different routes with the Rails equivalent, to make them easier to understand. As a consequence the properties `store` and `modify` in the `Resource` object would be renamed `create` and `update` respectively.

## Proposed solution

The proposed feature can be easily implemented by editing the `Resource` class to accept a new parameter and by updating the `resource` method on `RouteBuilder`s to create the new route by default. See below for details.

## Code snippets

With the proposed solution, `drop.resource("items")` would generate a `/items/new` path that directs to the `Resource.new` action of the returned `ResourceRepresentable`.

See PR #741 .

## Impact

The proposed changes as implemented in the PR above have no impact on the existing apps.

## Alternatives considered

I've considered not building this feature in favor of a manually configured route for the `new` action
and having a separate controller show the contents of the potential `new` view as part of `index`.

gtranchedone:
Fixes #740

### DIFF
--- a/Sources/Vapor/Routing/Resource.swift
+++ b/Sources/Vapor/Routing/Resource.swift
@@ -10,6 +10,7 @@ public final class Resource<Model: StringInitializable> {
     public var new: SimpleRequest?
     public var create: SimpleRequest?
     public var show: ItemRequest?
+    public var edit: ItemRequest?
     public var update: ItemRequest?
     public var replace: ItemRequest?
     public var destroy: ItemRequest?
@@ -22,8 +23,9 @@ public final class Resource<Model: StringInitializable> {
         new: SimpleRequest? = nil,
         create: SimpleRequest? = nil,
         show: ItemRequest? = nil,
-        replace: ItemRequest? = nil,
+        edit: ItemRequest? = nil,
         update: ItemRequest? = nil,
+        replace: ItemRequest? = nil,
         destroy: ItemRequest? = nil,
         clear: SimpleRequest? = nil,
         aboutItem: ItemRequest? = nil,
@@ -33,8 +35,9 @@ public final class Resource<Model: StringInitializable> {
         self.new = new
         self.create = create
         self.show = show
-        self.replace = replace
+        self.edit = edit
         self.update = update
+        self.replace = replace
         self.destroy = destroy
         self.clear = clear
         self.aboutItem = aboutItem
@@ -59,8 +62,8 @@ public extension Resource {
         self.init(index: index,
                   create: store,
                   show: show,
-                  replace: replace,
                   update: modify,
+                  replace: replace,
                   destroy: destroy,
                   clear: clear,
                   aboutItem: aboutItem,
@@ -84,14 +87,14 @@ extension RouteBuilder where Value == Responder {
         var itemMethods: [Method] = []
         var simpleMethods: [Method] = []
 
-        func item(_ method: Method, _ item: Resource<Model>.ItemRequest?) {
+        func item(_ method: Method, subpath: String? = nil, _ item: Resource<Model>.ItemRequest?) {
             guard let item = item else {
                 return
             }
 
             itemMethods += method
 
-            self.add(method, path, ":id") { request in
+            let closure: (HTTP.Request) throws -> HTTP.ResponseRepresentable = { request in
                 guard let id = request.parameters["id"]?.string else {
                     throw Abort.notFound
                 }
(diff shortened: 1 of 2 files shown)


Sources/Vapor/Routing/Resource.swift
```diff
@@ -3,42 +3,75 @@ import HTTP
 import TypeSafeRouting
 
 public final class Resource<Model: StringInitializable> {
-    public typealias Multiple = (Request) throws -> ResponseRepresentable
```
LoganWright: @gtranchedone unfortunately, we can't rename public types w/o breaking semver, let's revert these 👍

Sources/Vapor/Routing/Resource.swift
```diff
@@ -3,42 +3,75 @@ import HTTP
...
         self.clear = clear
         self.aboutItem = aboutItem
         self.aboutMultiple = aboutMultiple
     }
 }
 
+public extension Resource {
+
+    @available(*, deprecated: 1.3, message: "Use init(index:new:create:show:replace:update:destroy:clear:aboutItem:aboutMultiple:)")
+    convenience public init(
```
LoganWright: Having both of these w/ defaults, are there any ambiguous errors?

Sources/Vapor/Routing/Resource.swift
```diff
@@ -52,16 +85,16 @@ extension RouteBuilder where Value == Responder {
 
     public func resource<Model: StringInitializable>(_ path: String, _ resource: Resource<Model>) {
         var itemMethods: [Method] = []
-        var multipleMethods: [Method] = []
+        var simpleMethods: [Method] = []
```
LoganWright: let's keep the original naming for this one as well.

Sources/Vapor/Routing/Resource.swift
```diff
@@ -124,8 +173,8 @@ extension RouteBuilder where Value == Responder {
         _ path: String,
         _ type: Model.Type = Model.self,
         closure: (Resource<Model>) -> ()
-    ) {
-        let resource = Resource<Model>()
+        ) {
```
LoganWright: rm whitespace here

Sources/Vapor/Routing/Resource.swift
gtranchedone: Indeed, I'll make sure no ambiguity remains.

Sources/Vapor/Routing/Resource.swift
gtranchedone: Will do

LoganWright:
@gtranchedone closing this for now, just trying to keep an eye on what's new. Reopen this when the changes are updated and we'll get it merged 👍

PR #731, opened 2016-11-29
Middleware for CORS support [WIP]

nickskull:
This is the first draft of the middleware for proper CORS support in responses. 

It's still WIP as there are tests missing and there might be some slight changes, but I'd love some feedback from anyone who has more experience with handling CORS.

I've tried to follow the diagram [located here](https://www.html5rocks.com/static/images/cors_server_flowchart.png) as closely as possible. 

If there is no feedback I will finish tests and verify that it works on Linux and it should be ready for merging! 🎉 

## What is it?

CORS is a standard mechanism that can be used by all browsers for implementing cross-domain requests. The spec defines a set of headers that allow the browser and server to communicate about which requests are (and are not) allowed. CORS continues the spirit of the open web by bringing API access to all.  

[http://enable-cors.org/](http://enable-cors.org/)

## Impact on existing code  

None as this is a purely additive feature.

PR #753, opened 2016-12-14
Cors

LoganWright:
Closing #731

### DIFF
--- a/Sources/Vapor/CORS/CORSConfiguration.swift
+++ b/Sources/Vapor/CORS/CORSConfiguration.swift
@@ -84,6 +84,12 @@ public struct CORSConfiguration {
 }
 
 extension CORSConfiguration: ConfigInitializable {
+
+    /// Creates the CORS configuration of the the Vapor's settings config dictionary.
+    /// This enables setting up CORS using a .json configuration file in the project.
+    ///
+    /// - Parameter config: The settings config dictionary that should be used to extract settings.
+    /// - Throws: Node extraction errors, if extraction fails.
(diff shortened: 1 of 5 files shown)


skreutzberger:
Does the pull request contain tests or an example? I am currently trying to compare it with my own old [CORS implementation](https://github.com/SwiftyBeaver/SBWebsite/blob/master/Sources/SBWebsite/Middleware/CorsMiddleware.swift) (which also has no tests but was never meant to be part of Vapor) but I do not immediately which one I should take now.

@LoganWright Maybe we can push a bit the need for tests, examples and a page or section in the documentation about a feature or change with each pull request? In that way contributors get used to not just only contribute code but also docs or examples which would help the general userbase.

nickskull:
@skreutzberger No, I've started writing tests yesterday, so they're not yet ready, but my plan is to have it done today or tomorrow.

As for example, I think the easiest will be to update the docs with section on how to add CORS support.

@LoganWright Please don't merge before I finish the tests 😇

nickskull:
I've added some very basic tests, have a look when you have time :)

nickskull:
@LoganWright I am not entirely sure what is going on with codecov. Am I missing any docs anywhere or is it codecov just being annoying? :)  Otherwise this PR should be ready for review.

I'm going to be making a PR to the docs with the changes soon so that people know how to use this :)

LoganWright:
@nickskull the code cov stuff is in the first comment, they just update the same comment each time, and factor in the change in code cov %. Updated to master so travis can run again, then let's merge this.

PR #764, opened 2016-12-20
Added LoginRedirectMiddleware

Ponyboy47:
Works just like the protect middleware, except if users are unauthenticated this automatically redirects to the login page instead of throwing an error. 

I've found this middleware to be very useful for my servers. Makes it really easy to set up routes that users need to be authenticated for, but send them to the login page automatically instead of making the user manually enter the login page url.

Could probably be renamed to just "RedirectMiddleware" though

### DIFF
--- a/Sources/Auth/Authentication/LoginRedirectMiddleware.swift
+++ b/Sources/Auth/Authentication/LoginRedirectMiddleware.swift
@@ -1,0 +1,16 @@
+import HTTP
+
+public class LoginRedirectMiddleware: Middleware {
+    public let loginRoute: String
+    public init(loginRoute: String) {
+        self.loginRoute = loginRoute
+    }
+    
+    public func respond(to request: Request, chainingTo next: Responder) throws -> Response {
+        guard try request.subject().authenticated else {
+            return Response(redirect: loginRoute)
+        }
+
+        return try next.respond(to: request)
+    }
+}


LoganWright:
@Ponyboy47 thanks for the contribution, I think this is kind of cool, but I'd like to get some other voices to weigh in on it before a merge.

natebird:
I like this. I was going to suggest genericizing it but it does serve this very specific purpose for unauthenticated users to go to a route which is usually the login page - so I'm fine with the name too.

Casperhr:
We did exactly this middleware also when building admin panel.
The only thing it misses is a flash message to session

I only used it to store user on request.storage so it was accessable in all views

sroebert:
Generally you want to redirect back to the original page after login, maybe that would be a nice addition.

bygri:
When returning the redirect to `loginRoute`, the route of the original request should be appended to the login route so that @sroebert's suggestion can be implemented.

This is what Django does: a request of `GET /protected/` redirects to `GET /login/?next=/protected/`. The standard implementation of the login form then returns a redirect to the value of `next` if set.

This means that somebody accessing a protected resource will be seamlessly directed there after a successful login.

adriencanterot:
Hi, great idea ! Would there be a way to handle the error thrown by ProtectMiddleware in a way that it would redirect to the login page though ?

PR #772, opened 2016-12-28
AbortError: error codes, messages and metadata

BrettRToomey:
This adds support for custom errors with error codes and metadata (without breaking public API). Thusly, making it easier for Abort and Logging middleware to handle default and custom errors. As well as making it possible for them to determine whether or not they need to report the errors to some kind of bug tracking/internal system.

@LoganWright @tannernelson should I/we deprecate the old error enumeration or just keep the default implementation as is. I think it's fine as is, but I don't know what you have in mind.

@Casperhr Did I miss anything you were looking for? I was thinking the `shouldReport: Bool` would just go into the `metadata` node or that we could just make our own conformation to `AbortError` that has the `shouldReport` field. I prefer the latter, honestly.

### DIFF
--- a/Sources/Vapor/Error/Abort.swift
+++ b/Sources/Vapor/Error/Abort.swift
@@ -1,4 +1,24 @@
 import HTTP
+import Node
+
+/**
+    Represents errors that can be thrown in any Vapor closure.
+    Then, these errors can be caught in `Middleware` to give a
+    desired response.
+ */
+public protocol AbortError: Error {
+    /// Textual representation on the error.
+    var message: String { get }
+    
+    /// An integer representation of the error.
+    var code: Int { get }
+    
+    /// The HTTP status code to return.
+    var status: Status { get }
+    
+    /// `Optional` metadata.
+    var metadata: Node? { get }
+}
 
 /**
     A handful of standard errors that can be thrown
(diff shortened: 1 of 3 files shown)


BrettRToomey:
@LoganWright @tannernelson should I add test coverage for the default case?

Sources/Vapor/Error/AbortMiddleware.swift
```diff
@@ -23,31 +23,26 @@ public class AbortMiddleware: Middleware {
...
         if request.accept.prefers("html") {
-            return ErrorView.shared.makeResponse(status, message)
+            return ErrorView.shared.makeResponse(error.status, error.message)
         }
 
         let json = try JSON(node: [
             "error": true,
-            "message": "\(message)"
+            "message": "\(error.message)",
+            "code": error.code,
```
BrettRToomey: @Casperhr Done.

BrettRToomey:
Hmmm... It looks like the codecov tests are failing to compile on the last step of the macOS Travis build. Is this normal or not?

Sources/Vapor/Error/Abort.swift
```diff
@@ -1,4 +1,24 @@
...
+    var message: String { get }
+    
+    /// An integer representation of the error.
+    var code: Int { get }
+    
+    /// The HTTP status code to return.
+    var status: Status { get }
+    
+    /// `Optional` metadata.
+    var metadata: Node? { get }
```
LoganWright: should we provide a default for this? ignore if one exists later.

Sources/Vapor/Error/AbortMiddleware.swift
```diff
@@ -23,31 +23,26 @@ public class AbortMiddleware: Middleware {
...
-            return try AbortMiddleware.errorResponse(request, .internalServerError, "Something went wrong")
-        } catch Abort.custom(let status, let message) {
-            return try AbortMiddleware.errorResponse(request, status, message)
+        } catch let error as AbortError {
+            return try AbortMiddleware.errorResponse(request, error)
         }
     }
 
-    public static func errorResponse(_ request: Request, _ status: Status, _ message: String) throws -> Response {
+    public static func errorResponse(_ request: Request, _ error: AbortError) throws -> Response {
```
LoganWright: can't change this API, let's keep original and then add this signature as a new function that calls the other internally

Sources/Vapor/Error/Abort.swift
Casperhr: internal server error should be default

BrettRToomey: A default for which? The protocol or the metadata field?

BrettRToomey:
Making a quick change and cancelled the build.

BrettRToomey:
@LoganWright Ready for you, good sir :)

PR #774, opened 2016-12-30
Retain cycle between Request and Auth Header Helper

fedefrappi:
When using Auth Header the Helper class holds a strong reference to Request, which itself holds a strong reference to Helper through the storage dictionary.
It seems that this causes every request to leak, this is mostly noticeable with large multipart requests. 

This pull request should break the retain cycle and avoid the leak.

### DIFF
--- a/Sources/Auth/Authentication/Header/Request+Authorization.swift
+++ b/Sources/Auth/Authentication/Header/Request+Authorization.swift
@@ -2,13 +2,13 @@ import HTTP
 import Turnstile
 
 public final class Helper {
-    public let request: Request
+    public weak var request: Request?
     public init(request: Request) {
         self.request = request
     }
 
     public var header: Authorization? {
-        guard let authorization = request.headers["Authorization"] else {
+        guard let authorization = request?.headers["Authorization"] else {
             return nil
         }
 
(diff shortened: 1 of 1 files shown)


Sources/Auth/Authentication/Header/Request+Authorization.swift
```diff
@@ -2,31 +2,31 @@ import HTTP
 import Turnstile
 
 public final class Helper {
-    public let request: Request
+    public weak var request: Request?
```
LoganWright: This would unfortunately constitute a breaking change, anyone have ideas on how to maintain API here without introducing a crash scenario.

vzsg: I don't think it's possible to change this without breaking the API. Perhaps instead of changing this, the cycle could be broken by removing the cached Helper instance?

LoganWright:
@fedefrappi, @vzsg  I inverted the reference preservation so that we could maintain existing apis w/o relying on a reference cycle. I pulled your commits in as well @fedefrappi.

Please take a look at #777 and let me know if it will resolve your issue 👍

fedefrappi:
Nice trick @LoganWright! Thank you!

PR #779, opened 2017-01-06
Add in logger in non-deprecated init()

cwoloszynski:
This change allows users of the non-deprecated init() of Droplet to pass in an optional logger.  This will allow development of loggers that capture ALL logged output, including items that occur during the initialization of the logger.  Currently, all logs currently always go to the ConsoleLogger during initialization and only after this can the logger be replaced for subsequent logging.

### DIFF
--- a/Sources/Vapor/Droplet/Droplet.swift
+++ b/Sources/Vapor/Droplet/Droplet.swift
@@ -163,7 +163,8 @@ public class Droplet {
         workDir workDirProvided: String? = nil,
         environment environmentProvided: Environment? = nil,
         config configProvided: Settings.Config? = nil,
-        localization localizationProvided: Localization? = nil
+        localization localizationProvided: Localization? = nil,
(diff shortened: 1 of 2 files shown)


Sources/Vapor/Droplet/Droplet.swift
```diff
@@ -163,17 +163,23 @@ public class Droplet {
...
 
-        // logging is needed for emitting errors
         let terminal = Terminal(arguments: arguments)
-        let log = ConsoleLogger(console: terminal)
+        if let provided = logProvided  {
+            self.log = provided
+        } else {
+            // logging is needed for emitting errors
+            self.log = ConsoleLogger(console: terminal)
+            self.log.enabled = LogLevel.all
```
LoganWright: omit this line, it doesn't do anything since we modify based on env. a few lines later

cwoloszynski: Done.

PR #784, opened 2017-01-08
Plug a tiny memory leak in FileManager.contentsOfDirectory

vzsg:
Nothing great to write home about, I was just playing around with Xcode's debugger, and found this leftover string.


vzsg:
Not at all, done.

vzsg:
#806 is where the real fun is at.

PR #793, opened 2017-01-13
Align the Vapor Log APIs to use StaticStrings

cwoloszynski:
In working to integrate XCGLogger into Vapor, the API for the LogProtocol and the API for XCGLogger diverge in a manner that cannot be integrated.  The XCGLogger uses `StaticString` in the APIs for Filename and Functions to allow the compiler to better optimize the log output.

This PR introduces the changes to address this and improve the Vapor.LogProtocol to allow XCGLogger integration.

### DIFF
--- a/Sources/Vapor/Log/ConsoleLogger.swift
+++ b/Sources/Vapor/Log/ConsoleLogger.swift
@@ -33,8 +33,8 @@ public class ConsoleLogger: LogProtocol {
     public func log(
         _ level: LogLevel,
         message: String,
-        file: String = #file,
-        function: String = #function,
+        file: StaticString = #file,
+        function: StaticString = #function,
         line: Int = #line
     ) {
         if enabled.contains(level) {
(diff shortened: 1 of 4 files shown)


IanKeen:
For what it's worth, `StaticString` may not be around in Swift 4
image
https://github.com/apple/swift/blob/master/docs/StringManifesto.md

Joannis:
@IanKeen Vapor will probably need to be updated for Swift 4 anyways. So I don't think that's an issue

IanKeen:
Exactly, I'm saying why introduce `StaticString` when its likely to go, especially when theres no clear replacement..

`String` to `StaticString` seems like pre-optimization anyway :)

cwoloszynski77:
I don't read that quote as saying it will go away. Just that it needs to align to the rest of the changes to String.

cwoloszynski:
I think it is clear (at least to me) that the community does not want to move this forward.  I'll fork XCGLogger and address it on their side and see if I can convince them to drop the StaticString from their APIs.

PR #794, opened 2017-01-13
Better error handling of CORSConfiguration init from config json

nickskull:


Sources/Vapor/CORS/CORSConfiguration.swift
```diff
@@ -91,32 +110,49 @@ extension CORSConfiguration: ConfigInitializable {
...
+        do {
+            let originString: String = try cors.extract("allowedOrigin")
+            switch originString {
+            case "all", "*": self.allowedOrigin = .all
+            case "none", "": self.allowedOrigin = .none
+            case "origin", "Origin": self.allowedOrigin = .originBased
+            default: self.allowedOrigin = .custom(originString)
+            }
+        } catch {
+            throw CORSConfigurationError.missingRequiredConfigurationKey("allowedOrigin")
```
LoganWright: should we have a default instead to `*` or something?

Sources/Vapor/CORS/CORSConfiguration.swift
```diff
@@ -91,32 +110,49 @@ extension CORSConfiguration: ConfigInitializable {
...
+        }
 
         // Get allowed headers
-        let headersArray: [String] = try cors.extract("allowedHeaders")
-        self.allowedHeaders = headersArray.joined(separator: ", ")
+        do {
+            let headersArray: [String] = try cors.extract("allowedHeaders")
+            self.allowedHeaders = headersArray.joined(separator: ", ")
+        } catch {
+            throw CORSConfigurationError.missingRequiredConfigurationKey("allowedHeaders")
```
LoganWright: In general, I'm curious if we should opt for defaults over errors on this one.

nickskull: I don't think having defaults for the origin, headers and methods makes sense if you choose to use the JSON config. Same was as they are non-optional in the other initialiser and you are required to provide them. It seems to me similar to having MySQL config automatically put in localhost as the db IP if none is present in the JSON.

If someone wants to use the defaults then I think they should use `CORSConfiguration.default`.

Casperhr:
I argued for the same thing

If you want to use a config and you fail to set it up. It should just crash hard with a descriptive message

Sources/Vapor/CORS/CORSConfiguration.swift
Casperhr: I argued for the same thing

If you want to use a config and you fail to set it up. It should just crash hard with a descriptive message

Sources/Vapor/CORS/CORSConfiguration.swift
nickskull: How would that work if you for example put in as allowed origin `http://google.com`? We can't match against that (and matching on http or similar seems a bit unsafe). So that's why I made it with default setting a custom allowed origin string.

nickskull: How would that work if you for example put in as allowed origin `http://google.com`? We can't match against that (and matching on http or similar seems a bit unsafe). So that's why I made it with default setting a custom allowed origin string.

PR #803, opened 2017-01-25
Modified the LoginRedirectMiddleware to support adding a cameFrom val…

Ponyboy47:
…ue to the URI query

I added default values to everything new in the initializer so that this wouldn't be a breaking change. Although, this does depend on a pull request I made to vapor/engine
https://github.com/vapor/engine/pull/93

### DIFF
--- a/Sources/Auth/Authentication/LoginRedirectMiddleware.swift
+++ b/Sources/Auth/Authentication/LoginRedirectMiddleware.swift
@@ -1,16 +1,76 @@
+import Foundation
 import HTTP
+import URI
 
-public class LoginRedirectMiddleware: Middleware {
+/**
+ LoginRedirectMiddleware
+
+ Middleware object used to automatically redirect users to a specific web page if they are not authenticated
+
+ Generally used to redirect to the site's login page. Optionally can redirect the user back to the page they were trying to access before they logged in.
+ */
+public class LoginRedirectMiddleware : Middleware {
     public let loginRoute: String
-    public init(loginRoute: String) {
+    public let cameFromRedirect: Bool
+    public let defaultRedirectRoute: String
+    public let ignoredCameFromRoutes: [HTTP.Method : [String]]
+
+    /**
+     Designated LoginRedirectMiddleware initializer
+
+     - Parameters:
+     - loginRoute: The absolute URI path to where users will be redirected if they are unauthenticated
+     - cameFromRedirect: Whether or not to redirect users back to the page they came from before authenticating (default = false)
+     - defaultRedirectRoute: The default absolute URI path to return users to after authenticating when the cameFrom value is removed since the route is a part of the ignoredCameFromRoutes array (default = "/")
+     - ignoredCameFromRoutes: Routes to avoid redirecting to after the user authenticates, such as the logout page. (default = [])
+     */
+    public init(loginRoute: String, cameFromRedirect: Bool = false, defaultRedirectRoute: String = "/", ignoredCameFromRoutes: [HTTP.Method : [String]] = [:]) {
         self.loginRoute = loginRoute
+        self.cameFromRedirect = cameFromRedirect
+        self.defaultRedirectRoute = defaultRedirectRoute
+        self.ignoredCameFromRoutes = ignoredCameFromRoutes
     }
-    
+
     public func respond(to request: Request, chainingTo next: Responder) throws -> Response {
-        guard try request.subject().authenticated else {
-            return Response(redirect: loginRoute)
+        // Make sure we aren't going to the loginRoute
+        if (request.uri.path != loginRoute) {
+            // Test to see if we're authenticated
+            guard try request.subject().authenticated else {
+                // Not authenticated, so build the redirect
+                let redirect = try Request(method: .get, uri: loginRoute)
+                if (cameFromRedirect) {
+                    // Add the cameFrom query to the redirect URI
+                    redirect.uri.addQuery(withKey: "cameFrom", value: request.uri.description)
+                    // Remove the cameFrom query if it happens to be one of the ignored routes
+                    for (method, routes) in ignoredCameFromRoutes {
+                        for route in routes {
+                            if (request.method == method && request.uri.path == route) {
+                                redirect.uri.removeQuery(forKey: "cameFrom")
+                                break
+                            }
+                        }
+                    }
+                    // If we have no cameFrom query, use the defaultRedirectRoute
+                    if (redirect.uri.getQueryValue(forKey: "cameFrom") == nil) {
+                        redirect.uri.addQuery(withKey: "cameFrom", value: defaultRedirectRoute)
+                    }
+                }
+                // Redirect to the loginRoute
+                return Response(redirect: redirect.uri.description)
+            }
         }
 
+        // We are authenticated, so continue on to the next middleware
         return try next.respond(to: request)
     }
 }
+
+extension Request {
+    public func cameFrom() throws -> Request? {
+        if let cameFrom = self.uri.getQueryValue(forKey: "cameFrom") {
+            self.uri = try URI(cameFrom)
+        }
+        return self
+    }
+}
+


Sources/Auth/Authentication/LoginRedirectMiddleware.swift
```diff
@@ -1,16 +1,76 @@
...
+        self.defaultRedirectRoute = defaultRedirectRoute
+        self.ignoredCameFromRoutes = ignoredCameFromRoutes
     }
-    
+
     public func respond(to request: Request, chainingTo next: Responder) throws -> Response {
-        guard try request.subject().authenticated else {
-            return Response(redirect: loginRoute)
+        // Make sure we aren't going to the loginRoute
+        if (request.uri.path != loginRoute) {
```
LoganWright: I think in general this code block is a little difficult to parse through. I believe we can clean up quite a bit by reworking some of the guards/ifs

```swift
// check if already at loginRoute
guard request.uri.path != loginRoute else { return try next.respond(to: request) }
// check if user is already authorized
guard try !request.subject().authenticated else { return try next.respond(to: request) }

// not logging in and not authenticated, so redirect
let redirect = try Request(method: .get, uri: loginRoute)
// ....
```

Sources/Auth/Authentication/LoginRedirectMiddleware.swift
```diff
@@ -1,16 +1,76 @@
...
+ LoginRedirectMiddleware
+
+ Middleware object used to automatically redirect users to a specific web page if they are not authenticated
+
+ Generally used to redirect to the site's login page. Optionally can redirect the user back to the page they were trying to access before they logged in.
+ */
+public class LoginRedirectMiddleware : Middleware {
     public let loginRoute: String
-    public init(loginRoute: String) {
+    public let cameFromRedirect: Bool
```
LoganWright: Not crazy about the `cameFrom` naming convention, I'd prefer to call this `redirectOrigin` and `ignoredOriginRequests`

Sources/Auth/Authentication/LoginRedirectMiddleware.swift
```diff
@@ -1,16 +1,76 @@
...
+    public let defaultRedirectRoute: String
+    public let ignoredCameFromRoutes: [HTTP.Method : [String]]
```
LoganWright: This is confusing because it feels like this should be ignored from redirect, but they aren't ignored later on, we still redirect.

Sources/Auth/Authentication/LoginRedirectMiddleware.swift
```diff
@@ -1,16 +1,76 @@
...
+                    // Remove the cameFrom query if it happens to be one of the ignored routes
+                    for (method, routes) in ignoredCameFromRoutes {
+                        for route in routes {
+                            if (request.method == method && request.uri.path == route) {
+                                redirect.uri.removeQuery(forKey: "cameFrom")
+                                break
+                            }
+                        }
+                    }
+                    // If we have no cameFrom query, use the defaultRedirectRoute
```
LoganWright: This code block is confusing to me, I think we can remove `defaultRedirectRoute` entirely. It adds API that needs to be maintained without adding anything for end user. If it's replaced at `nil`, end user can just use their own value.

Sources/Auth/Authentication/LoginRedirectMiddleware.swift
```diff
@@ -1,16 +1,76 @@
...
+
     public func respond(to request: Request, chainingTo next: Responder) throws -> Response {
-        guard try request.subject().authenticated else {
-            return Response(redirect: loginRoute)
+        // Make sure we aren't going to the loginRoute
+        if (request.uri.path != loginRoute) {
+            // Test to see if we're authenticated
+            guard try request.subject().authenticated else {
+                // Not authenticated, so build the redirect
+                let redirect = try Request(method: .get, uri: loginRoute)
```
LoganWright: Creating a request here just to get the `uri` is a little odd. Let's just make a URI and add it that way.

Sources/Auth/Authentication/LoginRedirectMiddleware.swift
```diff
@@ -1,16 +1,76 @@
...
+                if (cameFromRedirect) {
+                    // Add the cameFrom query to the redirect URI
+                    redirect.uri.addQuery(withKey: "cameFrom", value: request.uri.description)
+                    // Remove the cameFrom query if it happens to be one of the ignored routes
+                    for (method, routes) in ignoredCameFromRoutes {
+                        for route in routes {
+                            if (request.method == method && request.uri.path == route) {
+                                redirect.uri.removeQuery(forKey: "cameFrom")
```
LoganWright: In general, it's preferable to only add things if they need to as opposed to always adding and then removing if not needed. 

I think this is also confusing with the naming about what the behavior of the 'ignoredCameFromRoutes` should be. I'd expect any route in cameThrough to be ignored.

We could also check ignoredCameFromRoutes at the top of this function.

```
guard !ignoredOrigins.contains(request.uri.path) else { // pass onto next }
```

Sources/Auth/Authentication/LoginRedirectMiddleware.swift
```diff
@@ -1,16 +1,76 @@
...
     public func respond(to request: Request, chainingTo next: Responder) throws -> Response {
-        guard try request.subject().authenticated else {
-            return Response(redirect: loginRoute)
+        // Make sure we aren't going to the loginRoute
+        if (request.uri.path != loginRoute) {
+            // Test to see if we're authenticated
+            guard try request.subject().authenticated else {
+                // Not authenticated, so build the redirect
+                let redirect = try Request(method: .get, uri: loginRoute)
+                if (cameFromRedirect) {
```
LoganWright: Let's avoid parens on ifs/guards/fors to keep with the Vapor/Swift style 👍

Sources/Auth/Authentication/LoginRedirectMiddleware.swift
```diff
@@ -1,16 +1,76 @@
...
+                    if (redirect.uri.getQueryValue(forKey: "cameFrom") == nil) {
+                        redirect.uri.addQuery(withKey: "cameFrom", value: defaultRedirectRoute)
+                    }
+                }
+                // Redirect to the loginRoute
+                return Response(redirect: redirect.uri.description)
+            }
         }
 
+        // We are authenticated, so continue on to the next middleware
```
LoganWright: this comment is no longer applicable

Sources/Auth/Authentication/LoginRedirectMiddleware.swift
```diff
@@ -1,16 +1,76 @@
...
         return try next.respond(to: request)
     }
 }
+
+extension Request {
+    public func cameFrom() throws -> Request? {
```
LoganWright: also use `origin` naming here

Sources/Auth/Authentication/LoginRedirectMiddleware.swift
```diff
@@ -1,16 +1,76 @@
...
+        if let cameFrom = self.uri.getQueryValue(forKey: "cameFrom") {
+            self.uri = try URI(cameFrom)
```
LoganWright: This is a really unexpected side effect of this function, we shouldn't be changing self on a get function this way

Sources/Auth/Authentication/LoginRedirectMiddleware.swift
LoganWright: this function signature isn't accurate to the function

LoganWright:
@Ponyboy47 I'm going to close this for now, only because we've been losing track of what PRs need our attention. This is not a rejection, just closing it to remove it from our queue 👍 

Once you make some of the changes mentioned, go ahead and reopen! 🙌 

Thanks 😄

Ponyboy47:
Reopening since I made the requested improvements :)

https://github.com/Ponyboy47/vapor/commit/5b47f26a0305ca09455aeeee4dc4dbdc9776dcb3

PR #837, opened 2017-02-10
Add name to the cookie factory closure.

kgn:
This is to make it clear this name should be used.


LoganWright:
@kgn this is a great fix for some of the older behavior we have, but we've moved a lot of the ownership of this to things like sessions so this PR is no longer applicable. Thank you so much for your Vapor contributions!

kgn:
@LoganWright great!

PR #838, opened 2017-02-11
Allow Non-parameterised Queries

0xTim:
This PR allows queries to be set on the `request.query` dictionary, even if there are no parameters specified in the query - i.e. `https://example.com/info?allInfo&testData`

Previously, this would return an empty `query` dictionary. This PR sets the `Node` value of these keys to an empty string so they can easily be read from the `request`.

This PR also fixes the AuthTests - `httpOnly` should always be set to `true` for Authentication Cookies to prevent tokens being stolen via XSS.


Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
```diff
@@ -7,22 +7,30 @@ extension Node {
...
-                    }
-                }
-
-                var keyData = percentDecoded(token[0], nonEncodedTransform: replacePlus) ?? []
+                keyData = percentDecoded(token[0], nonEncodedTransform: replacePlus) ?? []
                 let valueData = percentDecoded(token[1], nonEncodedTransform: replacePlus) ?? []
 
-                var value: Node = .string(valueData.string)
+                value = .string(valueData.string)
+            } else if token.count == 1 {
```
vzsg: I think you should introduce an extra `Bool` parameter (default value `false`) to the initializer to enable this branch, because this method is also used for parsing x-www-form-urlencoded bodies, where missing the equals sign is invalid syntax (and the line should be at least ignored).

This line would become:

```swift
} else if token.count == 1 && allowEmptyValues {
...
}

```

And it should only be set to `true` in the `query` computed property.

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
```diff
@@ -3,26 +3,34 @@ import Core
...
+            
+            let replacePlus: (Byte) -> (Byte) = { byte in
+                if byte == .plus {
+                    return .space
+                } else {
+                    return byte
+                }
+            }
+            
+            var key = ""
```
LoganWright: this doesn't need to be here

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
```diff
@@ -3,26 +3,34 @@ import Core
...
+            var value:Node = .string("")
+            var keyData: Bytes?
```
LoganWright: let's use `let keyData: Bytes`

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
```diff
@@ -3,26 +3,34 @@ import Core
 import Node
 
 extension Node {
-    public init(formURLEncoded data: Bytes) {
+    public init(formURLEncoded data: Bytes, allowEmptyValues: Bool = false) {
         var urlEncoded: [String: Node] = [:]
 
         for pair in data.split(separator: .ampersand) {
+            
+            let replacePlus: (Byte) -> (Byte) = { byte in
```
LoganWright: while we're in here, let's declare this somewhere outside of the `for-loop` I don't think we need to create it every time

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
```diff
@@ -3,26 +3,34 @@ import Core
...
+                if byte == .plus {
+                    return .space
+                } else {
+                    return byte
+                }
+            }
+            
+            var key = ""
+            var value:Node = .string("")
```
LoganWright: `value: Node`

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
```diff
@@ -3,26 +3,34 @@ import Core
...
-
-                var keyData = percentDecoded(token[0], nonEncodedTransform: replacePlus) ?? []
+                keyData = percentDecoded(token[0], nonEncodedTransform: replacePlus) ?? []
                 let valueData = percentDecoded(token[1], nonEncodedTransform: replacePlus) ?? []
 
-                var value: Node = .string(valueData.string)
+                value = .string(valueData.string)
+            } else if token.count == 1 && allowEmptyValues {
+                keyData = percentDecoded(token[0], nonEncodedTransform: replacePlus) ?? []
+            }
```
LoganWright: you'll need `else { keyData = nil }` here

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
```diff
@@ -3,26 +3,34 @@ import Core
...
 
+            if var keyData = keyData {
```
LoganWright: let's guard

```swift
guard let keyData = keyData else { 
    self = .object(urlEncoded)
    return
}
```

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
```diff
@@ -37,7 +45,7 @@ extension Node {
                     keyIndicatedArray = true
                 }
 
-                let key: String = keyData.string
+                key = keyData.string
```
LoganWright: this'll be `let key = keyData.string` now

LoganWright:
Some good stuff in here, most comments just cleanup stuff 👍

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
0xTim: Do not remember adding this! But done

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
0xTim: 👍

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
0xTim: 👍

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
0xTim: 👍

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
0xTim: 👍

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
0xTim: 👍

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
0xTim: This doesn't match to the above comment - can't assign `nil` to a `Bytes` variable. Declare it as optional and use the `guard` to check it, or instead of `else { keyData = nil }` do `else { self = .object(urlEncoded); return` ?

Sources/Vapor/FormURLEncoded/StructuredData+FormURLEncoded.swift
0xTim: Adding the the optional query below, the original code is reassigning `keyData` on line 42, so it has to be a `var`?

0xTim:
Fixed the merge conflicts and made most of the changes requested. Only query was around declaring `keyData` with `let keyData: Bytes` - I've implemented it one way but you may want it done a different way. See the outdated comments for context

LoganWright:
Closing moved to #890

PR #846, opened 2017-02-15
Add ability to customize HTTP methods for resource requests.

Mordil:
This adds support for a user to configure the Vapor's HTTP method mapping for resource endpoints.

This gives users, like myself, who want to have item updates to be done via HTTP `PUT` rather than the default `PATCH`.

The user can configure it as such:

```swift
import Vapor

ResourceMethodMap.set(modify: .put)

let drop = Droplet()
let postCtrl = PostController()
drop.resource("posts", postCtrl)
```

Technically this can be customized on a per resource base, since the mapping is referenced at resource creation.

```swift
import Vapor

internal final class V1Collection: RouteCollection, EmptyInitializable {
    internal typealias Wrapped = HTTP.Responder

    internal required init() {}

    internal func build<Builder:RouteBuilder>(_ builder: Builder) where Builder.Value == Responder {
        // mounts all endpoints under the /v1 path
        let v1 = builder.grouped("v1")

        let postCtrl = PostController()
        v1.resource("posts", postCtrl)
    
        // moves endpoints to different methods
        let v2 = builder.grouped("v2")
        ResourceMethodMap.set(modify: .put, replace: .patch)
        v2.resource("posts", postCtrl)
    }
}
```

### DIFF
--- a/Sources/Vapor/Routing/Resource.swift
+++ b/Sources/Vapor/Routing/Resource.swift
@@ -44,6 +44,39 @@ public protocol ResourceRepresentable {
     func makeResource() -> Resource<Model>
 }
 
+/// Maintains the desired mapping for resource endpoints and their HTTP methods.
+/// - Note: Setting the values with set() will define the behavior for all resources.
+public final class ResourceMethodMap {
+    private(set) static var index: Method = .get
+    private(set) static var store: Method = .post
+    private(set) static var show: Method = .get
+    private(set) static var replace: Method = .put
+    private(set) static var modify: Method = .patch
+    private(set) static var destroy: Method = .delete
+    private(set) static var clear: Method = .delete
+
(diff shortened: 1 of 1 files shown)


LoganWright:
@Mordil in general, we're wary of singletons, and I'm concerned that this might be less than clear behavior to some users, particular working off of other's code or tutorials that might have differing behavior. 

Before we pull this in, I'd like to evaluate if this could be resolved with something like middleware that you modify, for example:

```
if request.method == .put {
    request.method = .patch
}
```

You could then group this:

```swift
let resourceGroups = drop.grouped(PutToPatchMiddleware)

resourceGroups.get("foo", FooController ...
```

Mordil:
That's a good point that I hadn't considered. I will adjust.

PR #879, opened 2017-03-06
Add --force option when reverting DB from command line

gtranchedone:
## Motivation

When tests are running, in the `XCTest.setup` method you may want to set up a Droplet connected to a test database (e.g. using a specific configuration in `MyApp/Config/test/`). However, you may also need to reset the DB in the `teardown` method. An easy way to do this is to revert all preparations to drop all tables.

Running `vapor run prepare --revert` or using the `Prepare` command in code, makes the console ask the user to confirm the reversion, hence blocking the tests. The `--force` option introduced by this PR ensures that, if passed, the tests aren't blocked as no confirmation will be required.

## Notes

I'm not sure how this is going to play with the removal of Vapor/Fluent in Vapor 2.0 at this point. I couldn't find similar code in the Fluent repository. However, I guess it can be discussed and something similar could be introduced in the next alpha of Fluent 2.0.

### DIFF
--- a/Sources/Vapor/Fluent/Prepare.swift
+++ b/Sources/Vapor/Fluent/Prepare.swift
@@ -9,6 +9,7 @@ public struct Prepare: Command {
 
     public let signature: [Argument] = [
         Option(name: "revert"),
+        Option(name: "force"),
     ]
 
     public let help: [String] = [
@@ -40,7 +41,7 @@ public struct Prepare: Command {
         }
 
         if arguments.option("revert")?.bool == true {
-            guard console.confirm("Are you sure you want to revert the database?", style: .warning) else {
+            guard arguments.option("force")?.bool == true || console.confirm("Are you sure you want to revert the database?", style: .warning) else {
                 console.error("Reversion cancelled")
                 return
             }


tanner0101:
Passing `-y` or `-n` will automatically say yes or no respectively to console prompts.

gtranchedone:
Well, it doesn't seem to work… 😄 

screen shot 2017-03-06 at 12 58 44

tanner0101:
@gtranchedone where are you getting the `console` property from? The `-y` is parsed at the console level, not the individual command level.

Something like this would work:

```swift
let drop = Droplet(arguments: ["vapor", "prepare", "--revert", "-y"])
drop.run()
```

or even 

```swift
let terminal = Terminal(arguments: ["vapor", "prepare", "--revert", "-y"])
Prepare(console: terminal, ...)
```

https://github.com/vapor/console/blob/88e7b2347636c534ef3e8e537ffd31e551e9a161/Sources/Console/Terminal/Terminal.swift#L138

gtranchedone:
@tanner0101 That method is part of an extension to `Droplet`.

The full code:

```swift
@testable import Vapor
@testable import Console

extension Droplet {
    
    static func dropletForTests() throws -> Droplet {
        let droplet = Droplet(arguments: ["dummy/path/", "prepare"],
                              environment: .test)
        return droplet
    }
    
    func setUp() throws {
        try runCommands()
    }
    
    func tearDown() throws {
        let p = Prepare(console: console,
                        preparations: preparations,
                        database: database)
        try p.run(arguments: ["--revert", "-y"])
    }
    
}

```

tanner0101:
Add `-y` here, should work.

```swift
Droplet(arguments: ["dummy/path/", "prepare", "-y"], ...
```

gtranchedone:
Although it's not obvious why, but this does work @tanner. Up to you if you want to keep the `force` option or close the PR.

LoganWright:
Since -y is supported, let's stick with that for now, although I personally think `--force` is kind of nice since it's a little more clear that we're going to be doing some more intense operations.

PR #901, opened 2017-03-13
Dynamic config

LoganWright:


LoganWright:
Made config easier to load from workDir and put above droplet, no longer relevant.

PR #922, opened 2017-03-21
fire sale -- everything must go

LoganWright:


LoganWright:
@tanner0101 technically wasn't a target, so nothing to remove 👍

PR #1010, opened 2017-05-16
[PROPOSAL] Sessions middleware improvements

0xTim:
This is to kick off a bit of a debate, but...

This PR adds the ability to set a value in the `request.storage` to change the expiry time of the cookie set in the `SessionsMiddleware`. It also passes the request in to the `CookieFactory` so that injected `CookieFactory`'s can set different values or use different storage keys.

Be interesting to see what people think!

One thing that should definitely be taken from this PR is the setting of `httpOnly` to `true`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ Packages
 .DS_Store
 *.xcodeproj
 Package.pins
+DerivedData/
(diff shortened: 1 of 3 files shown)


0xTim:
Cool, have moved it out so the `CookieFactory` gets the request. Also amended the test to show how to use it. Happy with the changes as they are now?

PR #1039, opened 2017-05-27
Fix FileMiddleware serving 0 byte responses

vzsg:
In the current state, the final version of FileMiddleware is completely broken, always serving 0 length files, as the file is closed before the first read operation.

Having the `defer { fclose(...) }` block inside the Response closure was not a mistake in my original PR. This PR moves it back where it belongs.

### DIFF
--- a/Sources/Vapor/File/FileMiddleware.swift
+++ b/Sources/Vapor/File/FileMiddleware.swift
@@ -61,18 +61,19 @@ public final class FileMiddleware: Middleware {
             guard let file = fopen(filePath, "r") else {
                 throw Abort.notFound
             }
-            defer {
-                fclose(file)
-            }
 
             // make copy of size for closure
             let chunkSize = self.chunkSize
 
             // return chunked response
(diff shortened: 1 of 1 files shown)


vzsg:
This is not the only solution, by the way. The cleanest option lifecycle-wise would be moving fopen inside the closure, but there's a race condition either way.

Sources/Vapor/File/FileMiddleware.swift
```diff
@@ -61,18 +61,19 @@ public final class FileMiddleware: Middleware {
...
-                fclose(file)
-            }
 
             // make copy of size for closure
             let chunkSize = self.chunkSize
 
             // return chunked response
             return Response(status: .ok, headers: headers, chunked: { stream in
-                var buffer = Array(repeating: 0, count: chunkSize)
+                defer {
```
LoganWright: Let's add a note here explaining the situation, so it doesn't get moved again

PR #1074, opened 2017-06-26
Fix Int.random on Linux

unitymind:
Important, on Linux `libc.random`:
> By default, this function always generates the same sequence of numbers each time you run the program. (This is usually desirable when debugging, or when comparing two different runs.) If you need to produce a different sequence on every run, you must seed rand by calling srand (see section srand) before the first call to rand, and make sure to use a different argument to srand each time.

And we should seed rand before.

The same issue in `Random` module: `Sources/Random/OSRandom.swift`.

Review, merge and modify `Random` module too :)

### DIFF
--- a/Sources/Vapor/Hash/Int+Random.swift
+++ b/Sources/Vapor/Hash/Int+Random.swift
@@ -1,11 +1,20 @@
+import Foundation
 import libc
 
 extension Int {
     /// Generates a random number between (and inclusive of)
-    /// the given minimum and maxiumum.
+    /// the given minimum and maximum.
+    static private var randomInitialized: Bool = false
+
     public static func random(min: Int, max: Int) -> Int {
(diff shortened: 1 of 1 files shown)


0xTim:
Would be good to see a simple test that runs on Linux to generate two random numbers and make sure they are different!

calebkleveter:
So the builds randomly fail? 😆

unitymind:
@0xTim Not easy to confirm via tests :) I am going to issue due database uniq key constrain violation (when tests runs in parallel or twice in line).

`srand` seed initialization requires call only once per whole cycle of application lifetime.

May be include test that assert that sequential (with minimal time execution gap)  seed number generation (which used for `srand`) is different?

LoganWright:
Moving this branch to #1075

PR #1075, opened 2017-06-29
Fix linux int random

LoganWright:


unitymind:
Great move to shared code via `Random` :)

PR #1099, opened 2017-08-02
Swift4

LoganWright:
There are a few warnings from 3 to 4, but this now fully builds on Ubuntu 16 w/ the latest Aug 1 snapshot.


LoganWright:
This is also building for me on Ubuntu14 now, so should be all set there 👍

hhanesand:
👏 👏 👏

PR #1155, opened 2017-09-13
Application helpers

Joannis:
- [x] Type safe controllers
- [x] HTTP Client
- [x] Type-safe HTTP Client
- [x] Session helpers


Sources/Development/Models/User.swift
```diff
@@ -3,7 +3,7 @@ import HTTP
 import Leaf
 import Vapor
 
-final class User: JSONCodable, ContentCodable, ResponseRepresentable {
+final class User: JSONCodable, ResponseRepresentable {
```
tanner0101: Ideally we could get rid of `JSONCodable` here in favor of just `Codable`

Sources/Vapor/Controller.swift
```diff
@@ -0,0 +1,112 @@
+import Core
+import Foundation
+import HTTP
+import Service
+import Routing
+
+public protocol ContentEncoder {
+    var type: MediaType { get }
+    func encodeBody<E: Encodable>(from entity: E) throws -> Body
```
tanner0101: i think `encode(_:to:)` might look nicer here

Sources/Vapor/Server.swift
```diff
@@ -3,6 +3,6 @@ import HTTP
 /// Servers are capable of binding to an address
 /// and subsequently responding to requests sent
 /// to that address.
-public protocol Server {
+public protocol HTTPServer {
```
tanner0101: what's the reasoning for this change? I think Vapor.Server is descriptive enough

Sources/Vapor/TypeSafeRequest.swift
```diff
@@ -0,0 +1,68 @@
+import Core
+import Service
+import HTTP
+
+/// An HTTP Request with a type-safe body
+public final class TypeSafeRequest<Content: Codable> {
```
tanner0101: I think the functionality of this class could be achieved through extensions on HTTP.Message. I don't really see the utility of having a "type safe" middleman. Ideally you'd convert the http message directly into the struct or class you want to work w/ without needing to pass through a middle man.

Tests/VaporTests/ControllerTests.swift
```diff
@@ -0,0 +1,44 @@
...
+import XCTest
+
+class ControllerTests: XCTestCase {
+    func testRouting() throws {
+        let app = Application()
+        let sync = try app.make(SyncRouter.self)
+        
+        let controller = SyncController(for: app)
+        
+        controller.on(.get, input: Login.Input.self, to: "user", "joannis") { loginRequest in
```
tanner0101: this is nice, but i don't see the benefit over doing something like:

```swift
app.get(.get, to: "user", "joannis") { req in
    let loginRequest = req.content(LoginRequest.self)
    return Login.Output(...)
}
```

which takes 0 additional code and lets users build on top however they see fit.

Sources/Vapor/Server.swift
Joannis: HTTP.Server makes sense, but Vapor.Server refers to a Server capable of serving HTTP Requests. Vapor itself is a framework that can/will be able to send SMTP messages to an SMTP Server, serve as a WebSocket Server, communicates with a Database Server (Such as MongoDB.Server or MySQL.Server).

Within an individual module like HTTP it makes sense to call it a `Server` without prefix, because you already have the prefix. But this change is the same reason "general" naming such as a module named `Models` can have a severe negative impact. (Which also applies to `Core` and `Engine`, in my opinion)

Joannis:
The failing test on Linux is because the API returns a Substring on this swift 4 snapshot. The Xcode 9 build, however, returns a normal `String`

Sources/Development/main.swift
```diff
@@ -6,10 +6,9 @@ import Routing
 import Service
 import Vapor
 
-extension View: ContentEncodable, ResponseRepresentable {
-    public func encodeContent(to message: Message) throws {
-        message.mediaType = .html
-        message.body = Body(data)
+extension View: ResponseRepresentable {
+    public func makeResponse(for request: Request) throws -> Response {
+        return Response(body: Body(self.data))
```
tanner0101: the media type is no longer being set to HTML here.

Sources/Vapor/Content.swift
```diff
@@ -0,0 +1,69 @@
...
+    /// Encodes to the `MediaType` statelessly
+    static func encodeBody<E: Encodable>(from entity: E) throws -> Body
+}
+
+/// A type capable of decoding an `Encodable` from a `Body`
+public protocol ContentDecoder {
+    static var type: MediaType { get }
+    
+    /// Decodes from the `MediaType` statelessly
+    static func decode<D: Decodable>(_ entity: D.Type, from body: Body) throws -> D
```
tanner0101: I think `decodeBody(_:Body, to: Type)` would match better with `encodeBody`

Sources/Vapor/Error.swift
```diff
@@ -0,0 +1,53 @@
...
+    public var stackTrace: [String]
+    
+    public init(
+        identifier: String,
+        reason: String,
+        file: String = #file,
+        function: String = #function,
+        line: UInt = #line,
+        column: UInt = #column
+        ) {
```
tanner0101: spacing

Sources/Vapor/Server.swift
tanner0101: That's a good point. We should make sure to rename `Vapor.Client` -> `HTTPClient` as well.

Sources/Vapor/SessionCookieMiddleware.swift
```diff
@@ -0,0 +1,35 @@
...
+/// Checks the cookies for each `Request`
+public final class SessionCookieMiddleware: Middleware {
+    /// The cookie to look for
+    let cookieName: String
+
+    /// Fallback handler when the `Cookie` is missing
+    public var missingCookie: ((Request) throws -> (Response))
+    
+    /// Checks the `Cookie.Value` for each `Request`
+    public var cookieValidator: ((Cookie.Value) throws -> ())
```
tanner0101: this seems like it would need access to the request to set the session there. 
we need to make sure we support all the features from: https://github.com/vapor/vapor/blob/master/Sources/Sessions/SessionsMiddleware.swift
there have been quite a few issues raised/PRs to get to that point so I think people will notice if they're gone.

Tests/VaporTests/ControllerTests.swift
```diff
@@ -0,0 +1,46 @@
...
+import Routing
+import XCTest
+
+class ControllerTests: XCTestCase {
+    func testRouting() throws {
+        let app = Application()
+        let sync = try app.make(SyncRouter.self)
+        
+        sync.get("user", "example") { req in
+            let loginRequest = try req.decode(as: Login.Input.self)
```
tanner0101: I think the following is more readable:  
```swift
let loginRequest = try req.content(Login.Input.self)
```
`decode` seems a bit too technical for something that will be used so commonly for some reason. Though I guess it matches nicely with conforming to `Codable`

Tests/VaporTests/ControllerTests.swift
```diff
@@ -0,0 +1,46 @@
...
+            
+            return Login.Output(token: loginRequest.username + loginRequest.password)
+        }
+        
+        let input = Login.Input(username: "example", password: "test")
+        let body = try JSONEncoder.encodeBody(from: input)
```
tanner0101: using encode/decode when something is getting returned seems confusing to me. I think we need to come up with a convention here. Maybe like:

Representable/Initializable
```swift
func makeX() -> X
func init(_: X)
```

Codable
```swift
func encode(to: X)
static func decode(from: X) -> Self
```

Joannis: How would this work with "Coders", rather than a Decodable? Something that can encode and decode entities to a type. You don't want to implement an encode/decode for each type, for each format.

Tests/VaporTests/ControllerTests.swift
Joannis: `req.content(as: Login.Input.self)` would be better IMO. But `decode` is really straight to the point. I suggest keeping it as is and finding out what seems logical after an alpha

Sources/Development/main.swift
Joannis: Good point, I'll also move it out of the `Development` target

Joannis: Small addition: This should be added in a Leaf provider instead. Not inside `Vapor`, so I'll leave that for another PR.

Joannis:
CircleCI fails on the swift build it's running.

Tests/VaporTests/ControllerTests.swift
tanner0101: In that case I think it should be:

```swift
let loginRequest = try req.decode(Login.Input.self)
```

```swift
let loginRequest = LoginRequest(...)
let req = Request()
try req.encode(loginRequest)
```

Tests/VaporTests/ControllerTests.swift
tanner0101: This should be:

```swift
let input = Login.Input(username: "example", password: "test")
let req = Request(path: "/user/example/")
req.encode(input)
```

this is where we would then allow you to do optionally

```swift
try req.encode(input, as: .json)
```

if you want to override the default behavior

Joannis: I really like this, added!

Sources/HTTP/Cookies/RFC1123.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+
+internal struct RFC1123 {
```
tanner0101: some comments here would be nice

Sources/HTTP/Streams/Client.swift
```diff
@@ -2,28 +2,40 @@ import Core
...
     }
-
-    public func inputStream(_ input: SerializedMessage) {
-        tcp.inputStream(input.message)
-        input.onUpgrade?(tcp)
+    
+    /// Sends a single `Request` and returns a future that can be completed with a `Response`
+    ///
+    /// The `Client` *must* not be used during the Future's completion.
+    public func send(request: RequestRepresentable) throws -> Future<Response> {
```
tanner0101: this should probably be `respond(to: Request)`. why is Client no longer a Stream?

Sources/Vapor/Application.swift
```diff
@@ -8,7 +8,7 @@ import Service
 /// Core framework class. You usually create only
 /// one of these per application.
 /// Acts as a service container and much more.
-public final class Application: Container {
+open class Application: Container {
```
tanner0101: i don't think we need subclassing _and_ extendable

Sources/Vapor/CORSConfiguration.swift
```diff
@@ -0,0 +1,98 @@
...
+                guard let origin = request.headers["Origin"] else {
+                    return string
+                }
+                return string.contains(origin) ? origin : string
+            }
+        }
+    }
+    
+    /// Configuration used for populating headers in response for CORS requests.
+    public struct Configuration {
```
tanner0101: I'd like to avoid nested structs if possible. I think this would be better as `CORSConfiguration`

Sources/Vapor/CORSMiddleware.swift
```diff
@@ -0,0 +1,60 @@
+import Web
+import Core
+import HTTP
+
+/// Middleware that adds support for CORS settings in request responses.
```
tanner0101: This should be in `Vapor/CORS/` folder

Sources/Vapor/Client.swift
```diff
@@ -0,0 +1,29 @@
...
+import Dispatch
+import HTTP
+import TCP
+
+public typealias HTTPClient = HTTP.Client
+
+// MARK: Connect
+extension HTTP.Client {
+    /// Connects an `HTTP.Client` to an HTTP Server
+    public static func connect(to uri: URI, on queue: DispatchQueue) throws -> HTTP.Client {
```
tanner0101: I think the static route here is a bad idea. We should create an `HTTPClient` and `ConnectedHTTPClient` class that use instance methods. `HTTPClient` has instance methods for creating a `ConnectedHTTPClient`

Sources/Vapor/Content.swift
```diff
@@ -0,0 +1,69 @@
...
+import Foundation
+import HTTP
+
+/// A type capable of encoding an `Encodable` to a `Body`
+public protocol ContentEncoder {
+    /// The supported `MediaType`
+    static var type: MediaType { get }
+    
+    /// Encodes to the `MediaType` statelessly
+    static func encodeBody<E: Encodable>(from entity: E) throws -> Body
```
tanner0101: I don't think we should use static methods here. This precludes users from creating content encoders that decide their media type at init time.

Sources/Vapor/Content.swift
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/// A type capable of decoding an `Encodable` from a `Body`
+public protocol ContentDecoder {
+    static var type: MediaType { get }
+    
+    /// Decodes from the `MediaType` statelessly
+    static func makeBody<D: Decodable>(_ body: Body, from entity: D.Type) throws -> D
```
tanner0101: this function signature doesn't read very well

Sources/Vapor/Content.swift
```diff
@@ -0,0 +1,69 @@
...
+    }
+    
+    /// Decodes from the `MediaType` statelessly
+    public static func makeBody<D>(_ body: Body, from entity: D.Type) throws -> D where D : Decodable {
+        return try JSONDecoder().decode(D.self, from: body.data)
+    }
+}
+
+/// A class that keeps track of all `ContentEncoder`s and `ContentDecoder`s
+public final class ContentCoders {
```
tanner0101: `ContentCoders` should probably be a protocol. Or the init should allow you to pass in which coders you want. Some way the user needs to be able to override which ContentCoders the application uses.

Sources/Vapor/Client.swift
Joannis: This is the exact same pattern we used in WebSocket, though. And I don't think separating the factory out into `HTTPClient` is helpful. I think it'll add confusion to differentiate between the static factory `HTTPClient` and `ConnectedHTTPClient`

Sources/Vapor/Content.swift
Joannis: Good point. `static var type` also becomes `var type` then?

Sources/Vapor/Content.swift
Joannis: Is `decodeBody` better?

Sources/Vapor/Content.swift
Joannis: You can instantiate a `ContentCoders` already and configure it using the `encoders` and `decoders` variable.

Sources/HTTP/Streams/Client.swift
Joannis: I think `send` is a more sensible API since you're not responding here, you're requesting.

Sources/Vapor/Client.swift
tanner0101: How will this work with the application container and services, then?

Sources/Multipart/Error.swift
```diff
@@ -0,0 +1,69 @@
...
+    public var stackTrace: [String]
+    
+    public init(
+        identifier: String,
+        reason: String,
+        file: String = #file,
+        function: String = #function,
+        line: UInt = #line,
+        column: UInt = #column
+        ) {
```
tanner0101: spacing

Sources/Multipart/Error.swift
```diff
@@ -0,0 +1,69 @@
...
+        self.column = column
+        self.stackTrace = Error.makeStackTrace()
+    }
+    
+    public static func invalidMessage(
+        file: String = #file,
+        function: String = #function,
+        line: UInt = #line,
+        column: UInt = #column
+        ) -> Error {
```
tanner0101: spacing

Sources/Multipart/Multipart.swift
```diff
@@ -0,0 +1,62 @@
...
+    /// Creates a new part
+    init(data: Data, key: String?, headers: Headers) {
+        self.data = data
+        self.key = key
+        self.headers = headers
+    }
+}
+
+/// A Multipart, commonly used in HTTP Forms and SMTP emails
+public struct Multipart {
```
tanner0101: We should avoid having type names w/ same name as module. That prevents people from being able to disambiguate. I think this could be called `Content`. (so `Multipart.Content` when disambiguated).

Sources/Multipart/MultipartParser.swift
```diff
@@ -0,0 +1,207 @@
+import Foundation
+import Bits
+
+/// An enum with no cases can't be instantiated
+///
+/// This parser can only be used statically, a design choice considering the way multipart is best parsed
+public final class MultipartParser {
```
tanner0101: This can be just `Parser`.

Sources/Multipart/TransferEncoding.swift
```diff
@@ -0,0 +1,65 @@
+import Foundation
+import Async
+import Bits
+
+/// Encodes buffers
+public protocol TransferEncoder {
```
tanner0101: What problem does this protocol solve? Should it be moved to a separate module like HTTP?

Sources/Vapor/CORSConfiguration.swift
tanner0101: Still think we should avoid nesting here.

Sources/Multipart/Multipart.swift
Joannis: Yes, but not `Content`. This is the overarching structure. I'd expect the `Content` struct to be the `Multipart`'s `Part`. Maybe `Multipart.Form`?

Sources/Multipart/TransferEncoding.swift
Joannis: SMTP, Multipart etc can have their content encoded according to a set of "formats". Base64, Binary, 7-bits encoding and even custom types such as hexadecimal. HTTP supports this, too. But it's primarily related to Multipart and specifically SMTP.

Sources/Vapor/Content.swift
```diff
@@ -0,0 +1,69 @@
...
+    /// An array of MediaType + ContentEncoders
+    public var encoders: [(MediaType, ContentEncoder.Type)]
+    
+    /// An array of MediaType + ContentDecoders
+    public var decoders: [(MediaType, ContentDecoder.Type)]
```
tanner0101: This could be: `public var decoders: [(MediaType, ContentDecoder)]` (no `.Type`)

Sources/Multipart/Multipart.swift
tanner0101: `Form` seems good to me.

Sources/Development/Models/User.swift
```diff
@@ -1,13 +1,20 @@
...
 import Vapor
 
-final class User: JSONCodable, ContentCodable, ResponseRepresentable {
+final class User: Codable, ResponseRepresentable {
     var name: String
     var age: Int
     var child: User?
     var futureChild: Future<User>?
+    
+    func makeResponse(for request: Request) throws -> Response {
```
tanner0101: This should be automatically implemented if the type also conforms to `Codable`.

I think we could do something like this:

```swift
extension ResponseRepresentable where Self: Codable {
    func makeResponse(for request: Request) throws -> Response {
        let coders = try request
            .requireApplication()
            .make(ContentCoders.self, for: Self.self)

        let (mediaType, coder) = coders.defaultCoder

        let res = Reponse(status: .ok)
        res.body = try Body(coder.encode(self))
        res.mediaType = mediaType

        return res
    }
}
```

Sources/Development/main.swift
```diff
@@ -1,15 +1,16 @@
...
 import Service
 import Vapor
 
-extension View: ContentEncodable, ResponseRepresentable {
-    public func encodeContent(to message: Message) throws {
-        message.mediaType = .html
-        message.body = Body(data)
+extension View: ResponseRepresentable {
+    public func makeResponse(for request: Request) throws -> Response {
+        return Response(headers: [
```
tanner0101: We should set the `mediaType` property on `Response` here for consistency.

Sources/HTTP/Streams/Client.swift
tanner0101: 👍

Sources/HTTP/URI/URI.swift
```diff
@@ -105,3 +105,46 @@ extension String {
         return self == "https" || self == "wss"
     }
 }
+
+// MARK: String literal
+import Foundation
+
+extension URI: ExpressibleByStringLiteral {
```
tanner0101: Should we use the URI parser we get from CHTTP here for consistency?

Sources/Multipart/Multipart.swift
```diff
@@ -0,0 +1,60 @@
+import Foundation
+import Core
+
+/// A single part
+public struct Part {
```
tanner0101: should be in `Part.swift`

Sources/Multipart/Multipart.swift
```diff
@@ -0,0 +1,60 @@
...
+    /// Creates a new part
+    init(data: Data, key: String?, headers: Headers) {
+        self.data = data
+        self.key = key
+        self.headers = headers
+    }
+}
+
+/// A Multipart, commonly used in HTTP Forms and SMTP emails
+public struct Form {
```
tanner0101: should be in `Form.swift`

Sources/Multipart/MultipartParser.swift
```diff
@@ -0,0 +1,207 @@
+import Foundation
```
tanner0101: File should be named `Parser.swift`

Sources/Multipart/MultipartParser.swift
```diff
@@ -0,0 +1,207 @@
...
+import Core
+
+/// An enum with no cases can't be instantiated
+///
+/// This parser can only be used statically, a design choice considering the way multipart is best parsed
+public final class Parser {
+    fileprivate let boundary: Data
```
tanner0101: comments missing.

Sources/Multipart/MultipartParser.swift
```diff
@@ -0,0 +1,207 @@
...
+    fileprivate let fullBoundary: Data
+    fileprivate let data: Data
+    var position = 0
+    var multipart = Form(parts: [])
+    
+    /// Creates a new parser for a Multipart form
+    init(data: Data, boundary: Data) {
```
tanner0101: should this be public?

Sources/Multipart/MultipartParser.swift
```diff
@@ -0,0 +1,207 @@
...
+        // Decodes the part
+        partData = try decoder.decode(partData)
+        
+        let part = Part(data: partData, key: name, headers: headers)
+        
+        multipart.parts.append(part)
+    }
+    
+    /// Parses the `Data` and adds it to the Multipart.
+    fileprivate func parse() throws {
```
tanner0101: would it be possible for this to accept data and return a `Form`? That way you could initialize one parser for a given boundary and re-use it? that would also make this parser design more function which could be easier to maintain.

Sources/Multipart/TransferEncoding.swift
```diff
@@ -0,0 +1,64 @@
...
+    
+    /// Keeps track of all encoding types and their encoder and decoder
+    public static var registery: [String: (encoder: TransferEncoderBuilder, decoder: TransferDecoderBuilder)] = [
+        "binary": (BinaryTransferCoder.init, BinaryTransferCoder.init)
+    ]
+    
+    public static let binary = BinaryTransferCoder()
+}
+
+public final class BinaryTransferCoder: TransferEncoder, TransferDecoder {
```
tanner0101: needs comments

Sources/Multipart/URI.swift
```diff
@@ -0,0 +1,108 @@
+/*
```
tanner0101: We should import `HTTP` or move `URI` to a separate module so we don't dupe code.

Joannis: I forgot to remove this copy of the file.

Sources/Multipart/MultipartParser.swift
Joannis: It's possible, but reusing the boundary doesn't usually happen.

Sources/Multipart/MultipartParser.swift
Joannis: In the futures, maybe. But I intentionally don't expose any APIs that we don't need to. I think this API wouldn't be used often, if at all, and only adds extra API that might need to be broken in the future because we didn't consider an edge case.

Sources/Development/main.swift
Joannis: Normally, yes, but this is the development module, so I don't think it's necessary to make this anything beyond functional for debugging

Sources/Development/Models/User.swift
Joannis: I actually removed that yesterday because I interpreted that you didn't accept it. I'll reintroduce it.

Sources/Core/Futures/PromiseError.swift
```diff
@@ -1,31 +0,0 @@
-import Debugging
-
-/// Errors that can be thrown while using promises.
-public struct PromiseError: Swift.Error, Debuggable, Traceable, Encodable {
```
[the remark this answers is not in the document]
Joannis: There is only one error a promise can emit, so creating a generic error isn't ideal. I don't think the `Async` module should rely on any other Vapor modules. The Async module should (in my opinion) be as objective as possible so that it can be integrated in other libraries easily and without "bloat".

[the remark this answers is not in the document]
tanner0101: That makes sense. I agree.

tanner0101:
please fix merge conflicts and tests. then LGTM. nice work!

PR #1197, opened 2017-10-14
Fixed never executed warning

valeriomazzeo:
screen_shot_2017-10-14_at_13 47 51_360


tanner0101:
i'm not seeing this warning. could you show how to reproduce?

valeriomazzeo:
There’s no particular step to reproduce it. Just import vapor in a new project and compile.

I am not sure why you can’t see it either.

tanner0101:
screen shot 2017-10-25 at 11 53 20 pm
i don't see any warnings for Xcode 9

valeriomazzeo:
@tanner0101 if you build the `Run` scheme the warning is not there, but if you build your `MyProject-Package` scheme then it shows up.

0xTim:
I've seen this as well

valeriomazzeo:
@tanner0101 thanks for Vapor 2.4.2, the warnings are now gone, but this still remains... any chance to get it merged?

tanner0101:
The solution as implemented here would result in an error signal even when the application exits normally (such as after printing routes from `vapor run routes`)

PR #1224, opened 2017-10-27
Add Ubuntu 17.04 (zesty) support

ashokgelal:
According to Vapor APT repository (https://repo.vapor.codes/), it looks like Ubuntu 17.04 is supported but this script file rejects it. If Ubuntu 17.04 is still not officially supported (or there are other reasons), please feel free to reject and close this pull request.

### DIFF
--- a/Utilities/apt.sh
+++ b/Utilities/apt.sh
@@ -1,9 +1,9 @@
 function add_vapor_apt() {
     eval "$(cat /etc/lsb-release)"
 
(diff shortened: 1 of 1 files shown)


0xTim:
I think the main issue is that Swift isn't officially support on 17.04 (https://swift.org/download/#releases) but @joscdk can advise!

calebkleveter:
PR name is misleading. Maybe it should say 'Removed unsupported Ubuntu version 17.04 (zesty)'?

PR #1234, opened 2017-10-31
Added support for pipelining

John-Connolly:
This PR adds support for Redis pipelining.  It concatenates a series of commands to be executed in bulk in order to save round trip time.  

e.g
```swift
let result = try connection.pipeline {
        [
           .array([.bulkString("SET"), .bulkString("hello"), .bulkString("world")]),
           .array([.bulkString("SET"), .bulkString("hello1"), .bulkString("world")])
        ]
 }.blockingAwait(timeout: .seconds(1))
```

### DIFF
--- a/Sources/Redis/Pipeline.swift
+++ b/Sources/Redis/Pipeline.swift
@@ -1,0 +1,52 @@
+import Async
+import Bits
+import Foundation
+
+/// A pipeline which wraps a redis client
+public final class Pipeline<DuplexByteStream: Async.Stream> where DuplexByteStream.Input == ByteBuffer, DuplexByteStream.Output == ByteBuffer, DuplexByteStream: ClosableStream {
+    
+    public let client: RedisClient<DuplexByteStream>
+    private var commands = [RedisData]()
+    
+    
+    init(client: RedisClient<DuplexByteStream>) {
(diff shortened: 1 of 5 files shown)


Sources/Redis/Pipeline.swift
```diff
@@ -0,0 +1,23 @@
+import Async
+
+extension RedisClient {
+    /// Accepts an array of commands, to be sent in one request.
+    public func pipeline(_ commands: () -> ([RedisData])) -> Future<[RedisData]> {
```
Joannis: I'd make this a `Pipeline` object that internally has a `[RedisData]` buffer. That way we can add helpers in the future.

Joannis: (Something similar to what you had before)

John-Connolly:
If it helps I can also document this on the Vapor3 documentation branch

Joannis:
Some documentation would be awesome! I'm going to commit one tiny internal change to `Pipeline` and then I'll merge it. `Pipeline` doesn't need to be generic since it can contain the serialiser rather than the RedisClient. The serialiser is in charge of the process, simplifying the Pipeline API.

Thanks for the PR, great work!

PR #1237, opened 2017-11-01
Async notifications

Joannis:
- Add lightweight notifications to Vapor 3
- Use notifications for errors and closing in streams

This PR also implements capturing the first and capturing the first N results from a stream.

For example; in `findOne` for databases

```swift
let promise = Promise<DatabaseEntity>()

databaseCursor.onClose = {
  promise.complete(nil)
}

databaseCursor.drain { result in
  promise.complete(result)
}

return promise.future
```

And now:

```swift
return databaseCursor.captureFirst()
```


Joannis:
Obsoleted

PR #1285, opened 2017-11-27
Improving WebSocket support.

rafaelGuerreiro:
Add websockets shortcut methods in request, router.
Add support to subprotocols.

The way subprotocols are designed is still WIP. Please, would someone review it?

### DIFF
--- a/Sources/Vapor/Websocket+Router.swift
+++ b/Sources/Vapor/Websocket+Router.swift
@@ -1,0 +1,38 @@
+import Async
+import WebSocket
+import HTTP
+import Routing
+
+extension Router {
+    public typealias WebSocketClosure = (Request, WebSocket) throws -> Void
+
+    @discardableResult
+    public func socket(_ path: PathComponent...,
+        subprotocols: [String]? = nil,
+        use closure: @escaping WebSocketClosure) -> Route {
+        return socket(path, supportedProtocols: { _ in subprotocols ?? [] }, use: closure)
+    }
+
+    @discardableResult
+    public func socket(_ path: PathComponent...,
+        supportedProtocols: @escaping ([String]) -> [String] = { $0 },
+        use closure: @escaping WebSocketClosure) -> Route {
+        return socket(path, supportedProtocols: supportedProtocols, use: closure)
+    }
+
+    private func socket(_ path: [PathComponent],
+                        supportedProtocols: @escaping ([String]) -> [String],
+                        use closure: @escaping WebSocketClosure) -> Route {
+
+        let responder = RouteResponder { request in
+            return try request.upgradeToWebSocket(subprotocols: supportedProtocols) { websocket in
+                try closure(request, websocket)
+            }
+        }
+        let route = Route(method: .get, path: path, responder: responder)
+        self.register(route: route)
+
+        return route
+    }
+}
+
--- a/Sources/Vapor/Websocket+Router.swift
+++ b/Sources/Vapor/Websocket+Router.swift
@@ -1,31 +1,15 @@
-import Async
 import WebSocket
-import HTTP
-import Routing
 
 extension Router {
     public typealias WebSocketClosure = (Request, WebSocket) throws -> Void
 
     @discardableResult
-    public func socket(_ path: PathComponent...,
+    public func websocket(_ path: PathComponent...,
         subprotocols: [String]? = nil,
-        use closure: @escaping WebSocketClosure) -> Route {
-        return socket(path, supportedProtocols: { subprotocols ?? $0 }, use: closure)
-    }
-
-    @discardableResult
-    public func socket(_ path: PathComponent...,
-        supportedProtocols: @escaping ([String]) -> [String] = { $0 },
-        use closure: @escaping WebSocketClosure) -> Route {
-        return socket(path, supportedProtocols: supportedProtocols, use: closure)
-    }
-
-    private func socket(_ path: [PathComponent],
-                        supportedProtocols: @escaping ([String]) -> [String],
-                        use closure: @escaping WebSocketClosure) -> Route {
-
+        use closure: @escaping WebSocketClosure
+    ) -> Route {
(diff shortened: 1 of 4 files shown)


Sources/Vapor/Websocket+Router.swift
```diff
@@ -0,0 +1,22 @@
+import WebSocket
+
+extension Router {
+    public typealias WebSocketClosure = (Request, WebSocket) throws -> Void
+
+    @discardableResult
+    public func websocket(_ path: PathComponent...,
+        subprotocols: [String]? = nil,
```
Joannis: I'd prever, instead of setting the protocols in an array, to provide a `WebSocketSettings` struct with a default value. That way we can add more settings in this `.websocket()` route in the future. I'm already thinking of settings like `maxPacketSize` for configuring memory usage, only accepting `was` (secure) connections etc.

Sources/Vapor/Websocket+Router.swift
```diff
@@ -0,0 +1,22 @@
+import WebSocket
+
+extension Router {
+    public typealias WebSocketClosure = (Request, WebSocket) throws -> Void
+
+    @discardableResult
+    public func websocket(_ path: PathComponent...,
```
Joannis: As part of Vapor 3 we require every PR to have inline documentation for all functions, variables and types. We also require adding a PR for [the vapor docs](https://github.com/vapor/documentation) for API changes. But since we're in an early stage I'll fill this one in for you 🙂

Tests/WebSocketTests/SubProtocolMatcherTests.swift
```diff
@@ -0,0 +1,86 @@
+@testable import WebSocket
+import XCTest
+
+final class SubProtocolMatcherTests : XCTestCase {
```
Joannis: Great job!

Sources/WebSocket/Websocket+Request.swift
```diff
@@ -0,0 +1,63 @@
...
+        self.router = router.filter { !$0.isEmpty }
+    }
+
+    func matching() throws -> String? {
+        if request.isEmpty && router.isEmpty {
+            return nil
+        }
+
+        if request.isEmpty || router.isEmpty {
+            print("Unsupported subprotocol. Request: \(request). Router: \(router)")
```
Joannis: This print statement was left in from debugging I'd assume.

Sources/WebSocket/Websocket+Request.swift
```diff
@@ -0,0 +1,63 @@
+import HTTP
+
+extension Request {
+    public func upgradeToWebSocket(subprotocols: [String]? = nil,
```
Joannis: I'd prefer a `WebSocketSettings` struct here, too.

Sources/WebSocket/Websocket+Request.swift
rafaelGuerreiro: Yeah, I was going to ask you how do I properly log it without polluting the error message.

Joannis: For now, add a TODO comment such as:

`//TODO: Add log warning here"

Sources/Vapor/Websocket+Router.swift
rafaelGuerreiro: Are you thinking in a protocol, like this?
```
protocol WebSocketSettings {
    func accept(subprotocols: [String]) throws -> Strings

    var maxPacketSize: UInt16 { get }

    var onlyAcceptSecureConnection: Bool { get }
}

extension WebSocketSettings {
    func upgradeToWebSocket(request: Request, body: @escaping (WebSocket) throws -> Void) throws -> Response {
         // check if it's secure and onlyAcceptSecureConnection is true
         // check if subprotocols are accepted
         // set websocket maxPacketSize
    }
}
```

Sources/Vapor/Websocket+Router.swift
rafaelGuerreiro: I'll add the inline docs. I forgot that 🤦‍♂️

rafaelGuerreiro:
From WebSocket [RFC 6455, page 22](https://tools.ietf.org/html/rfc6455#page-22)

> /subprotocol/
>           Either a single value representing the subprotocol the server
>           is ready to use or null.  The value chosen MUST be derived
>           from the client's handshake, specifically by selecting one of
>           the values from the |Sec-WebSocket-Protocol| field that the
>           server is willing to use for this connection (if any).  If the
>           client's handshake did not contain such a header field or if
>           the server does not agree to any of the client's requested
>           subprotocols, the only acceptable value is null.  The absence
>           of such a field is equivalent to the null value (meaning that
>           if the server does not wish to agree to one of the suggested
>           subprotocols, it MUST NOT send back a |Sec-WebSocket-Protocol|
>           header field in its response).  The empty string is not the
>           same as the null value for these purposes and is not a legal
>           value for this field.  The ABNF for the value of this header
>           field is (token), where the definitions of constructs and
>           rules are as given in [RFC2616].

Does it mean that I should keep it that way? I'm not sure if we should accept the request and return null in case the subprotocol is not accepted. It will be depending on the client to refuse the response.

Sources/Vapor/Websocket+Router.swift
Joannis: I wouldn't do a protocol.

```swift
public struct WebSocketSettings {
   public var maxPacketSize: Int = Int(UInt16.max)
   public var subProtocols: [String] = []
   public var requireSSL: Bool = false
  
   // empty initialiser with sensible defaults
   public init() {}
}
```

Joannis: the reason not to do a protocol is because protocols really shouldn't have fluctuating requirements. With a struct we can add extra configuration options in the future

Joannis:
RE: Page 22 of RFC6455. I think the current implementation is fine. People with specific needs can still implement their own websocket upgrade mechanism.

PR #1304, opened 2017-12-02
File Middleware improvements and more performance (again)

Joannis:
- Fixes Redis null parsing
- Fixes a bug related to successive HTTP/1 requests being cached
- Improves MySQL performance by 5-20x
- Improves file serving performance by 20-100x

### DIFF
--- a/Sources/Core/File/File.swift
+++ b/Sources/Core/File/File.swift
@@ -22,7 +22,7 @@ public final class File: FileReader, FileCache {
 
     /// See FileReader.read
     public func read<S>(at path: String, into stream: S, chunkSize: Int = 2048)
-        where S: Async.InputStream, S.Input == Data
+        where S: Async.InputStream, S.Input == ByteBuffer
     {
         func onError(_ error: Error) {
             stream.onError(error)
@@ -55,7 +55,7 @@ public final class File: FileReader, FileCache {
                     }
                 } else {
                     if let data = data {
-                        stream.onInput(Data(data))
+                        Data(data).withByteBuffer(stream.onInput)
                     } else {
                         onError(FileError(.readError(error, path: path)))
                     }
--- a/Sources/HTTP/Serializer/BodySerializer.swift
+++ b/Sources/HTTP/Serializer/BodySerializer.swift
@@ -24,6 +24,8 @@ extension HTTPBody {
             }
         case .stream(let bodyStream):
             bodyStream.stream(to: ChunkEncoder()).stream(to: outputStream.outputStream)
+        default:
+            return
         }
     }
 }
--- a/Sources/HTTP/Serializer/BodySerializer.swift
+++ b/Sources/HTTP/Serializer/BodySerializer.swift
@@ -24,8 +24,6 @@ extension HTTPBody {
             }
         case .stream(let bodyStream):
             bodyStream.stream(to: ChunkEncoder()).stream(to: outputStream.outputStream)
-        default:
-            return
         }
     }
 }
--- a/Sources/MySQL/Connection/Connection.swift
+++ b/Sources/MySQL/Connection/Connection.swift
@@ -2,12 +2,13 @@ import Bits
 import Foundation
 import Async
 import TCP
+import TLS
 import Dispatch
 
 /// A connectio to a MySQL database servers
 public final class MySQLConnection {
-    /// The TCP socket it's connected on
-    var socket: TCPSocket
+    /// The socket it's connected on
+    var socket: ClosableStream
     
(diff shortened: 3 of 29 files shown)


Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -15,20 +15,28 @@ public final class FileMiddleware: Middleware {
...
-            throw Abort(.forbidden)
-        }
-
-        let filePath = publicDirectory + path
-        if reader.fileExists(at: filePath) {
-            return try Future(req.streamFile(at: filePath))
-        } else {
-            return try next.respond(to: req)
+        return try next.respond(to: req).map { response in
+            let reader = try req.eventLoop.make(FileReader.self, for: FileMiddleware.self)
```
kdawgwilk: Could you put this inside the if to avoid the overhead when it isn't a 404?

Joannis: Done

Sources/HTTP/Serializer/BodySerializer.swift
```diff
@@ -24,6 +24,8 @@ extension HTTPBody {
             }
         case .stream(let bodyStream):
             bodyStream.stream(to: ChunkEncoder()).stream(to: outputStream.outputStream)
+        default:
```
tanner0101: we should avoid default here so we get a warning when new cases are added.

Sources/MySQL/Parsing/Row+Parsing.swift
```diff
@@ -179,93 +185,95 @@ extension Row {
...
+                    append(.int8(numericCast(int)), forField: field)
                 }
-                
-                append(.int8(int), forField: field)
             }
         case .double:
             // Decodes this string as a Double
-            guard let double = Double(value) else {
-                throw MySQLError(.parsingError)
+            value.baseAddress!.withMemoryRebound(to: Int8.self, capacity: value.count) { pointer in
```
tanner0101: you can also access with `unsafeBaseAddress` (something like that) which is unwrapped.

Sources/Core/File/File.swift
```diff
@@ -55,7 +55,7 @@ public final class File: FileReader, FileCache {
                     }
                 } else {
                     if let data = data {
-                        stream.onInput(Data(data))
+                        Data(data).withByteBuffer(stream.onInput)
```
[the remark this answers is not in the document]
Joannis: Why? Almost all of Vapor 3's internals work with ByteBuffers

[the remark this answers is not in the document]
tanner0101: i imagine file will be used by vapor users as a nicer alternative to Foundation.FileManager.

Sources/MySQL/Connection/Connection.swift
```diff
@@ -72,48 +69,41 @@ public final class MySQLConnection {
...
+            try socket.connect(hostname: hostname, port: port).catch(authenticated.fail)
+            socket.stream(to: parser)
+
+            self.socket = socket
+            self.socketWrite = socket.onInput
+        } else {
+            let socket = try TCPClient(on: eventLoop)
+            
+            try socket.connect(hostname: hostname, port: port).map { _ in
+                socket.start()
```
tanner0101: why don't we need to call `.start()` on the TLS client? are we missing something there?

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -15,20 +20,28 @@ public final class FileMiddleware: Middleware {
...
-            throw Abort(.forbidden)
-        }
-
-        let filePath = publicDirectory + path
-        if reader.fileExists(at: filePath) {
-            return try Future(req.streamFile(at: filePath))
-        } else {
-            return try next.respond(to: req)
+        return try next.respond(to: req).map { response in
+            if response.http.status == 404 {
```
tanner0101: we should continue checking for the file before we hit the router. this will be different than what we did in vapor 2, but it more closely resembles how nginx file serving works and will prevent bugs there. it should also improve the speed of serving files since we don't have to go through the router first.

Sources/MySQL/Connection/Connection.swift
Joannis: It already implicitly starts the connection when it does the TLS handshake

Sources/Core/File/File.swift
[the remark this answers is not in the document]
Joannis: Especially then we should really return `ByteBuffer` so it integrates with the ecosystem

Issue #1305, opened 2017-12-03
Toolbox applications


Joannis:
Vapor Toolbox

Vapor toolbox is currently a CLI tool for macOS and Linux that assists with setting up projects and deploying them to vapor cloud. It has a lot more potential that it can fulfill.

# Toolbox for macOS and iOS

Vapor toolbox for macOS and iOS would strive to be applications that maintain feature pairity to the extend it’s possible and sensible.
The iOS app will not have support for generating Xcode projects, for example, since this is only practical on macOS and possibly on Linux in the future. iOS will also not manage git repositories. It will, however, be able to commmunicate to vapor cloud with most if not all toolbox CLI features.

# New features and Vapor providers 

Besides communication with vapor cloud there are other features that should be added. These features largely require an API to be exposed during development so that toolbox can hook into the vapor application.

## Logging

Toolbox should be able to tail all vapor logs from within the toolbox and be able to filter through specific messages such as setting log levels.

### Notes

- Implementations could add their own logging service. 
- Since logs are generates on multiple workers, this needs to have a singleton construction which could impact performance. Ideally each worker would have their own log destination and add themselves to a global pool as to not impact performance.

## Statistics

- Performance usage per route
- CPU, memory and network usage
- latencies per route and global averages

PR #1307, opened 2017-12-04
Toolbox APIs

Joannis:
This PR works on experimental APIs that assist during development for integration with Vapor toolbox applications ( #1305 )

### DIFF
--- a/Sources/ToolboxProvider/ToolboxLogger.swift
+++ b/Sources/ToolboxProvider/ToolboxLogger.swift
@@ -1,0 +1,52 @@
+import Logging
+import Vapor
+import HTTP
+import Foundation
+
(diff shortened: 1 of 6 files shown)


Sources/ToolboxProvider/ToolboxLogger.swift
```diff
@@ -0,0 +1,52 @@
...
+    private var messages = [LogMessage]()
+    private var lock = NSLock()
+    
+    init(provider: ToolboxProvider) {
+        provider.registerLogger(self)
+    }
+    
+    public func drainMessages() -> [LogMessage] {
+        lock.lock()
+        lock.unlock()
```
linqingmo:
```swift
defer { lock.unlock() } // whether it should be?
```

PR #1333, opened 2017-12-09
Connection pool retain is discardableResult

rafiki270:
allows me to do:
```Swift
worker.pool.retain { connection in
        connection.administrativeQuery("DROP TABLE users")
}
```
without warning on `let _ = worker.pool`

### DIFF
--- a/Sources/MySQL/Connection/ConnectionPool.swift
+++ b/Sources/MySQL/Connection/ConnectionPool.swift
@@ -79,6 +79,7 @@ public final class MySQLConnectionPool {
     /// Retained connections can only be used for a single query at a time
     ///
     ///
+    @discardableResult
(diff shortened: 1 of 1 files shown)


Joannis:
I'm not sure if this is ideal. In most use cases (within routes) this could cause some confusion because people don't notice a warning.

rafiki270:
well, it was just an idea ... :)

Joannis:
Makes sense :) What use case do you have specifically? Maybe we can find another solution

rafiki270:
Well, I have an install phase when the app starts during which I install the db and apply and DB alterations that haven't been executed yet ... so need to use a lot of `connection.administrativeQuery(String)` stuff in a foreach loop and just trying to make my code a bit cleaner ... I know I could just compile a massive string with all the queries in one but guess it would be hard to debug any issues ...

PR #1347, opened 2017-12-12
[WIP] - Add support for autocomplete

joscdk:
Adding support for `--autocomplete`

This command will output one string with commands, arguments and options in one line to support Bash and zsh autocomplete

Todo:

- [X] Implement autocomplete
- [X] Unit tests
- [ ] Test in a real life example
- [ ] Include Bash autocomplete example
- [ ] Include zsh autocomplete example


joscdk:
@tanner0101 we should maybe soon get this merged.

PR #1403, opened 2018-01-03
Make HTTPStatus conform to ResponseEncodable

twof:
Done so statuses can be returned on their own.

Example:
```
        versionOneGroup.delete("park") { (req)  in
            return req.withConnection(to: .sqlite) { (db) in
                return db.query(Parking.self).first().flatMap(to: HTTPStatus.self) {
                    guard let park = $0 else {
                        throw Abort(.notFound, reason: "Could not find parking.")
                    }
                    
                    return park.delete(on: db).transform(to: HTTPStatus.ok)
                }
            }
        }
```

### DIFF
--- a/Sources/Vapor/Engine/ResponseCodable.swift
+++ b/Sources/Vapor/Engine/ResponseCodable.swift
@@ -1,3 +1,4 @@
+import HTTP
 /// Can be converted from a response.
 ///
 /// [Learn More →](https://docs.vapor.codes/3.0/http/response/#responseinitializable)
@@ -30,6 +31,15 @@ extension HTTPResponse: ResponseEncodable {
     public func encode(for req: Request) throws -> Future<Response> {
         let new = req.makeResponse()
         new.http = self
+        return Future(new)
+    }
+}
+
+extension HTTPStatus: ResponseEncodable {
+    /// See ResponseRepresentable.makeResponse
+    public func encode(for req: Request) throws -> Future<Response> {
+        let new = req.makeResponse()
+        new.http = HTTPResponse(status: self)
         return Future(new)
     }
 }


Joannis:
I've got mixed feelings about this. Not for nor against. @tanner0101 ?

twof:
This was done based on an example that Tanner gave in Slack. Should have given that context up front. Sorry about that.

https://qutheory.slack.com/archives/C4QRYJ6DD/p1515010856000374?thread_ts=1514985798.000268&cid=C4QRYJ6DD

```
router.delete("users", User.parameter) { req in
    return req.parameter(User.self).flatMap(to: Void.self) { user in
        return user.delete(on: req)
    }.map(to: HTTPStatus.self) {
        return .success
    }
}


router.get("deleteFirstUser") { req in
    return User.query(on: req).first().flatMap(to: Void.self) { user in
        guard let user = user else {
            throw Abort(.notFound)
        }
        return user.delete(on: req)
    }.map(to: HTTPStatus.self) {
        return .success
    }
}
```

0xTim:
Just a couple of quick notes - I'm able to return an `HTTPResponse` wrapped in a Future already so returning a status isn't much more todo from a framework point of view? Or not much more for a developer to do depending on your opinion.

Also the code above may not be what you want if my understanding of it is correct:

```swift
router.delete("users", User.parameter) { req in
    return req.parameter(User.self).flatMap(to: Void.self) { user in
        return user.delete(on: req)
    }.map(to: HTTPStatus.self) {
        return .success
    }
}
```
This will return a 200 OK once the user has been extracted from the parameter rather than once the user has been deleted - the delete will then happen later (potentially), and if it fails you'll still get a 200 OK. What you want is:

```swift
router.delete("users", User.parameter) { req in
    return req.parameter(User.self).flatMap(to: HTTStatus.self) { user in
        return user.delete(on: req).map(to: HTTPStatus.self) {
            return .success
        }
    }
}
```

@Joannis can you confirm that my thoughts about the code above are correct?

twof:
To add onto this, here's the delete example that will replace the current one if this gets accepted. 

```
        route.delete("user") { (req) in
            return User.query(on: req).first().flatMap(to: HTTPStatus.self) { user in
                guard let user = user else {
                    throw Abort(.notFound)
                }
                
                return user.delete(on: req).transform(to: HTTPStatus.ok)
            }
        }
```

0xTim:
I did not know about that function! That is very handy @twof and cuts down on a load of code

PR #1439, opened 2018-01-19
Add missing headers to request

mcdappdev:
This expands upon #1437 to also fix the second `send` method

### DIFF
--- a/Sources/Vapor/Engine/Client.swift
+++ b/Sources/Vapor/Engine/Client.swift
@@ -36,6 +36,7 @@ extension Client {
             try req.content.encode(content)
             req.http.method = method
             req.http.uri = url
+            req.http.headers = headers
             return try self.respond(to: req)
         }
     }


twof:
Shoot I thought I fixed this here:
https://github.com/vapor/vapor/pull/1437/files

What's the difference between `req.headers` and `req.http.headers`?

mcdappdev:
Not sure, to be honest, I just know there are two `send` functions and one of them didn't attach the headers

twof:
Oh this is a completely different method! Was the other one correctly attaching headers? Did you test this one?

mcdappdev:
Correct, this one is different from the one you fixed. I tested it myself and it works now (the first function does, too), but I haven't written tests. Did you write tests?

twof:
Nah I didn't write any tests. I was just wondering if you tried the method out at all, because I didn't try mine. Just assumed it was a simple enough change that it should work.

mcdappdev:
Cool 👍 Yeah I tried both and they work

PR #1517, opened 2018-02-26
Adds .json() to content

twof:
Adds `.json()` to `Content` so any models that conform to `Content` can easily be turned into JSON

### Checklist


Joannis:
What is the expected result when someone registers a different JSON en/decoder to the `ContentConfig`? Do we (de-)serialize using this other library or do we still use Foundation's?

My expectation would be that JSON would then be decoded using the appropriate `ContentDecoder` for JSON.

Is there a specific reason to return `Data` here? What is the envisioned use case? I personally prefer sticking to `HTTPBody` for most situations since it's more abstract and can be optimized further in the future.

twof:
Works for me. Thanks!

PR #1560, opened 2018-03-22
Check to see if path is a directory as well

### LINKED ISSUE #1559: RC2 - FileMiddleware intercepting root requests
If you enable the FileMiddleware and send a request to `/` then the FileMiddleware will intercept that and just return an empty 200 OK response instead of either a 404 or whatever you have actually registered to that route

0xTim:
Fixes #1559


0xTim:
Also fixes the CI pipeline

PR #1608, opened 2018-04-08
Adds get request with a query to Client

twof:
This PR adds `.get<C: Content>(_:headers:query:)` to `Client` so that users can pass in a `Content` to construct query parameters rather than manually constructing the query themselves.

### Checklist

### DIFF
--- a/Sources/Vapor/HTTP/Client.swift
+++ b/Sources/Vapor/HTTP/Client.swift
@@ -74,6 +74,19 @@ extension Client {
 /// MARK: Content
 
 extension Client {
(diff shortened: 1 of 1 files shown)


twof:
From Discord

"""
```swift
self.httpClient.post("https://www.googleapis.com/oauth2/v4/token") { post in
    try post.content.encode(content, as: .urlEncodedForm)
}
```
that's probably what you would want
that was the problem with the old method... you didn't have the same power as ContentContainer
that's why it will be moving to a closure
"""

So that means I can close this PR right?

PR #1684, opened 2018-05-16
HTTPRequest caching policy support

rafiki270:
**Requires:**
- [ ] Add caching to FoundationClient: https://github.com/vapor/vapor/pull/1674
- [ ] HTTPRequest caching support: https://github.com/vapor/http/pull/292


rafiki270:
After conversation with Tanner, I will try to add this to the FoundationClients send method instead

MrLotU:
@rafiki270 would it be allright to delete the branch? :)

rafiki270:
Go for it please

PR #1690, opened 2018-05-20
eliminate warnings for Swift 4.1

rpinz:
eliminate warnings for Swift 4.1

### Checklist

### DIFF
--- a/Sources/Vapor/HTTP/Accept.swift
+++ b/Sources/Vapor/HTTP/Accept.swift
@@ -33,6 +33,27 @@ extension Request {
             return []
         }
 
+        #if swift(>=4.1)
+        let accepts: [Accept] = acceptString.toCharacterSequence().split(separator: ",").compactMap { acceptSlice in
+            let pieces = acceptSlice.split(separator: ";")
+            guard let mediaType = pieces.first.flatMap({ String($0) }) else { return nil }
+
+            let preference: Double
+            if pieces.count == 2 {
+                let q = pieces[1].split(separator: "=")
+                if q.count == 2 {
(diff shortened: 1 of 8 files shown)


grundoon:
@rpinz Looks like these are for vapor2 branch, not master?

rpinz:
Correct.  These eliminate warnings while compiling Vapor2 with Swift 4.1.

grundoon:
Right, but look above: "rpinz wants to merge 1 commit into vapor:master from rpinz:warnings" – this merge is set to `master`

Sources/Vapor/HTTP/Accept.swift
```diff
@@ -33,6 +33,27 @@ extension Request {
             return []
         }
 
+        #if swift(>=4.1)
+        let accepts: [Accept] = acceptString.toCharacterSequence().split(separator: ",").compactMap { acceptSlice in
```
LoganWright: I think we can reduce code duplication between this and 4.0 a bit somehow.

Sources/Vapor/HTTP/AcceptLanguage.swift
```diff
@@ -11,6 +11,27 @@ extension Request {
             return []
         }
         
+        #if swift(>=4.1)
+        return acceptLanguageString.toCharacterSequence().split(separator: ",").compactMap { acceptLanguageSlice in
```
LoganWright: Same thing here, maybe we should break out some of the compactMap contents into a separate function or closure that we toss into the compactMap. What do you think?

LoganWright:
Absorbed by other PR.

PR #1737, opened 2018-06-23
Added fluent dependency and a makeCreatedLocationResponse function to…

grosch:
… Request

Add a new method to Request for makeCreatedLocationResponse.  It sets the HTTP status code to 201 and then returns the newly created object.  It also sets the `Location` header appropriately.  Over time we can add things like ETAGS or whatever seems appropriate.

### Checklist

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -42,6 +42,9 @@ let package = Package(
 
         // 🔌 Non-blocking, event-driven WebSocket client and server built on Swift NIO.
         .package(url: "https://github.com/vapor/websocket.git", from: "1.0.0"),
+
+        // Swift ORM framework (queries, models, and relations) for building NoSQL and SQL database integrations.
+        .package(url: "https://github.com/vapor/fluent.git", from: "3.0.0-rc"),
     ],
     targets: [
         // Boilerplate
@@ -66,7 +69,8 @@ let package = Package(
             "TemplateKit",
             "URLEncodedForm",
             "Validation",
-            "WebSocket"
+            "WebSocket",
+	    "FluentSQL"
         ]),
         .testTarget(name: "VaporTests", dependencies: ["Vapor"]),
     ]
--- a/Sources/Vapor/Request/Request.swift
+++ b/Sources/Vapor/Request/Request.swift
@@ -1,3 +1,6 @@
+import FluentSQL
+
+
 /// `Request` is a service-container wrapper around an `HTTPRequest`.
 ///
 /// Use `Request` to access information about the `HTTPRequest` (`req.http`).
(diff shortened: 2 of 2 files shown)


0xTim:
@grosch Vapor and Fluent deliberately don't know about each other so having Fluent as a dependency on the main repo likely won't fly. However this could be a good addition in a separate package, lower down the stack or in community.

vzsg:
Alternatively, you could change the signature of the function to something like the following to eliminate the dependency on Fluent:

```swift
func makeCreatedLocationResponse<T, ID>(obj: T, id: ID, as mediaType: MediaType = .json, customLocation: String? = nil) throws -> Response 
```

A KeyPath might also work.

Package.swift
```diff
@@ -42,6 +42,9 @@ let package = Package(
 
         // 🔌 Non-blocking, event-driven WebSocket client and server built on Swift NIO.
         .package(url: "https://github.com/vapor/websocket.git", from: "1.0.0"),
+
+        // Swift ORM framework (queries, models, and relations) for building NoSQL and SQL database integrations.
+        .package(url: "https://github.com/vapor/fluent.git", from: "3.0.0-rc"),
```
LoganWright: Unfortunately, we aren't comfortable including fluent as a dependency at this moment in order to keep the ecosystem as flexible as possible for use with other ORMs, etc.

Sources/Vapor/Request/Request.swift
```diff
@@ -153,6 +156,25 @@ public final class Request: ContainerAlias, DatabaseConnectable, HTTPMessageCont
...
 
+    /// Generate a creation `Response` for a `Model` that includes the new object as well as a 201 status code.
+    ///
+    /// - Parameters:
+    ///   - obj: The newly created object.
+    ///   - mediaType: The type of data to return the container with.
+    ///   - customLocation: A customer path location.
+    /// - Returns: A new 201 `Response` on the same container as the current `Request` populated with the new `Model`
+    /// - Throws: `FluentError` if the `Model` doesn't have an object ID.
+    func makeCreatedLocationResponse<T>(obj: T, as mediaType: MediaType = .json, customLocation: String? = nil) throws -> Response where T: Model {
```
LoganWright: Perhaps something more flexible here, I don't see offhand a reason that fluent needs to be included. @vzsg had a good idea here about using something more flexible like Codable.

Sources/Vapor/Request/Request.swift
```diff
@@ -153,6 +156,25 @@ public final class Request: ContainerAlias, DatabaseConnectable, HTTPMessageCont
...
+        let id = try obj.requireID()
```
LoganWright: Is there a reason that this needs to include an Id

Sources/Vapor/Request/Request.swift
```diff
@@ -153,6 +156,25 @@ public final class Request: ContainerAlias, DatabaseConnectable, HTTPMessageCont
...
+
+        let location = customLocation ?? "\(http.url.absoluteString)/\(id)"
```
LoganWright: I see it here, it's possible that the id of the fluent object wouldn't be the actual language code, or might be a situation where they need it at a different URL.

PR #1763, opened 2018-07-20
Conform HTTPStatus to Content

calebkleveter:
Conforms `HTTPStatus` (which is a typealias for NIO's `HTTPResponseStatus` enum) to the `Codable` and `Content` protocols. The biggest benefit of this is that you can now pass in `HTTPStatus` to a method that requires a type conforming to `Content`. For example, if you have [a method](https://github.com/skelpo/PayPal/blob/develop/Sources/PayPal/API/PayPalClient.swift#L55) that decodes a `Response` object a to `Content` type, `HTTPStatus` is now valid. 

### Checklist

### DIFF
--- a/Sources/Vapor/Response/HTTPStatus.swift
+++ b/Sources/Vapor/Response/HTTPStatus.swift
@@ -2,6 +2,7 @@
 public typealias HTTPStatus = HTTPResponseStatus
 
 extension HTTPStatus: Codable {
+    
     /// See [`Decodable.init(from:)`](https://developer.apple.com/documentation/swift/decodable/2894081-init).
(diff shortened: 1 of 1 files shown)


LoganWright:
Can you give some examples of where this might be useful? Right now, it seems odd to me to have status be content since content implies things that would be stored in a database or more accurately received and send as the body of an HTTPMessage, which status shouldn't really be.

Could the methods in question be changed to accept a more flexible protocol?

LoganWright:
We had some discussions on this internally and think we're going to opt to push this out of the core library. If there's functions that require additional flexibility, we'll look into making them accept more items 👍 

You're the best 😄

Issue #43, opened 2016-02-21
Benchmark require


wildfiremedia:
Wish some day you can comparing PHP 7.x/HHVM 3.5 on Laravel 5.x and Lumen benchmark.

mpclarkson:
I've benchmarked Vapor at 5000+ requests per second returning a json response using apache bench on my laptop. We're using PHP7 on one of our work dev servers and it is certainly much, much faster than 5. Not only is it faster but the memory usage is also significant lower. That being said, as PHP is interpreted, IMHO it's not going to be as fast as Swift, which is compiled.

wildfiremedia:
Thanks for your benchmark and I assumed you posted in HN as well. What you think of Go can be a potential as a Laravel framework clone if there is?

tannernelson:
I would love to have a benchmark of Vapor vs other web frameworks that we could put in the Wiki.

elithrar:
Two option:

* Add Vapor to the Techempower benchmarks: https://www.techempower.com/benchmarks/#section=code&hw=peak&test=fortune
* Benchmark using a tool that is aware of [Coordinated Omission](https://groups.google.com/forum/m/#!msg/mechanical-sympathy/icNZJejUHfE/BfDekfBEs_sJ) (aka not `ab`). Doing your own benchmarks is still hard though: any non-trivial (writing static text isn't a benchmark) example takes a bit of work to write, and implementing that in other languages to compare can be a rabbit hole. Setting some goals (must turn a 8K data structure into JSON, must serialize and HMAC a cookie, etc) would be good. 

Plus hey, 5000/req/s is 18 *million* requests an hour. Even at their peak hour, most web services won't see that kind of traffic!

wildfiremedia:
While I haven't try Vapor yet, if you could compile with -fast optimization
that will covert code to LLVM?

tannernelson:
@elithrar The web frameworks benchmark looks like a great resource for development. I will add passing those benchmarks to the roadmap.

tannernelson:
Added https://github.com/qutheory/vapor/wiki/Roadmap

PR #1766, opened 2018-07-21
[Don't merge] Use private container to get database connection

t-ae:
Multiple requests can grab same database connection.
See [DatabaseKit issue #43](https://github.com/vapor/database-kit/issues/43) for details.

`Request` tries to get database connection from [self (as `Container`)](https://github.com/vapor/vapor/blob/72273607b8b1fa5241b9d77a1e419eed7d5bc73d/Sources/Vapor/Request/Request.swift#L173). Internally the connection is [cached](https://github.com/vapor/database-kit/blob/9ea3fbea959b99560e44a4dfc8ee1f132a23e5a2/Sources/DatabaseKit/Connection/Container%2BCachedConnection.swift#L20) in the container.

But `Request` is `AliasedContainer`, so cache is shared among the requests.
If the server receives two (or more) requests simultaneously, they use the same database connection and it can crash [here](https://github.com/vapor/mysql/blob/8717e4a6389ff1ff56801f4e6411309105945c62/Sources/MySQL/Connection/MySQLConnection.swift#L81) for MySQL, or somewhere else for other DB. 

To make cache not shared, I changed to use `privateContainer` instead of `Request` itself.
There is [similar situation](https://github.com/vapor/vapor/blob/72273607b8b1fa5241b9d77a1e419eed7d5bc73d/Sources/Vapor/Sessions/Request%2BSession.swift#L13) so I think it may work well.

### Checklist

### DIFF
--- a/Sources/Vapor/Request/Request.swift
+++ b/Sources/Vapor/Request/Request.swift
@@ -170,7 +170,7 @@ public final class Request: ContainerAlias, DatabaseConnectable, HTTPMessageCont
             return eventLoop.newFailedFuture(error: error)
         }
         hasActiveConnections = true
-        return requestCachedConnection(to: database)
+        return privateContainer.requestCachedConnection(to: database)
     }
 
     // MARK: Request Codable
@@ -188,7 +188,7 @@ public final class Request: ContainerAlias, DatabaseConnectable, HTTPMessageCont
     /// Called when the `Request` deinitializes.
     deinit {
         if hasActiveConnections {
-            try! releaseCachedConnections()
+            try! privateContainer.releaseCachedConnections()
         }
     }
 }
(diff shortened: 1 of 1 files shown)


t-ae:
Well, it turned out my implementation has a problem.

To make cache not shared, I changed to use `privateContainer` instead of `Request` itself to call `requestCachedConnection`.
But internally, `requestCachedConnection` tries to [get `DatabaseConnectionPool`](https://github.com/vapor/database-kit/blob/master/Sources/DatabaseKit/Connection/Container%2BCachedConnection.swift#L31). The connection pools are also separated if we use `privateContainer`.

t-ae:
Created another PR for DatabaseKit.
https://github.com/vapor/database-kit/pull/44

After merged that, I'll continue this PR.

penny-coin:
Hey @t-ae, you just merged a pull request, have a coin! 

You now have 3 coins.

pankajsoni19:
Package.json still uses db kit 1.0.0. Can we update the supporting dependencies for vapor?

PR #1821, opened 2018-09-23
Add support to ErrorMiddleware for user defined CustomizedError with …

JovanMilenkovic:
This PR intends to add an easy way to throw your error in whatever format you wish with default ErrorMiddleware implementation backing you up by encoding the exact format in response body.

### DIFF
--- a/Sources/Vapor/Error/CustomizedError.swift
+++ b/Sources/Vapor/Error/CustomizedError.swift
@@ -1,0 +1,56 @@
+/// Errors conforming to this protocol will always be displayed by
+/// Vapor to the end-user (even in production mode where most errors are silenced).
+///
+///     extension MyError: CustomizedError { ... }
+///     throw MyError(...)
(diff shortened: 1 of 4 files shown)


calebkleveter:
How is this protocol different from `AbortError`?

JovanMilenkovic:
I'm not seeing much similarities between them except the status code and headers. AbortError results in a response with a body:
`{
      "error": true,
      "reason": "{custom message}"
}`
This is supposed to allow any custom error format. There was some previous discussion [here](https://github.com/vapor/vapor/issues/1804#issuecomment-423462273) that could maybe make it clearer.

MrLotU:
@JovanMilenkovic With your permission I’d like to close this PR in favor of #1875 because IMO it’s a cleaner implementation without having to add more protocols.

JovanMilenkovic:
Yes sure

PR #1851, opened 2018-12-11
Accept both %20 and + for url encoding space characters

baarde:
`+` is the correct way to encode a space character according to the reference (see vapor/url-encoded-form#19). However, we cannot change this test without updating URLEncodedForm first. But URLEncodedForm depends on Vapor for its tests, which means we need to change this test for URLEncodedForm's tests to pass. So this is an intermediate state.


penny-coin:
Hey @baarde, you just merged a pull request, have a coin! 

You now have 6 coins.

PR #1866, opened 2019-01-10
fixed webSocketConnect function not forwarding request headers

wshamp:
The webSocketConnect func NIOWebSocketClient was not forwarding the request headers to HTTPClient.websocket func.  This func has a default parameter for the headers if not passed, so any headers i.e. authentication headers, were being stripped preventing the verification in the upgrade handler.

I have just added the passed in request headers to the HTTPClient.websocket call.

I have added a test that verifies the upgrade handler gets the headers now.


penny-coin:
Hey @wshamp, you just merged a pull request, have a coin! 

You now have 1 coins.

PR #1876, opened 2019-01-19
style_guide.md

twof:
As Vapor grows, and more developers begin contributing, the project risks becoming a difficult to navigate mish-mash of coding styles. style_guide.md is created to enforce a level of quality and consistency across every repository within the org. The first style rules introduced will primarily be a formalization of existing code style.

Things that need to be decided:
Use of `self`. Enforced or no?

### Checklist


mcdappdev:
I like this idea, but I think there needs to be some distinction between this and what’s currently available on the docs site. I think the latter is more about overall architectural syles and this is more about syntax type stuff, am I right?

twof:
@mcdappdev Yeah exactly. This is a style guide for Vapor itself, and the current docs have the style guide for projects using Vapor.

Docs/style_guide.md
```diff
@@ -0,0 +1,117 @@
...
+public func encode(using container: Container) throws -> Future<Request> {
+    let req = Request(using: container)
+    try req.content.encode(self)
+    return Future.map(on: container) { req }
+}
+```
+
+### Extensions
+
+See Function Definitions
```
MrLotU: Maybe include that in extensions/protocol imps, we’d like a
```swift
//see `ProtocolVariable`
```
to not have to write doc comments twice

Docs/style_guide.md
```diff
@@ -0,0 +1,117 @@
...
+If closure definitons are small enough to fit on one line, then the starting and ending
+curly brace should be on the same line.
+
+```swift
+return Future.map(on: container) { req }
+```
+
+## Line Length and Line Breaks
+
+The maximum length for a line of code is XXXXXXX characters. Anything larger than that
```
MrLotU: Either 100 or 80 IMO

Docs/style_guide.md
```diff
@@ -0,0 +1,117 @@
...
+let foo = [String]()
+
+// Not preferred
+let foo: [String] = []
+```
+
+## Misc
+
+### Colons
+No matter whether it's a type anotation or in the middle of a dictionary, colons are always followed by a space.
```
MrLotU: And no space before the colon. So never do
```swift
let x : String = nil
```

Docs/style_guide.md
twof: Picking 100 for now. I've used 80 for Ruby before, but Swift tends to be more verbose, so I think it deserves an extra 20 characters

Docs/style_guide.md
```diff
@@ -0,0 +1,152 @@
...
+/// Use the `require...` methods to fetch coders by `MediaType`.
+public struct ContentCoders: ServiceType {
+```
+
+### Protocol Implementations
+Rather than writing doc comments twice, leave a refrence to the protocol being conformed to.
+
+```swift
+extension Parameter {
+    /// See `Parameter`.
```
MrLotU: Should we reference like this:
```swift
/// See `ProtocolOrClassName`
```
Or
```swift
/// See `ProtocolOrClassName.functionOrMemberName`
```

I think the 2nd is not needed, but there are things to be said for both of them.

twof:
@MrLotU good point. I've updated the wording of the poll

Docs/style_guide.md
twof: All of the snippets in the doc are existing code within the Vapor org. I'm going to say I favor the first simply because that's the prevailing style.

Mordil: IIRC jazzy will link to the most specific information is has available.

So
```swift
/// See `ProtocolOrClassName.functionOrMemberName(_:label:)`
```
will a generate a link to that method that will also have a fragment (`#`) to jump the user to that portion of the page.

Mordil:
For what it's worth on the enforcement of explicit `self.` - the style I _try_ to keep to, is if I'm accessing something that is **not** defined on the type (such as a super class) to use `self`.

Docs/style_guide.md
```diff
@@ -0,0 +1,152 @@
...
+and one tab in from the start of the signiture. The ending parenthases and curly brace goes
+on its own line at the same indentation as the start of the signiture.
+
+```swift
+internal init(
+    httpEncoders: [MediaType: HTTPMessageEncoder],
+    httpDecoders: [MediaType: HTTPMessageDecoder],
+    dataEncoders: [MediaType: DataEncoder],
+    dataDecoders: [MediaType: DataDecoder]
+) {
```
Mordil: What is the opinion on functions that have generic `where` clauses and returns values?
```swift
func foo<T>(
    firstLongLabel: Int,
    secondLongLabel: Bool,
    thirdLongLabel: String
) -> EventLoopFuture<T> where T: SomeProtocol {
    // body
}
```

Mordil: I can see several twists on that - what if the `where` clause is so long that the line is starting to hit the 100 character break rule?

Do we put the where clause on its own line, wrap at the last `,` portion, or?

```swift
// first style
) -> EventLoopFuture<T> where T: SomeProtocol, SomeOtherRestriction, SomeFurtherRestriction,
    SomeFinalRestriction {
    // body
}

// second
) -> EventLoopFuture<T>
    where T: SomeProtocol, SomeOtherRestriction, SomeFurtherRestriction, SomeFinalRestriction
{
    // body
}
```

Docs/style_guide.md
```diff
@@ -0,0 +1,152 @@
...
+```
+
+### Function Calls
+
+```swift
+throw FluentError(
+    identifier: "noDefaultDatabase",
+    reason: "A default database is required if no database ID is passed to `\(Self.self).query(_:on:)` or if `\(Self.self)` is being looked up statically.",
+    suggestedFixes: ["Set `\(Self.self).defaultDatabase` or to fix this error."]
+)
```
Mordil: What's do we think should be done for chaining with method calls?

```swift
let result = property.method(
    someInt: myInt,
    someString: "I'm showing a really long method call",
    someBool: true
).map(String.init)
    .map { $0 == "Foo" }
```

Docs/style_guide.md
```diff
@@ -0,0 +1,152 @@
...
+## Initializers
+
+### Arrays
+
+```swift
+// Preferred
+let foo = [String]()
+
+// Not preferred
+let foo: [String] = []
```
Mordil: This seems backwards to what I recall hearing Tanner say.

@tanner0101 Am I recalling incorrectly?

Docs/style_guide.md
```diff
@@ -0,0 +1,152 @@
...
+```swift
+/// Configured `HTTPMessageEncoder`s.
+private let httpEncoders: [MediaType: HTTPMessageEncoder]
+
+/// Returns an `HTTPMessageEncoder` for the specified `MediaType` or throws an error.
+///
+///     let coder = try coders.requireHTTPEncoder(for: .json)
+///
+/// - parameters:
+///     - mediaType: An encoder for this `MediaType` will be returned.
```
Mordil: This is a super minor point: But should the parameter name in the doc block be the caller's label, or the callee's label?

Like, should this be `for`?

twof: I'm going to say it should the the param name rather than the label. Labels typically don't offer specific information, and labels can also be used more than one in a function signature. I believe what we have now is in line with the style that XCode generates.

Docs/style_guide.md
twof: I would opt to place chained function calls on the same indent depth, one tab in from the start of the chain. So in your case I'd go with 

```swift
let result = property.method(
        someInt: myInt,
        someString: "I'm showing a really long method call",
        someBool: true
    ).map(String.init)
    .map { $0 == "Foo" }
```
Totally up for discussion, though this is in line with the current style AFAIK

Docs/style_guide.md
twof: I'd go with the style you suggested for the where + return situation. For long `where` clauses, this is an example of what I've done historically.

https://github.com/twof/VaporCRUDRouter/blob/master/Sources/CrudRouter/Obsoleted/CrudControllerObsoleted.swift#L38

```swift
    public func crudRegister<ChildType, ThroughType>(
        at path: PathComponentsRepresentable...,
        forSiblings relation: KeyPath<ModelType, Siblings<ModelType, ChildType, ThroughType>>,
        relationConfiguration: ((CrudSiblingsController<ChildType, ModelType, ThroughType>) throws -> Void)?=nil
    ) throws where
        ChildType: Content,
        ModelType.Database == ThroughType.Database,
        ChildType.ID: Parameter,
        ThroughType: ModifiablePivot,
        ThroughType.Database: JoinSupporting,
        ThroughType.Database == ChildType.Database,
        ThroughType.Left == ModelType,
        ThroughType.Right == ChildType {
            fatalError()
    }
```

Open to other suggestions though. This style is nice because it provides a visual distinction between the different part of function definition and it works in the most cases.

Mordil: What I've been doing lately (using your example):

```swift
public func crudRegister<ChildType, ThroughType>(
        at path: PathComponentsRepresentable...,
        forSiblings relation: KeyPath<ModelType, Siblings<ModelType, ChildType, ThroughType>>,
        relationConfiguration: ((CrudSiblingsController<ChildType, ModelType, ThroughType>) throws -> Void)?=nil
) throws // -> EventLoopFuture<Void>
    where ChildType: Content,
    ModelType.Database == ThroughType.Database,
    ChildType.ID: Parameter,
    ThroughType: ModifiablePivot,
    ThroughType.Database: JoinSupporting,
    ThroughType.Database == ChildType.Database,
    ThroughType.Left == ModelType,
    ThroughType.Right == ChildType
{
    fatalError()
}```

Comment (twof) [reply Docs/style_guide.md]: I actually think that reads better

File: Docs/style_guide.md (hunk 7 above)
Comment (MrLotU) [reply Docs/style_guide.md]: Neither are preferred, preferred is to ALWAYS initialize in an init. So you'd end up with this:
```swift
struct MyStruct {
    let initialEmpty: [String]

    init() {
        self.initialEmpty = []
    }
}

Comment (MrLotU) [reply Docs/style_guide.md]: Unless I too mixed up what tanner said :P

File: Docs/style_guide.md (hunk 5 above)
Comment (MrLotU) [reply Docs/style_guide.md]: I also prefer @Mordil's snippet 😄

File: Docs/style_guide.md (hunk 7 above)
Comment (stjernegard) [reply Docs/style_guide.md]: I think Vapor is currently using `name: [Type] = []`.
[Here](https://github.com/vapor/vapor/blob/6d743a3df125e1e977438a2032cc67bae443968b/Sources/Vapor/Error/Abort.swift#L14) is one example. There are a bunch more if you search for them. I can't find any cases where Vapor is using the `name = [Type]()` style.

That style also matches the style for creating default values for a function argument.
```
func foo(bar: [String] = [])
```

Comment (MrLotU) [reply Docs/style_guide.md]: But that's _inside_ an init, not top level

Comment (stjernegard) [reply Docs/style_guide.md]: Well, the part of the documentation in discussion here doesn't mention top-level properties specifically. Just a preference for initialising array variables.

Preferring to initialise all properties in the init probably deserves its own section of the style guide.

Comment (grundoon) [reply Docs/style_guide.md]: I have no skin in this game, but have lately been working on refining a style that is _intended_ to be easiest on those who will be reading the code in the future. And what I've found seems to work best is adherence to the "types and variables gravitate left" rule...

Akin to generally placing `@attribute` on its own line and access controls flush left, if the most important bit _can_ be placed left of its applicable operator, it probably should be (otherwise, as close as possible) – all solely for the sake of visual parsing. (and fwiw, thinking it through this way _did_ make me refactor my preferred init style)

```swift
var doubleArray: [Double] = [] // best
var doubleArray = [Double]() // okay, but consider an entire list of `=` jumping
var doubleArray = [] as [Double] // pretty sketchy

let newValue = currentValue + 42 // gravitate left 
let newValue = 42 + currentValue

if ourValue > 42 { ... } // gravitate left
if 42 < ourValue { ... }
```

Narsail:
What about using precommit hooks (with https://github.com/shibapm/Komondor) with https://github.com/nicklockwood/SwiftFormat to get rid of the discussion altogether? Every coder can more or less write in his own preferred style but it will be pushed in a standardized way.

PR #1824, opened 2018-09-29
Add Debuggable properties to Abort

mikkelu:
Fixes #1804 

Added the following properties to `Abort` so all inherited properties from `Debuggable` are accessible:
- possibleCauses
- suggestedFixes
- documentationLinks
- stackOverflowQuestions
- gitHubIssues

Also made `stackTrace` optional so it conforms to `Debuggable` as discussed in #1804 

### Checklist

- [x] Circle CI is passing (code compiles and passes tests).
- [x] There are no breaking changes to public API.
- [x] New test cases have been added where appropriate.
- [x] All new code has been commented with doc blocks `///`.

PR #1884, opened 2019-02-09
Feature request: Add `AbortError` initializer to `Abort`

mikkelu:
This PR creates a new initializer in `Abort` that takes an `AbortError` as a parameter. 

The motivation is that a lot of errors are created with Swift enums, but by using them, it is hard to capture source location information without hacky stuff. This PR will make it easy to create errors that can be thrown with Abort and capture source location: 
`throw Abort(MyErrorEnum.SomeError)`

The cool thing is it allows one to share error code between, say an iOS app and a Vapor backend. Let's say I have an error enum shared across multiple Swift projects: 
```swift
enum AuthError: String, AbortError {
  case notAuthorized
}

extension AuthError {

  var identifier: String {
    return rawValue
  }

  var reason: String {
    switch self {
    case .notAuthorized: return "You are not authorized"
    }
  }

  var status: HTTPResponseStatus {
    switch self {
    case .notAuthorized: return .unauthorized
    }
  }

}
```
Then I can check the error response in my iOS code and check the identifier for proper handling - and it will capture source location that I can use in my server side logging.

I think we should wait with this PR until #1824 gets merged so that we can add the remaining `Debuggable` properties to the new initializer.

### Checklist


penny-coin:
Hey @mikkelu, you just merged a pull request, have a coin! 

You now have 2 coins.

PR #1921, opened 2019-02-28
add Secure Middleware

satishbabariya:
Secure middleware provides protection against cross-site scripting (XSS) attack, content type sniffing, clickjacking, insecure connection, and other code injection attacks.

**Requires PR [https://github.com/vapor/http/pull/338](https://github.com/vapor/http/pull/338)**

### Checklist


0xTim:
My view is that given these headers are only used by browsers, this PR should live in Leaf (or be a [separate package](https://github.com/brokenhandsio/VaporSecurityHeaders))

joscdk:
I agree with @0xTim i would also much rather have it in the Security Headers package, and maybe have some official documentation on using it.

MrLotU:
Agreed, @joscdk is the Security Headers package already existent?

joscdk:
@MrLotU thanks to @0xTim it does :) https://github.com/brokenhandsio/VaporSecurityHeaders

MrLotU:
Awesome!

satishbabariya:
I think it should be in vapor middleware including other middlewares like,

1. Static File Middleware
2. CORS Middleware
3. Secure Middleware
4. CSRF Middleware
5. GZIP Middleware
6. Request ID Middleware
7. Logger Middleware
8. Proxy Middleware
9. Body Limit Middleware

Which also maintain by vapor, and can be accessed directly by the single package (From my point of view...),


**For example, if you want to serve Angular/React/Vue build's (Dist's) using Static Middleware.  Also API form Vapor**
For this kind of App if you have All three **File + Secure + CORS** Middleware in the same package and maintained by same repo that would be plus.

Sources/Vapor/Middleware/SecureMiddleware.swift
```diff
@@ -0,0 +1,130 @@
...
+                xssProtection: "1; mode=block",
+                xContentTypeOptions: "nosniff",
+                xframeOptions: "SAMEORIGIN",
+                strictTransportSecurity: "max-age=0",
+                contentSecurityPolicy: "default-src 'self'"
+            )
+        }
+
+        /// X-XSS-Protection provides protection against cross-site scripting attack (XSS)
+        public let xssProtection: String?
```
tanner0101: Would it be possible to make this and related properties not stringly typed? A struct that can serialize to a string would be really nice to use from Swift.

satishbabariya: can be made into something like this 
```swift 
enum XSSFiltering {
    enum BrowsersAction {
        case block
        case report(url: String)
    }
    case disable
    case enable(BrowsersAction?)
}
```

tanner0101: That would be great

satishbabariya: also if we are at it we can make an enum of X-Frame-Options,  Strict-Transport-Security, and X-Content-Type-Options.
only Content-Security-Policy which will stays string and will be on the user to set.

tanner0101: Yeah that would be awesome. Anything that can be represented reasonably as an enum or struct would be great. Regarding `Content-Security-Policy`, it seems like that has a fixed structure of `<policy-directive>; <policy-directive>`, so maybe we could at least have:
```swift
struct ContentSecurityPolicy {
    var directives: [String]
}
```
Where the directives are joined by a semicolon. If you wanted, you could even do:
```swift
struct ContentSecurityPolicy {
    struct Directive: ExpressibleByStringLiteral {
        let string: String
    }
    var directives: [Directive]

    init(_ directives: Directive...) { self.directives = directives }
}
```
That way you could still load up directives with a raw string, but it opens the door to adding some convenience extension for common directives.

```swift
extension ContentSecurityPolicy.Directive {
    static func defaultSrc(_ value: String) -> Directive {
        return .init(string: "default-src \(value)")
    }
}

let config = SecurityMiddleware.Configuration(
    contentSecurityPolicy: .init(.defaultSrc("https:"), "report-uri /foo/"))
)
```

satishbabariya: updated SecureMiddleware, tested with vapor 3.2.2. not tested with current master

0xTim:
I'm unsure about including this by default in Vapor since it's only relevant if you're using a browser for the most part.

Other concerns I have are by setting CSP to self only by default. I think CSP needs to be explicitly opt in, otherwise it's going to confuse users as to why their sites suddenly break.

Finally, enabling HSTS with a max age of 0 defeats the point of using it. And enabling it by default with any other value is a complete no go - you'll destroy localhost for the duration set (the recommended is a year)

Also there are no tests for this, the middleware should be extensively tested for all use cases.

I'm not trying to be negative btw! I think it's a great contribution and suggestion, but the implications need to be seriously discussed

0xTim:
@tanner0101 the latest Vapor Security Headers package contains a Swift API for CSP - https://github.com/brokenhandsio/VaporSecurityHeaders/pull/14

satishbabariya:
you can do something like this
```swift
/// Content-Security-Policy: default-src 'self'; img-src *; media-src media1.com media2.com; script-src userscripts.example.com

SecureMiddleware.ContentSecurityPolicy(arrayLiteral: .defaultSrc("'self'"), .imgSrc("*"), .mediaSrc("media1.com media2.com"), .scriptSrc("userscripts.example.com"))
```

Sources/Vapor/Middleware/SecureMiddleware.swift
```diff
@@ -0,0 +1,300 @@
+/// Middleware that provides protection against cross-site scripting (XSS) attack, content type sniffing, clickjacking, insecure connection and other code injection attacks.
+/// For configuration of this middleware please use the `SecureMiddleware.Configuration` object.
+///
+/// - note: Make sure this middleware is inserted before all your error/abort middlewares,
+///         so that even the failed request responses contain proper Security Headers.
+public final class SecureMiddleware: Middleware {
+    /// XSS Protection. with browser action.
+    ///
+    /// - block: Enables XSS filtering. Rather than sanitizing the page, the browser will prevent rendering of the page if an attack is detected.
```
tanner0101: Case comments should go on each `case ` declaration. That makes them easy to view in our generated API docs.

screen shot 2019-03-01 at 11 00 02 am

Sources/Vapor/Middleware/SecureMiddleware.swift
```diff
@@ -0,0 +1,300 @@
...
+    /// - report: Enables XSS filtering. If a cross-site scripting attack is detected, the browser will sanitize the page and report the violation. This uses the functionality of the CSP report-uri directive to send a report.
+    public enum BrowserAction: CustomStringConvertible {
+        case block
+        case report(url: String)
+        
+        public var description: String {
```
tanner0101: All public members need docblocks. For this you can just do something like:
```swift
/// `CustomStringConvertible` conformance.
```
But the more descriptive, the better.

Sources/Vapor/Middleware/SecureMiddleware.swift
```diff
@@ -0,0 +1,300 @@
...
+                return "1; report=\(url)"
+            }
+        }
+    }
+    
+    /// Cross Site Scripting (XSS) Configrations
+    ///
+    /// - disable: Disables XSS filtering.
+    /// - enable: Enables XSS filtering with configration
+    ///     - ".enable(nil)" Enables XSS filtering (usually default in browsers). If a cross-site scripting attack is detected, the browser will sanitize the page (remove the unsafe parts).
```
tanner0101: Quoted things should be back-ticked instead. Our generated API docs will code format that + sometimes create links.

tanner0101: doc block lines should also not be more than ~100 characters

Sources/Vapor/Middleware/SecureMiddleware.swift
```diff
@@ -0,0 +1,300 @@
...
+        }
+    }
+    
+    /// Content-Security-Policy header providing security against cross-site scripting (XSS), clickjacking and other code
+    /// injection attacks resulting from execution of malicious content in the trusted web page context.
+    ///
+    /// For more information, see [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)
+    public struct ContentSecurityPolicy: ExpressibleByArrayLiteral {
+        /// The policy is a string containing the policy directives describing your Content Security Policy.
+        public struct Policy: ExpressibleByStringLiteral {
```
tanner0101: nice 👍

Sources/Vapor/Middleware/SecureMiddleware.swift
```diff
@@ -0,0 +1,300 @@
...
+            
+            public static func scriptSrc(_ value: String) -> Policy {
+                return .init(stringLiteral: "script-src \(value)")
+            }
+            
+            public static func imgSrc(_ value: String) -> Policy {
+                return .init(stringLiteral: "img-src \(value)")
+            }
+            
+            public static func mediaSrc(_ value: String) -> Policy {
```
tanner0101: These all need doc blocks

Sources/Vapor/Middleware/SecureMiddleware.swift
```diff
@@ -0,0 +1,300 @@
...
+        /// man-in-the-middle (MITM) attacks.
+        ///   - contentSecurityPolicy: header providing security against cross-site scripting (XSS), clickjacking and other code
+        /// injection attacks resulting from execution of malicious content in the
+        public init(
+            xssProtection: XSSProtection? = nil,
+            xContentTypeOptions: XContentTypeOptions? = nil,
+            xframeOptions: XFrameOptions? = nil,
+            strictTransportSecurity: StrictTransportSecurity? = nil,
+            contentSecurityPolicy: ContentSecurityPolicy? = nil
+            ) {
```
tanner0101: spacing

satishbabariya:
all the comments are updated.

gperdomor:
@tanner0101 so the middlewares in the list would be included by default? If the answer is Yes I will send a PR to include `RequestIdMiddleware`

Sources/Vapor/Middleware/SecureMiddleware.swift
```diff
@@ -0,0 +1,332 @@
+/// Middleware that provides protection against cross-site scripting (XSS) attack, content type sniffing, clickjacking, insecure connection and other code injection attacks.
+/// For configuration of this middleware please use the `SecureMiddleware.Configuration` object.
+///
+/// - note: Make sure this middleware is inserted before all your error/abort middlewares,
+///         so that even the failed request responses contain proper Security Headers.
+public final class SecureMiddleware: Middleware {
```
tanner0101: Maybe we should call this `SecurityHeadersMiddleware`?

tanner0101:
@gperdomor not sure what you mean by that, can you clarify? 
@satishbabariya: Tim is right, we should definitely have some tests for this.

Sources/Vapor/Middleware/SecureMiddleware.swift
satishbabariya: no prob.

satishbabariya:
@tanner0101 sure I'll add test cases.

satishbabariya:
@tanner0101, ready to review, added some test cases also, only XCTestManifests not generated.

Joannis:
Bump: Can we still merge this for Vapor 3 users? And should this be replicated for Vapor 4?

tanner0101:
@Joannis this is targeting Vapor 4 already.

Tests/VaporTests/XCTestManifests.swift
```diff
@@ -31,6 +31,7 @@ extension ApplicationTests {
         ("testRequestQueryStringPercentEncoding", testRequestQueryStringPercentEncoding),
         ("testResponseEncodableStatus", testResponseEncodableStatus),
         ("testRootGet", testRootGet),
+        ("testSecurityHeadersMiddleware", testSecurityHeadersMiddleware),
```
tanner0101: This change should be undone as `XCTestManifests` is now gone on master.

tanner0101:
Removing `XCTestManifests.swift` should fix the merge conflict and allow me to merge this.

tanner0101:
Closing this since CI is still failing. Feel free to re-open if you get CI passing and we can merge 👍

PR #2028, opened 2019-07-28
Use FoundationNetworking on Linux for Swift 5.1

joscdk:
In Swift 5.1 on Linux, the network stuff is moved to `FoundationNetworking`. Since 5.1 is still pre-release, i think we should keep this as a PR for now, and when we get close to a 5.1 release, we can push this.

Since a lot of development on Linux happens through Sourcekit-LSP, which only supports Swift 5.1, this will also give a good example for Linux developers.

### Checklist


joscdk:
@tonyarnold i have poked @tanner0101 to see if we can get it merged :)

PR #2044, opened 2019-09-02
Add WebSocket onPong support

vkill:
Hi, there

We should use ping-pong to check WebSocket connection is active, because  we cannot receive any error when hosting provider block the network.

The sample code:
```
var closeScheduledTask: Scheduled<()>? = nil

ws.onPong { ws, _ in
  closeScheduledTask?.cancel()
}
eventLoop.scheduleRepeatedTask(initialDelay: TimeAmount.seconds(3), delay: TimeAmount.seconds(30)) { (task) -> Void in
  ws.send(raw: [UInt8](Data("ping".utf8)), opcode: .ping, promise: nil)
  closeScheduledTask = eventLoop.scheduleTask(in: TimeAmount.seconds(10)) {
     _ = ws.close()
  }
}
```

Please review it, thanks.


vkill:
Because #2074
Ref https://github.com/vapor/websocket-kit/pull/45

PR #2045, opened 2019-09-03
public WebSocketClient

vkill:
Hi, there
Maybe should support direct init WebSocketClient, because I want to change default maxFrameSize.

Please review it, thanks.


vkill:
Because #2074

PR #2057, opened 2019-09-17
fix for percent encoding breaking array parsing

abonham:
I found that when POSTing using url-form-encoding from a form checkbox with a name of "productId[]", the parser was not decoding before checking if the suffix was "]". This meant that, it would not decode arrays from input.

### Checklist


thecheatah:
@abonham I am the original author for this code. You should be percent decoding within `private func parseKey(string: Substring) throws -> URLEncodedFormEncodedKey {` You shouldn't be percent decoding everything. The decoder relies on the values being percent encoded. I would recommend that you add a test case to explicitly define what you are trying to fix.

From my understanding, the code currently decodes `arr[]` but doesn't decode `arr%5B%5D` which is the bug you are trying to fix.

PR #2059, opened 2019-09-17
Make session cookie `HttpOnly` by default

t089:
It is [good practice](https://www.owasp.org/index.php/HttpOnly) to mark session cookies as `HttpOnly`. This PR set `httpOnly: true` for the default cookie factory for session cookies.

### Checklist


t089:
The test failures seem to be unrelated to my changes ☹️

PR #2101, opened 2019-11-21
Feature/validation touch ups

siemensikkema:
This adds a view (final?) minor changes to the new Validation API. Hopefully this can be part of the next beta.

- change Validation.init API to better match the discussion in #2071 
- remove ValidationBuilder functionality (for now). I think it's not ready yet.
- add missing API docs.


Sources/Vapor/Validation/Validation.swift
```diff
@@ -1,63 +1,97 @@
...
 
-    public init<T: Decodable>(key: BasicCodingKey, as: T.Type = T.self, required: Bool = true, validator: Validator<T>) {
+    /// Creates a new `Validation` based on a `Validator`.
+    ///
+    /// - Parameters:
+    ///   - key: The key to the value to be validated.
+    ///   - as: The type of the value to be validated.
+    ///   - required: Fail on absense of the value.
+    ///   - validator: The validator to use for validation.
+    public init<T: Decodable>(_ key: String, as: T.Type = T.self, required: Bool = true, is validator: Validator<T>) {
```
MrLotU: If I read this change correctly, this removes the API allowing users to pass in a `BaseCodingKey` directly. Is that intended?

siemensikkema: @MrLotU Correct but you have to see it in the context of #2071 where it was first added, by me. Initially I had versions for `String`, `CodingKey` and `BasicCodingKey` but Tanner said `String` would be sufficient. This PR just ties up some loose ends from the discussion that did not make it in before the PR got merged.

Having said that, if you think that should be changed, I'm all ears 🙂

MrLotU: Nope. It’s fine by me, was just wondering! Rationale sounds good to me

PR #2148, opened 2020-01-15
Allow nested route building to throw and rethrow

baarde:
This PR changes the signature of the `RoutesBuilder.group(_:configure:)` methods to rethrow if the `configure` function throws.

### Rationale

The `RoutesBuilder.register(collection:)` method throws. However, the `configure` function used in the `RoutesBuilder.group(_:configure:)` methods does not throw.

Therefore the following example doesn't currently work:
```swift
let middleware = MyMiddleware()
let controller = MyController()
let routes = Routes()
// error: Invalid conversion from throwing function of type '(_) throws -> ()' to non-throwing function type '(RoutesBuilder) -> ()'
routes.group(middleware) { router in
    try router.register(collection: controller)
}
```

We can use `try!` or `try?` but I think it would be better to just rethrow the error:
```swift
try routes.group(middleware) { router in
    try router.register(collection: controller)
}
```

### Checklist

### Breaking changes

The change *may* break existing code.

This example **does not break**:
```swift
routes.group(middleware) { router in
    nonThrowingFunction(router)
}
```

But this one **does break**:
```swift
let function: ([Middleware], (RoutesBuilder) -> ()) -> ()
// error: Invalid conversion from throwing function of type '([Middleware], (RoutesBuilder) throws -> ()) throws -> ()' to non-throwing function type '([Middleware], (RoutesBuilder) -> ()) -> ()'
function = routes.group(_:configure:)
```
However it is very unlikely to occur in the wild.

### DIFF
--- a/Sources/Vapor/Routing/RoutesBuilder+Group.swift
+++ b/Sources/Vapor/Routing/RoutesBuilder+Group.swift
@@ -28,8 +28,8 @@ extension RoutesBuilder {
     /// - parameters:
     ///     - path: Group path components separated by commas.
     ///     - configure: Closure to configure the newly created `Router`.
-    public func group(_ path: PathComponent..., configure: (RoutesBuilder) -> ()) {
-        configure(HTTPRoutesGroup(root: self, path: path))
+    public func group(_ path: PathComponent..., configure: (RoutesBuilder) throws -> ()) rethrows {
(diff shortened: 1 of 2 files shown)


MrLotU:
To avoid breaking public API, maybe add the throwing & rethrowing option alongside the non-throwing ones?

PR #2242, opened 2020-03-14
Adds ETag related methods

grosch:
REST API's that return a modifiable object should include an ETag header for data sanity on subsequent PUT/PATCH calls.  This allows clients to easily include that header in their `get` routes, if desired.

```swift
func get(req: Request) throws -> EventLoopFuture<Response> {
    return Todo.find(req.parameters.get("todoID"), on: req.db)
        .unwrap(or: Abort(.notFound))
        .flatMapThrowing { try SomeDTO($0).eTagResponse() }
}
```


grosch:
Making too many changes for this to be ready to look at.

PR #2292, opened 2020-04-06
Add didRegister callback to LifecycleHandler

0xTim:
This PR adds a `didRegister` callback to `LifecycleHandler`. This is useful if you lifecycle type needs to initialise things before configuration, for example, initialise `Storage` with various keys for services it provides.

This avoids either having to have a separate setup function in the lifecycle handler, which is necessary to stop crashes. These occur if a user of a lifecycle tries to configure something before `boot` is called. For instance a repository type necessary for that lifecycle handler.


0xTim:
No leave it for now, I managed to work around it

Issue #2226, opened 2020-03-06
Unix domain sockets


TyrfingMjolnir:
Unix domain sockets are not yet implemented in Vapor.

I would like to use unix domain sockets in order to decreasing the network prerequisits of my project.

### Steps to reproduce

```
enum Bind {
    case hostname(_: String, port: Int)
    case unixDomainSocketPath(_:String)
}
```

### Expected behavior
/opt/mymicroservice --unixdomainsocket /tmp/mymicroservice

My micro service written in Vapor will become available through unix domain socket opposed to through TCP/IP stack.
### Actual behavior

It's a waste of resources to load the network stack for TCP/IP when unix domain socket is a valid option.

### Environment

Darwin devMB15R.local 19.3.0 Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST 2020; root:xnu-6153.81.5~1/RELEASE_X86_64 x86_64

-bash: vapor: command not found

* Vapor Framework version: 4.0.rc3.1
* Vapor Toolbox version: 5.2
* OS version: MacOS 10.15

Prior art: https://github.com/TyrfingMjolnir/swift-talk-backend/blob/master/Sources/NIOWrapper/NIOWrapper.swift#L266

dimitribouniol:
I have a working implementation of this in my local branch: https://github.com/dimitribouniol/vapor/tree/unix-domain-sockets

@tanner0101 Should I create a PR for this?

MrLotU:
Hiya @dimitribouniol, yes, please create a PR! Any comments/requests/improvements can be discussed there than 😄

Joannis:
This is available since Vapor [4.29.0](https://github.com/vapor/vapor/releases/tag/4.29.0)

PR #2371, opened 2020-05-28
Support for Unix Domain Sockets

dimitribouniol:
These changes add support for Unix Domain Sockets (#2226) to Vapor.

A server can be started using the `serve` command:
```
$ swift run Run serve --unix-socket /tmp/vapor.socket
```

Alternatively, the app can be configured in code to connect to a socket:
```swift
app.http.server.configuration.unixDomainSocketPath = "/tmp/vapor.socket"
```

In either scenario, if a socket path is specified, it takes precedence over binding the server to a hostname/port. Note that TLS may be combined with this options if you wish the server to only listen to encrypted connections made to the socket file.

The socket file must not exist prior to starting the server. If one is left behind during an incomplete shutdown, however, and the app has permission to delete it, it will be overidden.

To assist with making connections to socket paths, URI has also been extended with a static `withSocketPath()` constructor:
```swift
try app.client.get(.withSocketPath("/tmp/vapor.socket", path: "/foo"))
```
Note that making requests to other socket-path-based services requires AsyncHTTPClient 1.1.0 or newer.

### DIFF
--- a/Sources/Vapor/Commands/ServeCommand.swift
+++ b/Sources/Vapor/Commands/ServeCommand.swift
@@ -13,6 +13,9 @@ public final class ServeCommand: Command {
         
         @Option(name: "bind", short: "b", help: "Convenience for setting hostname and port together.")
         var bind: String?
+        
+        @Option(name: "unix-socket", short: nil, help: "Set the path for the unix domain socket file the server will bind to.")
+        var socketPath: String?
 
         public init() { }
     }
@@ -44,8 +47,13 @@ public final class ServeCommand: Command {
         let port = signature.port
             // 0.0.0.0:8080, :8080, parse port
             ?? signature.bind?.split(separator: ":").last.flatMap(String.init).flatMap(Int.init)
-
-        try context.application.server.start(hostname: hostname, port: port)
+        let socketPath = signature.socketPath // /tmp/vapor.socket
+        
+        if let socketPath = socketPath, !socketPath.isEmpty {
+            try context.application.server.start(socketPath: socketPath)
+        } else {
+            try context.application.server.start(hostname: hostname, port: port)
+        }
         self.server = context.application.server
 
         // allow the server to be stopped or waited for
--- a/Sources/Vapor/Error/UnixDomainSocketPathError.swift
+++ b/Sources/Vapor/Error/UnixDomainSocketPathError.swift
@@ -1,0 +1,41 @@
+import NIO
+
+/// Unix domain socket errors may be thrown when attepting to start the server from a unix domain socket path.
+public enum UnixDomainSocketPathError: AbortError {
+    case inaccessible(IOError)
+    case unsupportedFile(mode_t, String)
+    case couldNotRemove(IOError)
+    case socketInUse(IOError)
+    case noSuchDirectory(IOError)
+    
+    /// See `AbortError.status`
+    public var status: HTTPResponseStatus {
+        return .internalServerError
+    }
+
+    /// See `AbortError.identifier`
+    public var identifier: String {
+        switch self {
+        case .inaccessible: return "inaccessible"
+        case .unsupportedFile: return "unsupportedFile"
+        case .couldNotRemove: return "couldNotRemove"
+        case .socketInUse: return "socketInUse"
+        case .noSuchDirectory: return "noSuchDirectory"
+        }
+    }
+    
+    /// See `CustomStringConvertible`.
+    public var description: String {
+        return "Unix Domain Socket Path error: \(self.reason)"
+    }
+
+    /// See `AbortError.reason`
+    public var reason: String {
+        switch self {
+        case .inaccessible(let ioError), .couldNotRemove(let ioError), .socketInUse(let ioError), .noSuchDirectory(let ioError):
+            return ioError.description
+        case .unsupportedFile(_, let message):
+            return message
+        }
+    }
+}
--- a/Sources/Vapor/HTTP/Client/EventLoopHTTPClient.swift
+++ b/Sources/Vapor/HTTP/Client/EventLoopHTTPClient.swift
@@ -16,7 +16,7 @@ private struct EventLoopHTTPClient: Client {
     ) -> EventLoopFuture<ClientResponse> {
         do {
             let request = try HTTPClient.Request(
-                url: URL(string: client.url.string)!,
+                url: client.url.socketURL ?? URL(string: client.url.string)!,
                 method: client.method,
                 headers: client.headers,
                 body: client.body.map { .byteBuffer($0) }
--- a/Sources/Vapor/HTTP/Server/HTTPServer.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServer.swift
@@ -184,7 +184,13 @@ public final class HTTPServer: Server {
 
         // print starting message
         let scheme = configuration.tlsConfiguration == nil ? "http" : "https"
-        let address = "\(scheme)://\(configuration.hostname):\(configuration.port)"
+        let address: String
+        if let socketPath = configuration.unixDomainSocketPath {
+            address = "\(scheme)+unix: \(socketPath)"
+        } else {
+            address = "\(scheme)://\(configuration.hostname):\(configuration.port)"
+        }
+        
         self.configuration.logger.notice("Server starting on \(address)")
 
         // start the actual HTTPServer
@@ -295,6 +301,15 @@ private final class HTTPServerConnection {
             .childChannelOption(ChannelOptions.socket(IPPROTO_TCP, TCP_NODELAY), value: configuration.tcpNoDelay ? SocketOptionValue(1) : SocketOptionValue(0))
             .childChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: configuration.reuseAddress ? SocketOptionValue(1) : SocketOptionValue(0))
             .childChannelOption(ChannelOptions.maxMessagesPerRead, value: 1)
+        
+        if let socketPath = configuration.unixDomainSocketPath {
+            return bootstrap.bind(unixDomainSocketPath: socketPath).map { channel in
+                return .init(channel: channel, quiesce: quiesce, configuration: configuration)
+            }.flatMapErrorThrowing { error -> HTTPServerConnection in
+                quiesce.initiateShutdown(promise: nil)
+                throw error
+            }
+        }
         
         return bootstrap.bind(host: configuration.hostname, port: configuration.port).map { channel in
             return .init(channel: channel, quiesce: quiesce)
--- a/Sources/Vapor/HTTP/Server/HTTPServer.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServer.swift
@@ -321,30 +321,27 @@ private final class HTTPServerConnection {
         
         // Check if a socket path has been configured, and use it if it has been.
         if let socketPath = configuration.unixDomainSocketPath {
-            return prepareSocketFile(
+            return prepareSocketFileForBinding(
                 for: configuration,
                 with: application.eventLoopGroup.next(),
                 in: application.threadPool
-            ).flatMap { _ in
-                return bootstrap.bind(unixDomainSocketPath: socketPath).map { channel in
-                    return .init(
-                        channel: channel,
-                        quiesce: quiesce,
-                        configuration: configuration
-                    )
-                }
+            ).flatMap {
+                return bootstrap.bind(unixDomainSocketPath: socketPath)
+            }.map { channel in
+                return .init(
+                    channel: channel,
+                    quiesce: quiesce,
+                    configuration: configuration
+                )
             }.flatMapErrorThrowing { error -> HTTPServerConnection in
                 quiesce.initiateShutdown(promise: nil)
                 
-                // Transform `bind(descriptor:ptr:bytes:): Address already in use (errno: 48)` to a form the user can test for
-                if let ioError = error as? IOError {
-                    switch ioError.errnoCode {
-                    case EADDRINUSE: throw UnixDomainSocketPathError.socketInUse(ioError)
-                    case ENOENT: throw UnixDomainSocketPathError.noSuchDirectory(ioError)
-                    default: break
-                    }
+                // Transform `bind(descriptor:ptr:bytes:): Address already in use (errno: 48)`/`No such file or directory (errno: 2)` into a form the user can test for
+                switch error as? IOError {
+                case .some(let io) where io.errnoCode == EADDRINUSE: throw UnixDomainSocketPathError.socketInUse(io)
+                case .some(let io) where io.errnoCode == ENOENT: throw UnixDomainSocketPathError.noSuchDirectory(io)
+                default: throw error
                 }
-                throw error
             }
         }
         
@@ -373,26 +370,12 @@ private final class HTTPServerConnection {
             promise.fail(Abort(.internalServerError, reason: "Server stop took too long."))
         }
         self.quiesce.initiateShutdown(promise: promise)
-        return promise.futureResult.flatMapAlways { [configuration, channel] result in
-            let threadPool = NIOThreadPool(numberOfThreads: 1)
-            threadPool.start()
-            
-            return prepareSocketFile(
-                for: configuration,
-                with: channel.eventLoop,
-                in: threadPool
-            ).flatMapAlways { result in
-                let promise = channel.eventLoop.makePromise(of: Void.self)
-                threadPool.shutdownGracefully { error in
-                    if let error = error {
-                        promise.fail(error)
-                    } else {
-                        promise.succeed(())
-                    }
-                }
-                return promise.futureResult.flatMapThrowing { try result }
-            }.flatMapThrowing { try result }
+        do { // remove the socket file if it exists. This is ok to block since we are shutting down, and are not in an event loop
+            try removeSocketFileIfPresent(for: configuration)
+        } catch {
+            promise.fail(error)
         }
+        return promise.futureResult
     }
     
     var onClose: EventLoopFuture<Void> {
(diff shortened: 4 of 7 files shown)


dimitribouniol:
Thank you @MrLotU for your feedback! Honestly, I would love to have `removeSocketFile()` return a future, but I am utterly unfamiliar with how to do so (I'm not even sure I handled it correctly in `close()` 😅 - please double check that).

Would you be able to spare some time to show me some examples of how I should go about doing so, not only for creating a promise (which I _kinda_ get, but I assume I would need to make some new error types?), but also for chaining it in bind/close. I'm happy to hop onto Discord to discuss implementations.

Until then, the errors that NIO gives back are pretty sufficient - they only needed a little guidance via some logs to help users who wouldn't expect them. I can add some tests to make sure cleanup occurs (and that a non-socket file doesn't;t already exist), though I'm not too sure how I can simulate the other scenarios via unit test (invalid permissions, etc...).

dimitribouniol:
@MrLotU So I think I figured out this Future stuff, and have refactored `removeSocketFile()` into `prepareSocketFile() -> EventLoopFuture<Void>`, and have added tests for most of the situations a user could run into. Some that I'm not to sure of include:
- Testing against a path the user has no access to (`.inaccessible()`) — although I can contrive a path that only root has access to on macOS, I have no idea what such a path would look like on linux.
- Testing against a path the user can read, but cannot unlink (`couldNotRemove()`) — again, not sure how we can set up such a scenario in a unit test.
- Testing against a path NIO can create a socket file against, but that we cannot unlink when the server is finished — again, not sure how we can set up such a scenario in a unit test.

Something that I have not been able to actually figure out, is that if `HTTPServerConnection.close()` fails, tests won't actually detect this and fail, although an error will be logged to the console. This is the case even without dealing with socket paths, so I think I handled everything correctly, but do let me know if I've gone astray.

Also, note that `prepareSocketFile()` uses blocking API (`stat()` and `unlink()`), though will only be called before the server starts/after it stops. Please let me know if I should figure out a way around this, or if it should be fine in this scenario.

Finally, I don't know if I should be logging here as `.notice` or `.critical`. Specifically, these issues would prevent the server from even starting, so I think that warrants something like critical (especially with information to help the user resolve the issue), but I'll defer to all of you to let me know what I should use 😅.

dimitribouniol:
I added some additional error handling (I now check to see if the bootstrap failed for some socket related reasons), and migrated `prepareSocketFile()` to be non-blocking. Please let me know if I should change anything else!

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -322,6 +404,60 @@ private final class HTTPServerConnection {
...
+            // If stat() failed because a file doesn't exist, then don't do anything and report success — the socket file won't need to be removed.
+            guard statError == ENOENT else {
+                // If stat reports another error, we likely can't remove the file either, which means NIO will fail when binding, so tell the user by logging it so they can take action.
+                configuration.logger.critical("Could not access an existing socket file located at \(socketPath): POSIX Error \(statError). Please remove this file in order to start the server.")
+                throw UnixDomainSocketPathError.inaccessible(IOError(errnoCode: statError, reason: "Could not access an existing socket file located at \(socketPath)"))
+            }
+            
+            // stat() reported that the file doesn't exist, indicating NIO can go ahead and create it when it binds, so don't do anything and report success.
+            return
+        }
```
gwynne, line 436: A few issues:

1. This will not necessarily read the correct value for `errno`. In C this would be a sequence point violation; in Swift the semantics are well-defined, but not necessarily obvious.
1. Reading `errno` if the function call _didn't_ fail is not guaranteed to be well-defined, though hardly a catastrophe of course.
1. `.withCString` is superfluous; Swift will perform the bridging transparently.
1. Nested `guard` is a bit strange to read.

My recommendation:
```suggestion
        let statResult = lstat(socketPath, &socketFileStat)

        // Since we only want to continue if there is a socket file, stop here if an error occurred with stat().
        guard statResult == 0 else {
            let statError = errno
            
            if statError == ENOENT {
                // If stat() failed because a file doesn't exist, then don't do anything and report success — the socket file won't need to be removed.
                return
            }

            // If stat reports another error, we likely can't remove the file either, which means NIO will fail when binding, so tell the user by logging it so they can take action.
            configuration.logger.critical("Could not access an existing socket file located at \(socketPath): POSIX Error \(String(cString: strerror(statError))). Please remove this file in order to start the server.")
            throw UnixDomainSocketPathError.inaccessible(IOError(errnoCode: statError, reason: "Could not access an existing socket file located at \(socketPath)"))
        }
```

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -322,6 +404,60 @@ private final class HTTPServerConnection {
...
+        
+        let mode = socketFileStat.st_mode
+        
+        // We only want to remove a socket file if it is actually a socket file - the user may have mistyped something, and it would be unfortunate for them to lose anything important otherwise.
+        guard mode&S_IFSOCK == S_IFSOCK else {
```
gwynne, line 441: Style nit:
```suggestion
        guard mode & S_IFSOCK == S_IFSOCK else {
```

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -322,6 +404,60 @@ private final class HTTPServerConnection {
...
+        
+        // The file exists, and is a socket file, so try to unlink it.
+        let (unlinkResults, unlinkError) = socketPath.withCString { path in
+            (unlink(path), errno)
+        }
+
+        guard unlinkResults == 0 else {
+            // If an error occurs during unlink, NIO will fail when binding, so report the error.
+            throw UnixDomainSocketPathError.couldNotRemove(IOError(errnoCode: unlinkError, reason: "Could not remove an existing socket file located at \(socketPath)"))
+        }
```
gwynne, line 455: Same `errno` concerns as earlier. Suggestion:
```suggestion
        guard unlink(socketPath) == 0 else {
            // If an error occurs during unlink, NIO will fail when binding, so report the error.
            throw UnixDomainSocketPathError.couldNotRemove(IOError(errnoCode: errno, reason: "Could not remove an existing socket file located at \(socketPath)"))
        }
```

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -291,17 +319,52 @@ private final class HTTPServerConnection {
...
+                
+                // Transform `bind(descriptor:ptr:bytes:): Address already in use (errno: 48)` to a form the user can test for
+                if let ioError = error as? IOError {
+                    switch ioError.errnoCode {
+                    case EADDRINUSE: throw UnixDomainSocketPathError.socketInUse(ioError)
+                    case ENOENT: throw UnixDomainSocketPathError.noSuchDirectory(ioError)
+                    default: break
+                    }
+                }
+                throw error
```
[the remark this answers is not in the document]
dimitribouniol: Ah, so that's the incantation… (I re-wrote that more times than I'd like to admit 😅)
So indent the case lines too, or leave it like in the suggestion?

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -291,17 +319,52 @@ private final class HTTPServerConnection {
...
+                in: application.threadPool
+            ).flatMap { _ in
+                return bootstrap.bind(unixDomainSocketPath: socketPath).map { channel in
+                    return .init(
+                        channel: channel,
+                        quiesce: quiesce,
+                        configuration: configuration
+                    )
+                }
+            }.flatMapErrorThrowing { error -> HTTPServerConnection in
```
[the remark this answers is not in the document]
dimitribouniol: Legitimate question (still getting my mind around futures here...):

Are these the same thing, because the result from the bootstrap is passed to the flatMap closure, so flatMap will call the next map closure once the results are returned?

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
gwynne: Leave it as in the suggestion - I will die on the case indentation hill, but not this particular day. Sadly, staying consistent with the style of the surrounding code trumps ideological debates 😅.

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -310,7 +373,26 @@ private final class HTTPServerConnection {
             promise.fail(Abort(.internalServerError, reason: "Server stop took too long."))
         }
         self.quiesce.initiateShutdown(promise: promise)
-        return promise.futureResult
+        return promise.futureResult.flatMapAlways { [configuration, channel] result in
```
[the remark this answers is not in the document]
dimitribouniol: I think at one point it warned about self, so I just passed the specific things I needed directly. I guess it won't cause a retain cycle anyways, so it's probably baseless...

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
gwynne: Yes, that's right - whether nested or directly chained, the result in this case is the same. That won't always be true, though - sometimes nesting is a cleaner way to pass on a useful result.

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
gwynne: Captures of this form are effectively meaningless - all you did is avoid the requirement of explicitly referencing `self`. The exact behavior depends on whether `configuration` and `channel` are value or reference types, but I can't see that it'd make a difference in this case either way.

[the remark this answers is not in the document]
gwynne: Actually, to call it "meaningless" is unfair - capture lists can be subtle and tricky. It would be more accurate to say that in this specific case, the captures have no practical functional impact on behavior.

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -310,7 +373,26 @@ private final class HTTPServerConnection {
...
+                threadPool.shutdownGracefully { error in
+                    if let error = error {
+                        promise.fail(error)
+                    } else {
+                        promise.succeed(())
+                    }
+                }
+                return promise.futureResult.flatMapThrowing { try result }
+            }.flatMapThrowing { try result }
+        }
```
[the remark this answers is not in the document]
dimitribouniol: Oh boy. So I wanted to remove the socket file after I was done with it (same process essentially as when preparing to start the server). That said, I didn't want it to be blocking ("blocking bad!"), so I fiddled with futures and thread pools ("async good!")... long story short, the application's threadPool is no good at this point (already being shutdown long before close is called), so I needed to make my own (resisting the urge to just make blocking, since we are shutting down the server anyways...), but you can't just shutdown a threadPool after scheduling some work to it (it ain't no common variety DispatchQueue, and Linux tests would fail half the time), so I landed on this solution which shuts it down after the future resolves itself 😅

Long story short, I'm open to suggestions...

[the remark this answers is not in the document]
gwynne: My suggestion is quite straightforward: Don't bother with the thread pool, or with `stat`-ing the file, or any of that. Just fire off an `unlink()` on whatever thread the shutdown is taking place on. You're shutting down anyway. It's arguable whether you need a thread pool at all for the prepare method even during startup, but you certainly don't need one for removing the file during shutdown, to the best of my knowledge.

Sources/Vapor/HTTP/Server/HTTPServer.swift
dimitribouniol: So ironically enough, that's what I had, but then I read horror stories that the bridged C string may not survive long enough in memory, and that if ARC deallocates anything in-between `lstat()` and reading `errno`, things kind of start falling apart at the seems, and `errno` now represents the results of the free or something. Needless to say, I'm not an expert on POSIX anything (initial version was just blindly unlinking whatever the user passed it 😛), and yet another long story short, I'd love to just use FileManager, but I figured there was a good reason that this file didn't import Foundation, so I opted not to.

So, happy to make these changes, but should I just switch to FileManager instead, and trust it to do the right thing? Or are we basically covering everything we need to?

(also, 👍 on double guard being strange)

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
dimitribouniol: Honestly, I didn't notice the surrounding code - you likely have a bigger enemy in Xcode's ⌃I functionality 🤣

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
dimitribouniol: Can I take it as a guarantee that this will only be called when the server has actually started? ie, if the server could not start, this code path would not be hit? I don't want to blindly unlock things we didn't actually create (though I realize another process could hijack the socket file and replace it with it's own, but I consider that undefined behavior)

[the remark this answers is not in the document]
dimitribouniol: (Note that the intended purpose of `prepareSocketFile()` is mostly on startup to help the user determine if they need to manually delete a file referenced as the socket path, or to determine if it is probably ok to do it for them (the "suggested" practice for unix services apparently, but potentially catastrophic if the service is ever run as root for some silly reason)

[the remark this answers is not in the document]
gwynne: It doesn't really matter - you don't have any way to tell whether or not you were the one who created the socket file at any given time. It is undefined behavior for something else to put something at the path that was specified for your socket configuration. The only way to get into trouble there is if you errored out because there was something in the way that wasn't a socket and then the shutdown code runs anyway, but I don't think that's how it flows. That being said, I can't claim to be sure - you should double-check.

Sources/Vapor/HTTP/Server/HTTPServer.swift
gwynne: Importing `Foundation` is generally avoided when possible, and `FileManager` is not really much of a win here. It's not really any less unwieldy than the POSIX calls, in a lot of ways.

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
dimitribouniol: This got a little silly with the `error as! IOError` casts:
```suggestion
                switch (error as? IOError)?.errnoCode {
                case .some(EADDRINUSE): throw UnixDomainSocketPathError.socketInUse(error as! IOError)
                case .some(ENOENT): throw UnixDomainSocketPathError.noSuchDirectory(error as! IOError)
                default: throw error
                }
```

... so I'm trying something in between:
```suggestion
                guard let ioError = error as? IOError else { throw error }
                switch ioError.errnoCode {
                case EADDRINUSE: throw UnixDomainSocketPathError.socketInUse(ioError)
                case ENOENT: throw UnixDomainSocketPathError.noSuchDirectory(ioError)
                default: throw ioError
```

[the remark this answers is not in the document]
gwynne: Here's an even more interesting (and potentially confusing) option:
```suggestion
                switch error as? IOError {
                case .some(let io) where io.errnoCode == EADDRINUSE: throw UnixDomainSocketPathError.socketInUse(io)
                case .some(let io) where io.errnoCode == ENOENT: throw UnixDomainSocketPathError.noSuchDirectory(io)
                default: throw error
                }
```

dimitribouniol:
Ok, I think I got most of this. I refactored the checks into an asynchronous method called `prepareSocketFileForBinding()`, intended to be used when starting the server, and a simpler synchronous method called `removeSocketFileIfPresent()` (which the async version uses under the hood), intended for cleanup. This way we still do some basic sanity checks/error reporting to the user before unlinking random files.

dimitribouniol:
Rebased off of master to keep up with changes there.

Sources/Vapor/Commands/ServeCommand.swift
```diff
@@ -44,8 +47,13 @@ public final class ServeCommand: Command {
         let port = signature.port
             // 0.0.0.0:8080, :8080, parse port
             ?? signature.bind?.split(separator: ":").last.flatMap(String.init).flatMap(Int.init)
-
-        try context.application.server.start(hostname: hostname, port: port)
+        let socketPath = signature.socketPath // /tmp/vapor.socket
+        
+        if let socketPath = socketPath, !socketPath.isEmpty {
```
tanner0101, line 52: The socket path should never be empty since this is an `@Option` and not a flag. We can assert / precondition this.

Sources/Vapor/Error/UnixDomainSocketPathError.swift
```diff
@@ -0,0 +1,41 @@
+import NIO
+
+/// Unix domain socket errors may be thrown when attepting to start the server from a unix domain socket path.
+public enum UnixDomainSocketPathError: AbortError {
```
tanner0101, line 4: This should be organized alongside the code that throws it.

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -108,6 +111,7 @@ public final class HTTPServer: Server {
         public var logger: Logger
 
         public init(
+            socketPath: String? = nil,
```
tanner0101, line 114: I'd put this after hostname and port.

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -291,17 +319,49 @@ private final class HTTPServerConnection {
...
+                return .init(
+                    channel: channel,
+                    quiesce: quiesce,
+                    configuration: configuration
+                )
+            }.flatMapErrorThrowing { error -> HTTPServerConnection in
+                quiesce.initiateShutdown(promise: nil)
+                
+                // Transform `bind(descriptor:ptr:bytes:): Address already in use (errno: 48)`/`No such file or directory (errno: 2)` into a form the user can test for
+                switch error as? IOError {
```
tanner0101, line 340: -1 to wrapping these errors. We don't do this in other places.

Sources/Vapor/Utilities/URI.swift
```diff
@@ -175,3 +175,66 @@ public struct URI: ExpressibleByStringLiteral, CustomStringConvertible {
...
+        guard
+            let originalURL = URL(string: self.string),
+            let scheme = originalURL.scheme,
+            scheme == "http+unix" || scheme == "https+unix",
+            let host = originalURL.host,
+            let baseURL = URL(string: "unix://\(host)")
+        else { return nil }
+        
+        // prepare a new URL with a base URL that contains the path to the socket file.
+        // FIXME: We likely won't need this anymore once https://github.com/swift-server/async-http-client/pull/228 is merged
```
tanner0101, line 236: Looks like https://github.com/swift-server/async-http-client/pull/228 was merged, can we remove this?

Sources/Vapor/Utilities/URI.swift
```diff
@@ -175,3 +175,66 @@ public struct URI: ExpressibleByStringLiteral, CustomStringConvertible {
         return String(self.string[start..<end])
     }
 }
+
+
+extension URI {
```
tanner0101, line 180: Do we need these extensions? What are they used for?

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -322,6 +387,64 @@ private final class HTTPServerConnection {
     }
 }
 
+/// Check for and remove a previously established unix domain socket file if one exists, as is standard practice for unix services.
+/// - Parameters:
+///   - configuration: A configuration with a socket path and logger.
+///   - eventLoop: An event loop to report the future on.
+///   - threadPool: A thread pool to use when performing IO.
+/// - Returns: A future representing the completion of the necessary checks.
+private func prepareSocketFileForBinding(for configuration: HTTPServer.Configuration, with eventLoop: EventLoop, in threadPool: NIOThreadPool) -> EventLoopFuture<Void> {
```
tanner0101, line 396: Is this something NIO's bootstrap should be doing instead?

Sources/Vapor/HTTP/Server/HTTPServer.swift
tanner0101: Also should be `unixDomainSocketPath`

Sources/Vapor/HTTP/Client/EventLoopHTTPClient.swift
```diff
@@ -16,7 +16,7 @@ private struct EventLoopHTTPClient: Client {
     ) -> EventLoopFuture<ClientResponse> {
         do {
             let request = try HTTPClient.Request(
-                url: URL(string: client.url.string)!,
+                url: client.url.socketURL ?? URL(string: client.url.string)!,
```
tanner0101, line 19: What's the reasoning behind this change?

Sources/Vapor/Commands/ServeCommand.swift
dimitribouniol: So this should be enough then?
```suggestion
        if let socketPath = socketPath {
```

Sources/Vapor/HTTP/Client/EventLoopHTTPClient.swift
dimitribouniol: This ensures that an old-style AsyncHTTPClient socket URL can be properly processed. Once https://github.com/swift-server/async-http-client/pull/228 is released, we likely won't need to to check it first. That said, I did notice that the internal URI handling doesn't fair too well for http+unix URLs (either because the host contains percent encoded tokens (`/` for the socket path), or the scheme contained a `+` — I don't remember exactly), so we may still need it for those reasons.

Sources/Vapor/HTTP/Server/HTTPServer.swift
dimitribouniol: I put it before because it took precedence over the hostname (which can still be used for HTTP and TLS reasons?), but I'm happy moving it anywhere.

Sources/Vapor/HTTP/Server/HTTPServer.swift
dimitribouniol: The idea was to help (new) users address why the server couldn't be started, without giving them some low level error they may not readily know what to do with. Happy to remove them if you thing they don't add much.

Sources/Vapor/HTTP/Server/HTTPServer.swift
dimitribouniol: I'm honestly not sure. Without it, this will likely appear broken to the user, since the general practice seems to be to just override the socket path when starting a service on unix-based tools. NIO could certainly do more to clean up its socket path on close, but I'm not sure if it should override it on binding, since I imagine in other cases a user of NIO (rather than Vapor) may not want that?

Sources/Vapor/Utilities/URI.swift
dimitribouniol: These are here to facilitate in tests and in other scenarios to create a URI based off of a socket path-based server. I added something similar to URLs to AsyncHTTPClient (https://github.com/swift-server/async-http-client/pull/235 - merged, but still unreleased), but Vapor's convenience methods largely don't use URLs, so something like this is still useful.

As mentioned in an above comment, `.socketURL` is necessary since the homegrown URI parsing seems to have trouble with either the scheme or the host, I don't quite remember which.

dimitribouniol: Actually, it shouldn't matter if we can't pull out the scheme or host from the URI, since it is just being passed wholesale to AsyncHTTPClient...

Sources/Vapor/Utilities/URI.swift
dimitribouniol: It's still not released, so we wouldn't be able to version bump it in the package - is that ok?

dimitribouniol:
@tanner0101 Thank you for your feedback! I responded to most of your comments with clarification - please confirm which direction you'd like to go in with the extra context, and I'll happily make the changes :)

Sources/Vapor/HTTP/Server/HTTPServer.swift
tanner0101: I agree it's helpful. My concern is consistency. If we wrap IOError here, we should do it in other places, too. While that may be a good idea, it's out of scope for this PR.

Sources/Vapor/Utilities/URI.swift
tanner0101: Perhaps this PR could depend on `.branch("master")` for the time being. Then we can wait to merge until tagged. I definitely want to avoid duping fixes in both places though.

Sources/Vapor/HTTP/Server/HTTPServer.swift
tanner0101: I think cleaning up the socket file after close is definitely worth adding to NIO. I think overwriting an existing file would be nice, too. If the application crashes, you will be left with the file even if you have cleanup code. If NIO declines to accept these additions, then we can reconsider implementing them here.

tanner0101:
Thanks, replied. Feel free to DM me in Vapor chat if you have additional questions.

tanner0101:
Closing for inactivity. Feel free to re-open once the comments are addressed. Thanks!

Track at https://github.com/vapor/vapor/issues/2226

stephen-turner:
@dimitribouniol Just wondering, were you planning to pick this up again or not? At the moment, the lack of unix domain socket support is preventing us from switching to vapor. So it would be helpful to know whether we can still look forward to it being added, or whether we should look for an alternative. Thanks.

dimitribouniol:
I was, as we have the same requirement... I'll probably get a cleaned up version of this branch (without the socket checks) up this weekend and get it back in the review queue. I still need to investigate where in NIO to add the socket file checks, but for now, you can grab that code from here and run it before connecting as a workaround.

stephen-turner:
Great, thanks!

dimitribouniol:
Hello all, this has been revived in #2471, except without the code to handle an existing socket path, since that would fit better in NIO directly. If you would like to use this is a real project though, until support is added to NIO, I would highly recommend implementing something similar to the following methods yourself in your app, so that you can remove any socket files before-hand, and clean them up on shutdown:

https://github.com/dimitribouniol/vapor/blob/backup/unix-domain-sockets/Sources/Vapor/HTTP/Server/HTTPServer.swift#L390-L446

PR #2403, opened 2020-06-20
Allow XCTAssertContent to rethrow errors

ileitch:
Enables the use of `XCTUnwrap` within the `XCTAssertContent` closure (#2403).


Joannis:
Was there a conscious reason that it wasn't throwing yet @tanner0101 ?

0xTim:
I see no reason why this can't be added

PR #2471, opened 2020-08-12
Support for Unix Domain Sockets

dimitribouniol:
Adds support for Unix Domain Sockets (#2471, #2226, #2371).

A server can be started using the `serve` command:
```fish
vapor run serve --unix-socket /tmp/vapor.socket
```

Alternatively, the app can be configured in code to connect to a socket:
```swift
app.http.server.configuration.unixDomainSocketPath = "/tmp/vapor.socket"
```

In either scenario, if a socket path is specified, it takes precedence over binding the server to a hostname/port. Note that TLS may be combined with this options if you wish the server to only listen to encrypted connections made to the socket file.

The socket file must not exist prior to starting the server. If one is left behind during an incomplete shutdown, however, and the app has permission to delete it, it will be overidden.

To assist with making connections to socket paths, `URI.Scheme` has a `.httpUnixDomainSocket` property. 

```swift
try app.client.get(.init(scheme: .httpUnixDomainSocket, host: "/tmp/vapor.socket", path: "/foo"))
```


Sources/Vapor/Utilities/URI.swift
```diff
@@ -175,3 +175,51 @@ public struct URI: ExpressibleByStringInterpolation, CustomStringConvertible {
...
+    }
+    
+    /// Create an HTTPS URI using a unix domain socket path.
+    /// - Parameters:
+    ///   - socketPath: The socket path the server is listening on.
+    ///   - path: The URI path to request from the server.
+    ///   - query: The URI path to request from the server.
+    ///   - fragment: The URI fragment to request from the server.
+    /// - Returns: A correctly formatted URI that encodes the socket path as the hostname.
+    public static func withSecureSocketPath(
```
tanner0101, line 210: Could `secure` be a flag instead? Like:
```swift
URI.unixDomainSocketPath("/tmp/vapor.socket", secure: true, path: "/foo")
```

tanner0101: Also how do other schemes, like `wss` work with this? Maybe we should let them pass the scheme they want to use?
```swift
URI.unixDomainSocketPath("/tmp/vapor.socket", scheme: "wss", path: "/foo")
```
That would convert to `wss+unix`.

Sources/Vapor/Commands/ServeCommand.swift
```diff
@@ -45,7 +48,11 @@ public final class ServeCommand: Command {
             // 0.0.0.0:8080, :8080, parse port
             ?? signature.bind?.split(separator: ":").last.flatMap(String.init).flatMap(Int.init)
 
-        try context.application.server.start(hostname: hostname, port: port)
+        if let socketPath = signature.socketPath {
+            try context.application.server.start(socketPath: socketPath)
+        } else {
```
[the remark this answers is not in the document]
dimitribouniol: It can be useful to still have a hostname depending on where-else it is used, I would assume, but for the most part this seems reasonable!

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -176,10 +181,32 @@ public final class HTTPServer: Server {
...
+            configuration.unixDomainSocketPath = nil
+        }
+        
+        try start(with: configuration)
+    }
+    
+    public func start(socketPath: String) throws {
+        // override the socket path to bind to
+        var configuration = self.configuration
+        configuration.unixDomainSocketPath = socketPath
```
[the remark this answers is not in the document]
dimitribouniol: `socketPath` is non optional here - would it be better for it to be?

Sources/Vapor/Utilities/URI.swift
dimitribouniol: I honestly don't know - AsyncHTTPClient will only work with `http+unix` and `https+unix`, so unless there is special routing internally to connect `wss+unix` through the right socket, it wouldn't do anything.

How are `wss://` URLs handled currently?

Sources/Vapor/Commands/ServeCommand.swift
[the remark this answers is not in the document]
dimitribouniol: Scratch that, we aren't passing the hostname through with the socket path anyways, and it's likely too specialized to pass the hostname in this way.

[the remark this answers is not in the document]
dimitribouniol: Handled these, along with a few other ambiguous cases:
```swift
switch (signature.hostname, signature.port, signature.bind, signature.socketPath) {
case (.none, .none, .none, .some): break // unix socket
case (.none, .none, .some, .none): break // bind ("hostname:port")
case (_,     _,     .none, .none): break // hostname / port
default: throw ServeCommandError.incompatibleFlags
}
```

I opted to throw an error here so we can test for these, as I also included tests that specifically check for invalid combinations.

Sources/Vapor/Utilities/URI.swift
dimitribouniol: I opted to have a scheme here for future use, but this likely will need to be a change made to websocket-kit, as it handles connections as the client, but does not use URI?

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
tanner0101: Ah I see now. Hmm... with the addition of a socket path option this protocol becomes somewhat convoluted. I'm not sure there is much we can do without breaking API though. At the least, could you include a `TODO` note in the `Server` protocol definition to refactor that API next major?

If you want, perhaps some backward compatible tweaking to the protocol would make it clearer. One idea:

```swift
protocol Server {
    func start(address: BindAddress?) throws

    @available(deprecated:)
    func start(hostname: String?, port: Int?) throws
}

enum BindAddress {
    case hostname(_ string: String, port: Int?)
    case unixDomainSocket(path: String)
}

extension Server {
    public func start() throws { try self.start(address: nil) }

    public func start(address: BindAddress?) throws {
        fatalError("Server does not support starting on: \(address)")
    }

    @available(deprecated:)
    public func start(hostname: String?, port: Int?) throws { 
        if let hostname = hostname else { 
            try self.start(address: .hostname(hostname, port: port))
        } else {
            try self.start(address: nil)
        }
    }
}
```

Sources/Vapor/Utilities/URI.swift
tanner0101: This is the method for connecting a WebSocket: https://github.com/vapor/websocket-kit/blob/master/Sources/WebSocketKit/WebSocket%2BConnect.swift#L1-L8

You can see the Vapor tests here: https://github.com/vapor/vapor/blob/master/Tests/VaporTests/WebSocketTests.swift#L11

Note that AHC client is not used for this. I do wonder how you could connect a WebSocket to a Vapor server running over unix domain sockets though.

Sources/Vapor/Commands/ServeCommand.swift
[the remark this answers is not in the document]
tanner0101: Looks great

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
dimitribouniol: This definitely seems like a good idea in the long run, so there is a clean API going forward. I'll make the change, though do you still want the TODO is this case?

Sources/Vapor/Utilities/URI.swift
dimitribouniol: I'm glad I was looking at the same files, despite never really using web sockets 😅

This is just speculation, but since the server upgrades the existing connection, nothing would need to change there... on the client however, it would need to make a connection to the right socket, make an http connection, wait for the upgrade, and then it would be good to go, no?

I would imagine only this line would need to change:
https://github.com/vapor/websocket-kit/blob/master/Sources/WebSocketKit/WebSocketClient.swift#L117

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
dimitribouniol: Do we want a TODO for `Configuration.hostname`/`.port`/`.unixDomainSocketPath` as well? The refactor surface is small, though many vapor apps likely tweak it like such:
```
public func configure(_ app: Application) throws {
    app.http.server.configuration.hostname = "127.0.0.1"
```
... so we can't outright deprecate it.

Would SemVer play nice adding a set/get to those properties that modifies a single `address` property instead?

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -171,15 +176,41 @@ public final class HTTPServer: Server {
         self.didShutdown = false
     }
     
+    @available(*, deprecated, renamed: "start(address:)", message: "Please use `start(address: .hostname(hostname, port: port))` instead")
     public func start(hostname: String?, port: Int?) throws {
-        // determine which hostname / port to bind to
+        try start(address: .hostname(hostname, port: port))
+    }
```
dimitribouniol, line 182: @tanner0101 Not sure if we need to keep this one around or not, per SemVer. The protocol has the same default implementation, but I remember someone mentioning that the signature may not match, or something along those lines...

Would you need a TODO to clean it up?

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -171,15 +176,41 @@ public final class HTTPServer: Server {
...
+            }
+        case .unixDomainSocket(let socketPath):
+            // override the socket path to bind to
+            configuration.unixDomainSocketPath = socketPath
+        }
+        
+        try start(with: configuration)
+    }
+    
+    private func start(with configuration: Configuration) throws {
```
dimitribouniol, line 204: @tanner0101 Not sure if we want to keep these as two methods now that the public start method have been consolidated. The only benefit at this point is that it does make it cleaner and easier to read, and _maybe_ easier to add to in the future, but the only change would be deleting 202-204. Up to you if you want to keep the private method or not.

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
tanner0101: TL;DR, yes there are lots of ways we could make this nicer in a backward compatible way. I don't think any of these are _requirements_ for this PR though, so don't feel obligated.

---

> Would SemVer play nice adding a set/get to those properties that modifies a single address property instead?

Yes it should be possible to have something like:

```swift
app.http.server.configuration.address = .hostname("127.0.0.1", port: 8080)
```

And then `configuration.hostname` / `configuration.port` are changed to be computed properties that set that one instead. 

The question then becomes whether or not to deprecate `hostname` and `port`. Given how prevalent they are, I think not. But if the new API is just as good it could be worth considering. `.hostname(..., port:)` is a little clunky. Maybe an `ExpressibleByStringLiteral` conformance would make it nicer:

```swift
app.http.server.configuration.address = "127.0.0.1:8080"
app.http.server.configuration.address = "/tmp/vapor.socket"
```

Bikeshedding: Maybe `app.http.server.configuration.bind = ...`? Then it's similar to the `--bind` flag from the CLI.

Sources/Vapor/Utilities/URI.swift
tanner0101: > I would imagine only this line would need to change:
> https://github.com/vapor/websocket-kit/blob/master/Sources/WebSocketKit/WebSocketClient.swift#L117

Yup that looks like it. Looks like that should probably be a separate issue then. 

The end goal though should be that HTTP client and WebSocket client work the same way. In other words, if you can do:

```swift
let res = try app.client.get(.unixDomainSocketPath(socketPath, path: "/foo")).wait()
```

Then you should be able to do:

```swift
WebSocket.connect(.unixDomainSocketPath(socketPath, path: "/foo"))...
```

Sources/Vapor/Commands/ServeCommand.swift
```diff
@@ -38,14 +41,25 @@ public final class ServeCommand: Command {
 
     /// See `Command`.
     public func run(using context: CommandContext, signature: Signature) throws {
+        switch (signature.hostname, signature.port, signature.bind, signature.socketPath) {
+        case (.none, .none, .none, .some): break // unix socket
```
tanner0101, line 45: You can do `.some(let socketPath)` here for example (and in the rest, too) and then not need to check `if let socketPath = ...` later.

Sources/Vapor/HTTP/Server/HTTPServer.swift
tanner0101: This one can be deleted as it will now be provided from the extension on `Server`.

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -171,15 +176,41 @@ public final class HTTPServer: Server {
...
+            // clear out the unix domain socket path if a new hostname/port is provided so it won't be used
+            if hostname != nil || port != nil {
+                configuration.unixDomainSocketPath = nil
+            }
+        case .unixDomainSocket(let socketPath):
+            // override the socket path to bind to
+            configuration.unixDomainSocketPath = socketPath
+        }
+        
+        try start(with: configuration)
```
tanner0101, line 201: Missing `self.`

Sources/Vapor/HTTP/Server/HTTPServer.swift
tanner0101: I think this could be cleaned up a little bit. Line 194 where `configuration.unixDomainSocketPath` is set to `nil` seems a bit hacky. But honestly not super critical. 

If I were to clean it up, I think I'd make `HTTPServer.Configuration` use the same "bind address" type, then it's jsut a matter of:

```swift
public func start(address: BindAddress?) throws {
    let address = address ?? self.configuration.address
}
```

Sources/Vapor/Server/Server.swift
```diff
@@ -1,11 +1,42 @@
...
 extension Server {
+    /// Start the server with its default configuration, listening over a regular TCP socket.
+    /// - Throws: An error if the server could not be started.
     public func start() throws {
-        try self.start(hostname: nil, port: nil)
+        try self.start(address: .hostname(nil, port: nil))
+    }
+    
+    /// A default implementation that throws an unimplemented assertion.
+    public func start(address: BindAddress) throws {
```
tanner0101, line 29: I think this is going to need to call into `start(hostname:port:)` for backward compatibility. For example:

```swift
public func start(address: BindAddress?) throws {
    switch address {
    case .none:
        try self.start(hostname: nil, port: nil)
    case .some(let address):
        switch address {
        case .hostname(let hostname, let port):
            try self.start(hostname: hostname, port: port)
        case .unixDomainSocket:
            fatalError("Starting with unix domain socket not supported, \(Self.self) must implement start(address:)")
        }
    }
}
```

You can test this by adding a backward compatibility test like:

```swift
struct OldServer: Server {
    // only implements the old requirement
    func start(hostname: String?, port: Int?) throws
}

try OldServer().start() // must still work
try OldServer().start(hostname: ..., port: ...) // must still work
try OldServer().start(address: .hostname(...)) // must work (since ServeCommand uses it)
try OldServer().start(address: .unixDomainSocket(...)) // ok if this doesn't work
```

Sources/Vapor/Utilities/URI.swift
dimitribouniol: If we make sure to specify the scheme, it should be pretty simple to achieve using the URI overload methods already there:
```swift
let res = try app.client.get(.unixDomainSocketPath(socketPath, scheme: "wss", path: "/foo")).wait()
```
(which I think is what you were going for)

Let's file separate issues for this then, one for adding socket path binding to `websocket-kit`, and one for adding a switch in the `.get()` method once that is in place (though I guess this one could be done earlier).

Sources/Vapor/Server/Server.swift
```diff
@@ -1,11 +1,42 @@
+// TODO: Remove these deprecated methods for the next major release.
 public protocol Server {
     var onShutdown: EventLoopFuture<Void> { get }
+    
+    /// Start the server with the specified address.
+    /// - Parameters:
+    ///   - address: The address to start the server with.
+    func start(address: BindAddress) throws
```
tanner0101, line 8: `address` should be optional here IMO. The bind address being passed here is specifically an override. If no override is passed, the configured values should be used.

Sources/Vapor/Utilities/URI.swift
```diff
@@ -175,3 +175,31 @@ public struct URI: ExpressibleByStringInterpolation, CustomStringConvertible {
...
+extension URI {
+    /// Create a URI using a unix domain socket path.
+    /// - Parameters:
+    ///   - socketPath: The socket path the server is listening on.
+    ///   - scheme: The scheme to use to connect with. Will use HTTP by default, but HTTPS is also currently supported.
+    ///   - path: The URI path to request from the server.
+    ///   - query: The URI path to request from the server.
+    ///   - fragment: The URI fragment to request from the server.
+    /// - Returns: A correctly formatted URI that encodes the socket path as the hostname.
+    public static func unixDomainSocketPath(
```
tanner0101, line 189: Bike shedding again, sorry: Having `unixDomainSocketPath(...:path:)` is kind of tripping me up. 

This is the current initializer:

```swift
struct URI {
    public init(
        scheme: String? = nil,
        host: String? = nil,
        port: Int? = nil,
        path: String,
        query: String? = nil,
        fragment: String? = nil
    )
}
```

What if we offered something like: 

```swift
extension URI {
    public struct Host: ExpressibleByStringLiteral {
        public static func unixDomainSocket(path: String) -> Self {
            .init(value: path.addingPercentEncoding(withAllowedCharacters: .urlHostAllowed))
        }
        public let value: String
    }

    public init(
        scheme: String? = nil,
        host: Host? = nil,
        port: Int? = nil,
        path: String,
        query: String? = nil,
        fragment: String? = nil
    )
}
```

Although, now that I think about it, this is assuming only socket paths need to be URL escaped. Why not just do this for all hosts?

```diff
struct URI {
    public init(
        scheme: String? = nil,
        host: String? = nil,
        port: Int? = nil,
        path: String,
        query: String? = nil,
        fragment: String? = nil
    ) {
-       self.host = host
+       self.host = host.addingPercentEncoding(withAllowedCharacters: .urlHostAllowed)
    }
}
```

Tests/VaporTests/ServerTests.swift
```diff
@@ -20,6 +20,98 @@ final class ServerTests: XCTestCase {
...
+                }
+                
+                XCTAssertEqual(ServeCommandError.incompatibleFlags, serveError)
+            }
+            app.shutdown()
+        }
+        
+        var app = Application(Environment(
+            name: "testing",
+            arguments: ["vapor", "serve", "--port", "8123", "--unix-socket", "/path/to/socket"]
```
tanner0101, line 63: ❤️ tests

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
dimitribouniol: I agree that it is perhaps escaping the scope of what this PR enables. For now, to make the API as unambiguous as possible, should we change `unixDomainSocketPath` to `address`, using a BindAddress instead, and swap the implementations for hostname and port to computed property ones? This way, there is little doubt which one will be used since configuring all three would result in a single value for `address`, and we keep the API simple to use. Would that be fine for now?

Also, I feel like although it would be more convenient to be able to use `ExpressibleByStringLiteral`, it would probably invite errors since the compiler can no longer make sure you are formatting stuff properly anymore. Small issue over all, but something I just thought about.

Sources/Vapor/Server/Server.swift
dimitribouniol: Would it ok for the backwards compatibility test to generate deprecation warnings?

Sources/Vapor/Server/Server.swift
dimitribouniol: Ah, I didn't catch that nuance. Will do.

Sources/Vapor/Utilities/URI.swift
dimitribouniol: I'm honestly not sure what the repercussions would be for regular unicode host names here, so I took the safe approach 😅 I'm fine going in either direction, so please let me know 🙂

Tests/VaporTests/ServerTests.swift
dimitribouniol: Got to get every possibility 😛

Sources/Vapor/Utilities/URI.swift
tanner0101: > Let's file separate issues for this then, one for adding socket path binding to websocket-kit, and one for adding a switch in the .get() method once that is in place (though I guess this one could be done earlier).

+1 to the websocket-kit issue. 

Regarding Client.get though, I think that will need to remain separate from WebSockets. `Client` is built on AHC which doesn't support WebSocket (on purpose). This is how Vapor's docs recommend you create a WebSocket connection: 

```swift
WebSocket.connect(to: "ws://echo.websocket.org", on: eventLoop) { ws in
    // Connected WebSocket.
    print(ws)
}
```

So as long as `WebSocket.connect` works with unix sockets we are good to go.

Sources/Vapor/Server/Server.swift
tanner0101: Maybe marking the test as `@available(deprecated:)` will silence the warnings? Or perhaps there's some other secret Swift attribute to silence warnings. If not, then having the tests generate a warning is not the end of the world.

Sources/Vapor/Utilities/URI.swift
tanner0101: I think let's try percent encoding all host names by default. This seems like the right thing to do and IIRC `Foundation.URL` has this behavior as well. It also prevents us from needing to introduce new API which is great.

Sources/Vapor/Utilities/URI.swift
dimitribouniol: Ah, I misread what you said. Would this mean that we would want to have `websocket-kit` be ok receiving requests to `http+unix`? Otherwise it seems like a lot of extra work on the end user to construct the right URL. That said, `websocket-kit` doesn't use vapor's `URI`, so I guess it can have it's own method that matches the name that vapor uses, but provides a `ws+unix` scheme... (food for thought 😅)

Sources/Vapor/Utilities/URI.swift
dimitribouniol: Ok, that sounds good.

One more consideration before I start ripping out all the cables: having the public API *does* document what Vapor can do, and creating HTTP(S) requests over a unix domain socket is not necessarily the most obvious thing that is possible, though it can be more efficient than using TCP for local connections on the same machine, so it should be encouraged when possible. Is there a good place we can document and tell users to use `http+unix` and `https+unix` (and eventually `ws(s)+unix`) when they want such functionality?

Another side to consider: as someone with dyslexia, I _constantly_ confused `http+unix` with `unix+http` when getting this started, and still look it up each time — although not an issue for most, having some sort of built in constants would be nice. Maybe we have a `Scheme` type expressible by a string literal that offers some built in constants? Any thoughts on how we can improve this?

Sources/Vapor/Utilities/URI.swift
tanner0101: > Ah, I misread what you said. Would this mean that we would want to have websocket-kit be ok receiving requests to http+unix?

Yup. 

> so I guess it can have it's own method that matches the name that vapor uses, but provides a ws+unix scheme... (food for thought 😅)

Yeah I guess. I'm not exactly sure how it will look yet either haha. I'm guessing it will be similar to whatever AHC does internally.

Sources/Vapor/Utilities/URI.swift
tanner0101: > Is there a good place we can document and tell users to use http+unix and https+unix (and eventually ws(s)+unix) when they want such functionality?

Agreed, we should explain all of this thoroughly. I think here is a good place: https://github.com/vapor/docs/blob/master/4.0/docs/server.md

Docblocks in the code are always good, too 👍 

> Maybe we have a Scheme type expressible by a string literal that offers some built in constants?

Big +1 to this. I'm a fan of replacing `String` with more helpful types wherever possible.

Although I'd say let's do that as a separate issue / PR. Then we can consider other places that could benefit from this as well 👍

dimitribouniol: \*already knee deep pulling cables…*     Sounds like a plan! 😄

dimitribouniol: (joking aside) Thank you for all your feedback! I'll make all the changes right now and push them.

Sources/Vapor/Server/Server.swift
dimitribouniol: Are you opposed to using a single switch like this?
```swift
public func start(address: BindAddress?) throws {
    switch address {
    case .none:
         try self.start(hostname: nil, port: nil)
    case .hostname(let hostname, let port):
         try self.start(hostname: hostname, port: port)
    case .unixDomainSocket:
         throw ServerStartError.unsupportedAddress(message: "Starting with unix domain socket path not supported, \(Self.self) must implement start(address:).")
   }
}
```

tanner0101: Oh nice I forgot that worked. Even better 👍

Sources/Vapor/HTTP/Server/HTTPServer.swift
dimitribouniol: Just a heads up that we'll need to keep this after all, as it turns out having the default implementation on the protocol along with the new `start(address:)` that calls in to this method will cause an infinite loop since Swift uses static dispatch here, and will prefer the default implementation of `start(address:)` rather than the one implemented by the conforming type.

The solution that I arrived at was to not have a default implementation for `start(hostname:port:)`, have the default implementation for `start(address:)` call into that method, and have the conforming types implement both until we can remove the deprecated one. Not ideal, but it runs now and supports anyone who may have conformed to Server until the next major version.

Sources/Vapor/Server/Server.swift
dimitribouniol: I wasn't sure it would even work - Swift surprises all of us sometimes 😅

Sources/Vapor/HTTP/Server/HTTPServer.swift
tanner0101: > will prefer the default implementation of start(address:) rather than the one implemented by the conforming type.

Hmm.. are you sure there wasn't a typo? AFAIK, Swift should always prefer a conforming type's impl over anything in an extension.

dimitribouniol: It seems that's not the case within the extension on the protocol unfortunately 😔
https://forums.swift.org/t/overriding-protocol-default-implementation/5147

I thought I double checked, but let me push a dummy branch so you can take a look.

dimitribouniol: Scratch that, it seems to be working now?? All I did was undo my changes... I'm going to guess Xcode was using a cached file somewhere 😖

tanner0101: Phew I was worried for a second haha. I just did a proof of concept in a Playground and it seems to work as I would expect. 

Screen Shot 2020-08-13 at 5 05 56 PM

dimitribouniol: Figured it out: I was marking the method with `mutating` since I was saving what was passed in to check if values ended up in the right place, causing it to miss the conformance:
```
        struct OldServer: Server {
            var onShutdown: EventLoopFuture<Void> {
                preconditionFailure("We should never get here.")
            }
            func shutdown() { }
            
            var hostname:String? = ""
            var port:Int? = 0
            // only implements the old requirement
            mutating func start(hostname: String?, port: Int?) throws {
                self.hostname = hostname
                self.port = port
            }
        }
```

Switching to a class fixes it, since the method does not need to be mutating...

Sources/Vapor/HTTP/Server/HTTPServer.swift
dimitribouniol: So this still needed to be expanded since just the port may have been overridden, but it does look more consistent now.

Sources/Vapor/HTTP/Server/HTTPServer.swift
dimitribouniol: Indeed, it was because I had to change the local method to be mutating to ensure the values were coming through properly. All is good now, thank you for making me revisit it!

dimitribouniol:
Ok, I think I got everything. Let me know if I should make any more changes!

Sources/Vapor/Utilities/URI.swift
```diff
@@ -18,9 +64,27 @@ public struct URI: ExpressibleByStringInterpolation, CustomStringConvertible {
...
+            host: host,
+            port: port,
+            path: path,
+            query: query,
+            fragment: fragment
+        )
+    }
+    
+    public init(
+        scheme: Scheme?,
```
dimitribouniol, line 79: I don't think this can replace the `init(scheme: String? = nil,` version due to signatures or what-not, but if it can, we can add a default value here again. Or maybe this one gains the default value, and the other one looses it? I'm not sure any more, it's late 😅

dimitribouniol:
Rebased off of the latest master.

Sources/Vapor/Server/Server.swift
```diff
@@ -1,11 +1,62 @@
...
+    ///   - hostname: The hostname to start the server with, or nil if the default one should be used.
+    ///   - port: The port to start the server with, or nil if the default one should be used.
+    @available(*, deprecated, renamed: "start(address:)", message: "Please use `start(address: .hostname(hostname, port: port))` instead")
+    public func start(hostname: String?, port: Int?) throws {
+        try self.start(address: .hostname(hostname, port: port))
     }
 }
+
+/// Errors that may be thrown when starting a server
+public enum ServerStartError: Error {
```
tanner0101, line 58: This can probably be internal, I don't think anyone will need to catch this.

Sources/Vapor/Commands/ServeCommand.swift
```diff
@@ -83,3 +97,9 @@ public final class ServeCommand: Command {
         assert(self.didShutdown, "ServeCommand did not shutdown before deinit")
     }
 }
+
+/// Errors that may be thrown when serving a server
+public enum ServeCommandError: Error {
```
tanner0101, line 102: Could this either be:
- nested under `ServeCommand`
- made `internal`
To avoid polluting public namespace.

Sources/Vapor/Utilities/URI.swift
```diff
@@ -1,6 +1,52 @@
 import CURLParser
 
 public struct URI: ExpressibleByStringInterpolation, CustomStringConvertible {
+    /// A URI's scheme.
+    public enum Scheme: ExpressibleByStringInterpolation {
```
tanner0101, line 5: I think a `struct` would be better here since there are so many possible schemes. Enums are slightly more performant (smaller memory footprint) but the inflexibility is not really worth it. 

Using a struct with a `String` value would make it more straightforward to add new cases in the future:

```swift
struct Scheme: ExpressibleByStringLiteral {
    static let http: Self = "http"
    ...
    let value: String
}
```

Sources/Vapor/Utilities/URI.swift
```diff
@@ -1,6 +1,52 @@
...
+        
+        /// HTTPS
+        case https
+        
+        /// HTTP over Unix Domain Socket Paths. The socket path should be encoded as the host in the URI, making sure to encode any special characters:
+        /// ```
+        /// host.addingPercentEncoding(withAllowedCharacters: .urlHostAllowed)
+        /// ```
+        /// Do note that URI's initializer will encode the host in this way.
+        case httpUDS
```
tanner0101, line 17:
```suggestion
        case httpUnixDomainSocket
```
Swift style recommends against abbreviation. This will be a long one, but luckily autocomplete helps us here haha.

Sources/Vapor/Utilities/URI.swift
tanner0101: I think this should be the `= nil` method and the other method should not have a default parameter? I'm not 100% sure though. We should have a unit test that ensures both the old and new way work correctly.

The reason I think the new one should be `= nil` is so that if you do `let url = URI()` you don't have to go through a layer of indirection.

Sources/Vapor/Utilities/URI.swift
```diff
@@ -18,12 +64,30 @@ public struct URI: ExpressibleByStringInterpolation, CustomStringConvertible {
...
+        query: String? = nil,
+        fragment: String? = nil
     ) {
         var string = ""
-        if let scheme = scheme {
+        if let scheme = scheme?.scheme {
             string += scheme + "://"
         }
-        if let host = host {
+        if let host = host?.addingPercentEncoding(withAllowedCharacters: .urlHostAllowed) {
```
tanner0101, line 90: Could we do this in `URI.Host`'s initializer? That way if you are using `URI.Host.http`, etc, they are pre-escaped which should be much faster.

Sources/Vapor/Utilities/URI.swift
tanner0101: Issue here: https://github.com/vapor/vapor/issues/2474

Sources/Vapor/Commands/ServeCommand.swift
dimitribouniol: By nested, do you mean as `ServeCommand.ServeCommandError` or `ServeCommand.Error: Swift.Error`?

tanner0101: The latter

Sources/Vapor/Utilities/URI.swift
dimitribouniol: I had it as the long one originally, and waffled a few too many times. I'll change it back 😉

Sources/Vapor/Utilities/URI.swift
dimitribouniol: I didn't notice a `URI.Host` - I'll take a look!

Sources/Vapor/Utilities/URI.swift
dimitribouniol: That makes sense! I'll add a unit test as well.

Sources/Vapor/Utilities/URI.swift
tanner0101: Oh wait, I was confused. Ignore this one. I got this and scheme mixed up.

Sources/Vapor/Server/Server.swift
dimitribouniol: image
One downside is that the tests don't import internal members - is it ok with you to simplify the test that it doesn't check for a specific error, but just that an error was thrown?

Sources/Vapor/Utilities/URI.swift
dimitribouniol: I imagine this is no good for backwards compatibility: 
image
Maybe scheme can contain an optional value, so passing nil here will use the existing string initializer? Any other ideas?

dimitribouniol: (Also, `URI()` is resolving to `public init(string: String = "/")`, likely because the path has no default on the other initializers.)

dimitribouniol: I pushed the version that I have (which seems to work) - let me know if you would like me to try a different way around this issue.

tanner0101: Yup looks great, thanks!

dimitribouniol:
Ready when you are! 🎉

dimitribouniol:
Removed `Note that making requests to other socket-path-based services requires AsyncHTTPClient 1.2.0 or newer.` from the description, since Vapor now requires it anyways, so it should update on user's repos automatically.

tanner0101:
These changes are now available in [4.29.0](https://github.com/vapor/vapor/releases/tag/4.29.0)

MrLotU:
Thanks @dimitribouniol :D

tanner0101:
Thanks for the great additions!

dimitribouniol:
@tanner0101 @MrLotU Thank you both for helping make this the best it could be!

PR #2486, opened 2020-08-31
Support .jsonAPI media types for decoding content

Joannis:
When communicating with API's following the JSON API format, these API's emit a content-type of `vnd.api+json`. This content type is always used with JSON formatted responses, although they follow a certain JSON format. I think it's therefore helpful to provide a default encoder/decoder for this media type.

I'm not sure if this would be a semver-patch or semver-minor as it's additive behaviour but doesn't touch public API.


Joannis:
@0xTim and @gwynne while looking through PRs I found my own one still open. Care for a small review?

gwynne (approved):
Seems good to me. I'd just do semver-patch for the version.

0xTim (approved):
LGTM

PR #2490, opened 2020-09-06
Validation of unrequired key in JSON string with a null value

baldoph:
Given this JSON string:
[{ "name": null }](url)

for the validations:
`validations.add("name", as: String.self, is: .count(1...255), required: false)`

Validating this string with the given validation would produce the error "name cannot be null"

even though `required` is `false`.

### DIFF
--- a/Sources/Vapor/Validation/Validation.swift
+++ b/Sources/Vapor/Validation/Validation.swift
@@ -13,7 +13,11 @@ public struct Validation {
                     result = ValidatorResults.Skipped()
                 }
             } catch DecodingError.valueNotFound {
-                result = ValidatorResults.NotFound()
+                if required {
+                    result = ValidatorResults.NotFound()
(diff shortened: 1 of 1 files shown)


baldoph:
Such strings with null values can be produced using a JSONEncoder encoding. I stumbled upon that error when trying to validate models directly (vapor 3 style) with an extension: 

```
extension Validatable where Self: Encodable {
    
    func validate() throws {
        let jsonData = try JSONEncoder().encode(self)
        guard let jsonString = String(data: jsonData, encoding: .utf8) else {
            throw ...
        }
        try Self.validate(json: jsonString)
        
    }
}

struct Model: Encodable, Validatable {}

try Model().validate()
```

baldoph:
Nevermind, I found the proper to validate optionals in the Vapor tests :P

PR #2496, opened 2020-09-18
Create separate HEAD routes instead of forwarding to GET route handler

owainhunt:
Instead of converting HEAD requests to GET requests, register a separate HEAD route for routes with all-constant paths that returns a 200 (vapor#2492)


0xTim:
@owainhunt this was closed by the branch rename script because I'm assuming it can't find a branch to merge onto. I think this is still a worthwhile addition, so if you want to reopen it pointing at `main` we can get it merged!

owainhunt:
@0xTim I may have removed the original branch. I still have the code somewhere so I'll recreate it

PR #2501, opened 2020-10-02
Add option to configure graceful shutdown timeout

wacumov:
Allows the graceful shutdown timeout to be configured instead of just default to 10 seconds: 

```swift
app.http.server.configuration.shutdownTimeout = .seconds(30)
```


fouadhatem:
@wacumov can you confirm if graceful shutdown still works for you on Linux on the `4.30.0` release onwards?

wacumov:
@fouadhatem yes, it seems that graceful shutdown _sometimes_ does not work on Linux.

fouadhatem:
@wacumov thanks for that. Do you have an idea what might be the cause? I'm assuming you're using a Mac with more 8GB or more of RAM?

wacumov:
@fouadhatem No, I have no idea about the cause. Yes, I use a Mac with 16GB of RAM. On Mac OS graceful shutdown always works. I've reproduced ignoring SIGTERM in Docker (but sometimes the signal not being ignored in Docker too and the app shutdowns gracefully).

0xTim (approved):
LGTM

calebkleveter (approved):

0xTim:
Original comment:
The graceful shutdown works great, but in some cases, the current hardcoded value of the timeout (`.seconds(10)`) can be not enough. For example, a WebSocket server should not be shut down forcefully if there are connected (and active) clients. 
Of course, to set a reasonable timeout, one should know the expected duration of the session. But sometimes the value is known and exceeds 10 seconds.

This PR allows setting a timeout that is more suitable for a specific case.
```swift
app.http.server.configuration.shutdownTimeout = .seconds(30)
```

PR #2510, opened 2020-10-18
Added empty LinuxMain.swift file so Vapor starts on Fedora/CentOS/RHEL

tachoknight:
I'm looking into Vapor for a web project and discovered it wouldn't start on Fedora due to a missing `Tests/LinuxMain.swift` file. I discovered just having file there is sufficient to make Vapor build and run without any additional changes. 

This PR just saves other folks the time to track down the issue.


fouadhatem:
I remember that there was a distant discussion about this exact same issue. I believe it was resolved by simply adding the `--enable-test-discovery` option when running the app. Is this not working on the RHEL Linux family?

tachoknight:
> I believe it was resolved by simply adding the `--enable-test-discovery` option when running the app. Is this not working on the RHEL Linux family?

I just tested it and yes, it does work but I didn't know about that flag and it's not mentioned anywhere in the project (I ran `ag enable-test-discovery` in the vapor directory). 

I figure that having an empty `LinuxMain.swift` gives a better new-user experience; it allows new folks like me to simply get the code, build and run with no issues whatsoever.

Out of curiosity, what was the discussion around this file?

fouadhatem:
The discussion started [here](https://forums.swift.org/t/pitch-possible-solutions-to-enable-the-removal-of-linuxmain-swift/25471) and continues [here](https://forums.swift.org/t/test-discovery-on-linux/26203). Adding that empty file is a workaround for simpler projects, but if you rely heavily on testing in your app, then it will be worthwhile to read the info in the links and they will get you up to speed.

tachoknight:
Funny enough, the last link ends in a bug report. :)

While I get that it may be superfluous for larger projects, just having 
the file present would allow someone like me, starting brand-new, with 
the intent of discovering and testing out Vapor without spending time 
trying to figure out why it isn’t building.

fouadhatem:
Swift on Linux is relatively new. Swift as a whole is developing at breakneck speed. There are bound to be teething problems. From a newcomers perspective yes I agree that it would make sense to have it there. However, the maintainers if the project probably would beed to take into consideration the bigger oictur

gwynne:
There was a goodly amount of back-and-forth on this when the `--enable-test-discovery` functionality on Linux was first introduced, especially when I PR'd the _removal_ of `LinuxMain.swift` from the template repo (which `vapor new` uses). IIRC, the overall consensus ended up being that forcing users to add the flag (which is expected become the default in a near-future version of Swift as of the time of this writing):

1. Was worth avoiding people accidentally not running any tests at all, and
2. Would help promote use of test discovery in general, which is much preferred in any event.

As such, Vapor considers `LinuxMain.swift` (and its compatriots) to be deprecated, and will not provide it.

It is unfortunate that there isn't more visibility in the docs that this flag is required - we'd gladly welcome a PR against https://github.com/vapor/docs to improve that situation.

PR #2540, opened 2020-12-14
RFC Compliant Cookie Parsing

t-ae:
⚠️ Vapor now has stricter cookie parsing. If you're relying on non-standard cookie parsing (like non-ascii characters) you'lll need to migrate off before updating ⚠️ 

Changes cookie name parsing to conform to [RFC6265](https://tools.ietf.org/html/rfc6265) and [RFC2616](https://tools.ietf.org/html/rfc2616).

- Some characters (!"#$%&'*+^`) are newly allowed
- Non-ASCII characters are banned

#2540


t-ae:
This PR contains the improvements I suggested in #2511.

I also renamed `isDirectiveKey` to `isTokenCharacter` to clarify the definition is from RFC.

I afraid that banning non-ASCII characters can cause problem.
`isTokenCharacter` is used in `parseDirectives`, and `parseDirectives` is also used for other headers.
As far as I can see, none of them allows non-ASCII character as directive key. Can anyone verify it?

Here's the list of headers which use `parseDirectives`:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

0xTim (approved):
Thanks!

tanner0101:
These changes are now available in [4.41.4](https://github.com/vapor/vapor/releases/tag/4.41.4)

PR #2553, opened 2021-01-14
Extract multipart encoding/decoding from Vapor into MultipartKit

JaapWijnen:
Extract previously embedded multipart encoding and decoding functionality into a standalone package, MultipartKit (#2553).

The public API of MultipartKit is exported so there is no need to import MultipartKit if you already import Vapor.

See: https://github.com/vapor/multipart-kit


siemensikkema:
@JaapWijnen could you update this with the new MultipartKit tag?

siemensikkema:
Suggested title: Extract multipart encoding/decoding from Vapor into MultipartKit
Suggested body:

> Extract previously embedded multipart encoding and decoding functionality into a standalone package, MultipartKit (#2553).
>
> The public API of MultipartKit is exported so there is no need to import MultipartKit if you already import Vapor.
>
> See: https://github.com/vapor/multipart-kit

JaapWijnen:
@siemensikkema updated the title as well, somehow that didn't persist, probably forgot to save ;) All ready for merging! :)

siemensikkema (approved):
LGTM!

PR #2575, opened 2021-03-02
Conform MultipartError to AbortError

### LINKED ISSUE #2573: `MultipartError` always has HTTP status of 500
Since `MultipartError` does not conform to `Abortable` (or whatever) it always reports 500, which is incorrect IMO.

It seems to me all should be 400s apart from the `nesting` one.

This is for decoding submitted multipart data of course.

0xTim:
Conform MultipartError to AbortError so that failing requests return a sensible status code to the client, instead of an internal server error.

Resolves #2573

### DIFF
--- a/Sources/Vapor/Multipart/AbortError+MultipartError.swift
+++ b/Sources/Vapor/Multipart/AbortError+MultipartError.swift
@@ -1,0 +1,12 @@
+import MultipartKit
+
+extension MultipartError: AbortError {
+    public var status: HTTPResponseStatus {
+        switch self {
+        case .nesting:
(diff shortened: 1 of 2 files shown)


gwynne (approved):
I think `.notImplemented` makes more sense than `.unprocessableEntity`, because the input _is_ valid and can be processed, the code to do it just isn't there.

siemensikkema (approved):

Joannis (approved):

tanner0101:
These changes are now available in [4.41.2](https://github.com/vapor/vapor/releases/tag/4.41.2)

PR #2576, opened 2021-03-04
Fix potential race condition when accessing Application.http.client.shared

gwynne:
The accessor for `Application.http.client.shared` was performing a lookup in the storage dictionary using double-checked locking, which is not actually thread-safe (despite all appearances to the contrary). It now unconditionally takes the lock before performing the lookup.

Applications making heavy use of `Request.client` or `Application.client` may experience a performance regression from this change.

### DIFF
--- a/Sources/Vapor/HTTP/Client/Application+HTTP+Client.swift
+++ b/Sources/Vapor/HTTP/Client/Application+HTTP+Client.swift
@@ -17,25 +17,21 @@ extension Application.HTTP {
         let application: Application
 
         public var shared: HTTPClient {
+            let lock = self.application.locks.lock(for: Key.self)
+            lock.lock()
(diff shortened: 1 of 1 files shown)


gwynne:
@weissi Can you double-check my understanding on this one? 😰

weissi:
@gwynne Correct, double-checked locking is not legal in Swift (and also not in most other languages).

0xTim (approved):
LGTM

tanner0101:
These changes are now available in [4.41.3](https://github.com/vapor/vapor/releases/tag/4.41.3)

tanner0101:
These changes are now available in [4.41.3](https://github.com/vapor/vapor/releases/tag/4.41.3)

PR #2585, opened 2021-03-17
Avoid a `try!` and an IUO in `XCTVaporTests`

### LINKED ISSUE #2582: XCTVaporTests.setUp should be setUpWithError
I notice that `XCTVaporTests` implements `setUp` and then has a `try!`.

This should be `setUpWithError` and then losing the bang.

I have not made the PR since maybe you don’t want to do this:

* Possibly `setUpWithError` is newer than the version of XCTest Vapor supports
* Might not be a semantic version major change 
    * Depends if *both* are called
    * There are ordering considerations if both are called
    * OTOH this is testing code so maybe doesn’t matter).

However will PR if ok.

Edit: additionally if we do this `tearDown` should do `app?.shutdown()` since `app` could be `nil` if it throws.

Edit: code reference: https://github.com/vapor/vapor/blob/main/Sources/XCTVapor/XCTVaporTests.swift

Edit: `app` should be set back to `nil` after `shutdown` also really.

mxcl:
Improves `XCTVaporTests` by removing `try!` and an implicitly unwrapped optional. (#2585)

Resolves #2582

### DIFF
--- a/Sources/XCTVapor/XCTVaporTests.swift
+++ b/Sources/XCTVapor/XCTVaporTests.swift
@@ -5,13 +5,14 @@ public var app: (() throws -> Application) = {
 open class XCTVaporTests: XCTestCase {
     open var app: Application!
     
-    open override func setUp() {
+    open override func setUpWithError() throws {
         super.setUp()
-        self.app = try! XCTVapor.app()
+        self.app = try XCTVapor.app()
     }
     
     open override func tearDown() {
         super.tearDown()
-        self.app.shutdown()
+        self.app?.shutdown()
+        app = nil
     }
 }


0xTim (approved):
Thanks!

0xTim:
Original comment:

* After analysis this seems a breakage safe change. The XCTest runner calls `setUpWithError`, then `setUp`.
* `app?` since otherwise `tearDown` crashes the whole suite should `app()` throw or any other reason cause a user to not call `super`
* `app = nil` since otherwise a user that fails to call `super` in some circumstances could accidentally use the `app` from a previous run potentially resulting in scenarios as bad as masking failing tests

valeriomazzeo:
FYI - this has broken thousands of our unit tests that were all written in this form:

```swift
class MyTests: XCTVaporTests {

    override func setUp() {
        XCTVapor.app = { Application(.testing) }
        super.setUp()
        ...
```

🤔

mxcl:
You should probably set-up the `XCTVapor.app` in the class function for the test case:

```
class Tests: XCTVaporTests {
    override class func setUp() {
       XCTVapor.app = …
    }
}
```

But yeah this is unfortunate. In theory we could check for that scenario, but it seems this is more a symptom of the Vapor docs not describing how you should set up `XCTVapor.app` leading to undefined behavior for how this global variable will be used.

The fix that was applied is still more correct than it was, but I guess this is proof that once something is broken (but people have worked around it) you shouldn't change it without a major bump.

mxcl:
Proposal: https://github.com/mxcl/vapor/commit/85cf8476ce602215c0fa9673ebe943b01d8389a5

0xTim:
I've confirmed that @mxcl's proposal makes a basic test case pass, so if you'd like to PR it that would be awesome.

I agree that this is fundamentally a documentation issue and ultimately Vapor's fault. I admit I didn't even know `XCTVaporTests` existed since it's not mentioned in the docs and I've always set up the application myself manually (though with my use cases it tends to be wrapped in several layers of a `TestWorld` and `TestWorld.Context`).

PR #2587, opened 2021-03-18
Restore previous XCTVaporTests behavior if necc.

mxcl:
Some people were relying on the previous call order by coincidence and we should avoid punishing them.

Also fixes calling wrong `super` (whoops!).

Refs #2585

### DIFF
--- a/Sources/XCTVapor/XCTVaporTests.swift
+++ b/Sources/XCTVapor/XCTVaporTests.swift
@@ -1,26 +1,42 @@
+/**
+ We recommend configuring this in your test’s `class func setUp()`
+ */
 public var app: (() throws -> Application)!
 
 open class XCTVaporTests: XCTestCase {
     open var app: Application!
 
+    open override func setUpWithError() throws {
+
+        // The XCTest runner calls this function before setUp()
+
+        try super.setUpWithError()
+
+        // this optional check due to prior usage by users
+        // see: https://github.com/vapor/vapor/pull/2585#issuecomment-802144636
+        if let _app = XCTVapor.app {
+            self.app = try _app()
+        }
+    }
+
     open override func setUp() {
+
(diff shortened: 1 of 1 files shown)


mxcl:
Inverted the order of the two setUp functions to reflect the runner’s call order.

mxcl:
Diff is easier to read in the first commit, second reorganizes code but changes nothing.

0xTim (approved):
Thanks for this!

valeriomazzeo:
this still reaches `fatalError` when using the following code and running more than one test at the same time:

```swift
class MyTests: XCTVaporTests {

    override func setUp() {
        XCTVapor.app = { Application(.testing) }
        super.setUp()
        ...
```

mxcl:
I can’t see how this would occur bar some race condition, if you figure out the issue with the current code that is fixable please post about it. Otherwise I cannot see how the new code can behave differently to the old code for your usage.

valeriomazzeo:
@mxcl - no race condition at all, everything happens on the main thread.
This is the call order:
1. first test started
2. setUp
3. tearDown
4. first test passed
5. second test started
6. setUpWithError is called that sets `self.app`
7. setUp is call by `super.setUp` in the second test execution and it crashes because `self.app` is not `nil`

valeriomazzeo:
according to Apple documentation https://developer.apple.com/documentation/xctest/xctest/3521150-setupwitherror

`setUpWithError` should always be called first, but for some reasons it's not; the first time.

valeriomazzeo:
Because the first time `setUpWithError` is called, the app generator function is not assigned. Once the second test runs, `setUpWithError` sets `self.app` and subsequently `setUp` fails because `self.app` is not `nil`.

mxcl:
oof. my bad.

PR #2595, opened 2021-03-31
Expose the Channel's `ByteBufferAllocator` in `Request` and `Response`

### LINKED ISSUE #2588: Stop creating ByteBufferAllocators
We should stop creating our own ByteBufferAllocators and use the one from the `Channel`

0xTim:
Exposes a `ByteBufferAllocator` on `Request` and `Response` for users to use if needed.

This also migrates all the uses of `ByteBufferAllocator` to use the channel's one wherever possible to avoid creating them every time.

Resolves #2588


Sources/Vapor/Response/Response.swift
```diff
@@ -111,6 +111,8 @@ public final class Response: CustomStringConvertible {
             // ignore since Request is a reference type
         }
     }
+
+    private var byteBufferAllocator: ByteBufferAllocator
```
Joannis, line 115: Can we not get the ByteBufferAllocator from the `Body`?

Joannis (commented):
LGTM.

Sources/Vapor/Response/Response.swift
0xTim: Done

Joannis (approved):

PR #2596, opened 2021-03-31
Add stack trace to `RouteNotFound` error to improve `DebuggableError` conformance

### LINKED ISSUE #2532: RouteNotFound error does not provide stack trace
RouteNotFound does not fully implement `DebuggableError` protocol. stackTrace method is absent. So error context is not clear. e.g bugsnag `https://github.com/nodes-vapor/bugsnag` does not attach stack trace for `RouteNotFound` error.

### Steps to reproduce
Open any unexistent page, for this case `RouteNotFound` error is thrown.

### Expected behavior
For better understanding error context `stackTrace` property of `DebuggableError` should be implemented like for Abort(.notFound)

### Actual behavior
`stackTrace` is nil

### Environment

* Vapor Framework version:
4.35.0

0xTim:
Adds a stacktrace to the `RouteNotFound` error so that error reporting tools have access to it.

Resolves #2532﻿

### DIFF
--- a/Sources/Vapor/Responder/DefaultResponder.swift
+++ b/Sources/Vapor/Responder/DefaultResponder.swift
@@ -120,7 +120,13 @@ private struct NotFoundResponder: Responder {
     }
 }
 
-struct RouteNotFound: Error { }
+struct RouteNotFound: Error {
+    let stackTrace: StackTrace?
+
+    init() {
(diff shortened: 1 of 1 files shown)


madsodgaard (approved):
lgtm:)

PR #2597, opened 2021-04-03
Add support for configurable `Router`

maciesielka:
Adds support for a configurable `Router` type after noticing some stale conversation over at https://github.com/vapor/routing-kit/issues/92 discussing exactly this.

The changes here follow conventions for other types of app-level configurations, such that users can provide a new router instance like so:
```swift
app.routers.use(.other)
```

This is defaulted to the existing Trie implementation for backwards compatibility.


Sources/Vapor/Routing/AnyRouter.swift
```diff
@@ -0,0 +1,55 @@
+/// Type-erased `Router`
+public final class AnyRouter<Output>: Router {
```
maciesielka, line 2: This probably belongs in https://github.com/vapor/routing-kit. (Am happy to PR that over there if folks think these changes make sense in the first place)

0xTim:
So at an initial glance this looks good as is a worthwhile addition to Vapor. Why have you done down the type-erased route instead of sticking with the `Router` protocol? (It's likely I'm just missing something obvious from a skim read rather than this being a blocker but the protocol would fit better with the rest of Vapor though I suspect this is more a compiler limitation). I do agree that `AnyRouter` should live in `routing-kit`

maciesielka:
> Why have you done down the type-erased route instead of sticking with the `Router` protocol?

Main reason is that since `Router` has associated types, we can't actually store an instance of it to use later on without type-erasure because we can't name that type in Swift, particularly when `DefaultResponder` would like to use its own private `Output` type.

Since `DefaultResponder` is the thing that wants to store a router instance, I did explore one alternative of adding a new `Responder` type, something like `RoutingResponder` that's generic over a supplied `Router` type. I think that still requires a similar kind of `RouterFactory` if Vapor wants to maintain control of the `Output` type and the route registration. Definitely possible I didn't use the generics properly in those explorations, but even then it felt like the end goal of a user supplying a new `Responder` to achieve a new routing style seemed less intuitive (and maybe too tight a coupling) than reaching for `app.router.use(...)` like we can do more easily with the type-erased version.

Sources/Vapor/Routing/AnyRouter.swift
maciesielka: Removed in ef3b86a

PR #2598, opened 2021-04-07
Change Vapor's logging levels for errors.

### LINKED ISSUE #2586: Review Vapor's log level usage
Most alerting systems relying on analysing the logs produced by the application. Many 5XX errors are signs that something is wrong and anything logged at `error` or `critical` can be cause for paging people (and waking them up!). We should go through the framework and ensure that we aren't logging anything at `error` or `critical` unless absolutely necessary (for instance, failing to bind to a port or start).

We might want to provide some configuration so that when in `development` mode we log errors at `error` to easily catch bugs at development time, but certainly when running in `production` most server errors should be `warning` or below

madsodgaard:
> ⚠️ If you're relying on Vapor logging errors at the `.error` level you should review this release to ensure that you don't start losing errors. ⚠️ 

As per the [log level guidelines from the SSWG](https://github.com/swift-server/guides/pull/19) this changes many of the errors logged in Vapor to use the `.warning` level instead of `.error`. (#2598)

Fixes #2586.

### DIFF
--- a/Sources/Vapor/Logging/Logger+Report.swift
+++ b/Sources/Vapor/Logging/Logger+Report.swift
@@ -24,20 +24,25 @@ extension Logger {
         case let abort as AbortError:
             reason = abort.reason
             source = nil
-            level = .error
+            
+            if (500...599).contains(abort.status.code) {
+                level = .error
+            } else {
+                level = .warning
+            }
         case let localized as LocalizedError:
             reason = localized.localizedDescription
             source = nil
-            level = .error
+            level = .warning
         case let convertible as CustomStringConvertible:
             reason = convertible.description
             source = nil
-            level = .error
+            level = .warning
         default:
             reason = "\(error)"
             source = nil
-            level = .error
-        }
+            level = .warning
+        
         self.log(
             level: level,
             .init(stringLiteral: reason),
--- a/Sources/Vapor/Logging/Logger+Report.swift
+++ b/Sources/Vapor/Logging/Logger+Report.swift
@@ -42,6 +42,7 @@ extension Logger {
             reason = "\(error)"
             source = nil
             level = .warning
+        }
         
         self.log(
             level: level,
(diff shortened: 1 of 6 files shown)


0xTim:
Thanks for this! There are a couple of other places in vapor/vapor that we should change in this PR as well. All the `DotEnvFile+load.swift` errors (both the deprecated and undeprecated ones) should only log at warning. I think the HTTPServer logs are fine at error since they only occur at the start and end and point to serious errors. Storage errors should be warnings as well

Sources/Vapor/Logging/Logger+Report.swift
```diff
@@ -24,20 +24,25 @@ extension Logger {
         case let abort as AbortError:
             reason = abort.reason
             source = nil
-            level = .error
+            
+            if (500...599).contains(abort.status.code) {
+                level = .error
+            } else {
+                level = .warning
+            }
```
0xTim, line 32: I think this should probably be logged as a warning for all cases. There are several cases where an internal server error may occur (such as a dropped connection to a database) that may automatically be retried and be successful.  Anything logged at an error level should basically be considered something bad enough to wake people up. Since we can't work in any kind of flag for this, or configuration to choose which level, we should default to `.warning`. Anyone throwing `Abort(.internalServerError)` can alway add their own `logger.error` messages if they need.

However, this is a fairly big breaking change so I think we should throw this out to the community

ktoso: Yeah your wording is right @0xTim I think.

For the author or the PR, here is some guidance we're working on: https://github.com/swift-server/guides/pull/19

Logging an "error" is very very very dramatic, and frameworks should avoid doing so if possible, especially "on each request"; A warning would be better.

jdmcd: How does this change permeate into end-projects? Is it only applicable if users are using the `ErrorMiddleware` or does it happen no matter what? I agree that it should be a warning for all cases, I'm just trying to wrap my head around how that changes output in projects specifically.

0xTim: No it could affect anyone using `logger.report`. `ErrorMiddleware` does and _most_ people won't but it could affect them if they've copied the `ErrorMiddleware`

jdmcd: Got it, that makes sense. I am +1 then!

jordanebelanger: > I think this should probably be logged as a warning for all cases. There are several cases where an internal server error may occur (such as a dropped connection to a database) that may automatically be retried and be successful. Anything logged at an error level should basically be considered something bad enough to wake people up. Since we can't work in any kind of flag for this, or configuration to choose which level, we should default to `.warning`. Anyone throwing `Abort(.internalServerError)` can alway add their own `logger.error` messages if they need.
> 
> However, this is a fairly big breaking change so I think we should throw this out to the community

I would not say I am in disagreement here, but would like to add more to the discussion as to why perhaps it's worth logging internal errors at the error level.

To continue on the dropped connection example, one dropped connection from time to time is no big deal, but if you drop database connections often this could be the sign of a more serious, potentially developing issue. Now if we are talking about common 400es such as a validation or not found error, a rising amount of these should be of no concern and correlate with the amount of traffic you get. On the other hand, your internal error rate should remain constant regardless of traffic. 

Large applications usually rely on time series based log analysis so that developers can be alerted when say, 10 errors, reported at the error level or higher, happen within a 5 minutes timeframe. If all non curated request errors are logged at the warning level, this type of analysis will be more difficult and require a precise, more vapor focused, integration on the part of the devops side.

madsodgaard: After some more time to think about it, I am for logging `AbortError` as `warning`. `AbortError` is really a user-facing error, which is used to format an error into a HTTP response that the client understands, not necessarily the developer. I think @0xTim is right in that the developer should really be logging any actual internal errors themselves using `error` besides throwing or returning an `AbortError`.

jordanebelanger: Ya I see it now, @0xTim is right. I did not see before that this was in the context of logging `AbortError` specifically. My bad.

0xTim (approved):
LGTM thanks!

PR #2600, opened 2021-04-08
Change `AbortError` 5xx logging to always be `warning`

madsodgaard:
Changes logging level for `AbortError` to always be `warning` regardless of status code. (#2600)


0xTim (approved):
LGTM

PR #2601, opened 2021-04-08
Add expiration time to `Cache` protocol and in-memory cache.

madsodgaard:
Adds expiration time functionality to the `Cache` protocol and the default in-memory cache.

An expiration time can be set on a cache entry like so:
```swift
app.cache.set("username", to: "mads", expiresIn: .seconds(30))
```

### DIFF
--- a/Sources/Vapor/Cache/MemoryCache.swift
+++ b/Sources/Vapor/Cache/MemoryCache.swift
@@ -34,6 +34,16 @@ private struct MemoryCacheKey: LockKey, StorageKey {
 }
 
 private final class MemoryCacheStorage {
+    struct CacheEntryBox<T> {
+        var expiresAt: Date?
+        var value: T
+        
+        init(_ value: T) {
+            self.expiresAt = nil
+            self.value = value
+        }
+    }
+    
     private var storage: [String: Any]
     private var lock: Lock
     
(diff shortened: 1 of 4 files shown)


Sources/Vapor/Cache/MemoryCache.swift
```diff
@@ -79,10 +100,15 @@ private struct MemoryCache: Cache {
     }
     
     func set<T>(_ key: String, to value: T?) -> EventLoopFuture<Void>
-        where T: Encodable
+        where T : Encodable
```
0xTim, line 103: Not - remove the space between the T and :

Tests/VaporTests/CacheTests.swift
```diff
@@ -1,19 +1,26 @@
...
         
         try XCTAssertNil(app.cache.get("foo", as: String.self).wait())
         try app.cache.set("foo", to: "bar").wait()
         try XCTAssertEqual(app.cache.get("foo").wait(), "bar")
+        
+        // Test expiration
+        try app.cache.set("foo2", to: "bar2", expiresIn: .seconds(1)).wait()
+        try XCTAssertEqual(app.cache.get("foo2").wait(), "bar2")
+        sleep(1)
+        try XCTAssertEqual(app.cache.get("foo2", as: String.self).wait(), nil)
```
0xTim, line 16: Use `XCTAssertNil` here for consistency

0xTim (changes requested):
Couple of nits and then we're good to go!

0xTim (approved):
LGTM!

PR #2603, opened 2021-04-11
Fix parsing complex accept headers

### LINKED ISSUE #2283: Complex Accept header parsing fails
We try to differentiate between a request from the browser or a REST client, so we are able to provide the appropriate response (HTML vs. JSON).

We look at the "Accept" header to analyse the accepted answer.

```
   func index(req: Request) throws -> EventLoopFuture<Response> {
        let objects = SomeModel.query(on: req.db).all()
        if req.headers.accept.mediaTypes.contains(.html) {
           // Return HTML result
           return req.view.render("index")
        }
        else {
            // E.g. return JSON
            return objects.encodeResponse(for: req)
        }
    }
```

`req.headers.accept.mediaTypes.contains(.html)` seems not to parse the Accept header correctly. Using a basic Accept header like `Accept: text/html` will yield the correct result. But more complex headers like Safari's `Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8` fail to parse correctly.

`req.headers.accepts` is an empty array.
### Steps to reproduce

 - Use the above snippet for a route. 
 - Send a basic `Accept` header (`Accept: text/html`). This works fine.
 - Send a more complex `Accept` header, which expects a HTML response (e.g. Safari's `Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8`). You will receive a JSON response.

### Expected behavior

 - The request with a more complex `Accept` header (which in my case expects a HTML response) should be mapped to the correct media type.

PS:  I'm currently using this workaround 

`if req.headers["Accept"].first?.contains("html") ?? false {...` 

### Environment

Xcode 11.4

* Vapor Framework version: Vapor 4 RC 3.12
* Vapor Toolbox version: vapor-beta
* OS version: macOS 10.15.4

schwmi:
Fixes an issue when parsing complex `Accept` headers, e.g. `text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8` The accept header should now be correctly parsed and all accepted types available. Also fixes an issue in `HTTPMediaTypePreference` when comparing two media types. (#2603)

Fixes #2283

### DIFF
--- a/Sources/Vapor/HTTP/Headers/HTTPHeaders+Directive.swift
+++ b/Sources/Vapor/HTTP/Headers/HTTPHeaders+Directive.swift
@@ -101,11 +101,11 @@ extension HTTPHeaders {
                 if self.current.first == .semicolon {
                     self.pop()
                 }
-            } else if let comma = self.current.firstIndex(of: .comma) {
-                value = self.pop(to: comma)
-            } else if let semicolon = self.current.firstIndex(of: .semicolon) {
-                value = self.pop(to: semicolon)
-                self.pop()
+            } else if let separatorMatch = self.firstIndex(matchingAnyOf: .comma, .semicolon) {
+                value = self.pop(to: separatorMatch.index)
+                if separatorMatch.matchedCharacter == .semicolon {
+                    self.pop()
+                }
             } else {
                 value = self.pop(to: self.current.endIndex)
             }
@@ -138,6 +138,19 @@ extension HTTPHeaders {
                 } else if !character.isTokenCharacter {
                     return nil
                 }
+            }
+            return nil
+        }
+
+        /// Returns the first index matching any of the passed in Characters, nil if no match
+        private func firstIndex(matchingAnyOf characters: Character...) -> (index: Substring.Index, matchedCharacter: Character)? {
+            guard characters.isEmpty == false else { return nil }
+
+            for index in self.current.indices {
+                let character = self.current[index]
+                guard let matchedCharacter = characters.first(where: { $0 == character }) else { continue }
+
+                return (index, matchedCharacter)
             }
             return nil
         }
--- a/Sources/Vapor/HTTP/Headers/HTTPHeaders+Directive.swift
+++ b/Sources/Vapor/HTTP/Headers/HTTPHeaders+Directive.swift
@@ -101,11 +101,11 @@ extension HTTPHeaders {
                 if self.current.first == .semicolon {
                     self.pop()
                 }
+            } else if let comma = self.current.firstIndex(of: .comma) {
+                value = self.pop(to: comma)
             } else if let semicolon = self.current.firstIndex(of: .semicolon) {
                 value = self.pop(to: semicolon)
                 self.pop()
-            } else if let comma = self.current.firstIndex(of: .comma) {
-                value = self.pop(to: comma)
             } else {
                 value = self.pop(to: self.current.endIndex)
             }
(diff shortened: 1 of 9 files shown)


Sources/Vapor/HTTP/Headers/HTTPMediaTypePreference.swift
```diff
@@ -43,9 +43,9 @@ extension Array where Element == HTTPMediaTypePreference {
             if aq == bq {
                 return .orderedSame
             } else if aq > bq {
-                return .orderedAscending
-            } else {
                 return .orderedDescending
+            } else {
+                return .orderedAscending
```
schwmi, line 48: This was twisted - as `aq` is the left hand side input parameter the should return  `.orderedDescending` when larger than the right hand side.

Sources/Vapor/HTTP/Headers/HTTPHeaders+Directive.swift
```diff
@@ -101,11 +101,11 @@ extension HTTPHeaders {
                 if self.current.first == .semicolon {
                     self.pop()
                 }
-            } else if let semicolon = self.current.firstIndex(of: .semicolon) {
-                value = self.pop(to: semicolon)
-                self.pop()
-            } else if let comma = self.current.firstIndex(of: .comma) {
-                value = self.pop(to: comma)
+            } else if let separatorMatch = self.firstIndex(matchingAnyOf: .comma, .semicolon) {
```
schwmi, line 104: Instead of always preferring `semicolon` first we also check if there are `comma` along the way.

Tests/VaporTests/HTTPHeaderTests.swift
```diff
@@ -108,6 +108,26 @@ final class HTTPHeaderTests: XCTestCase {
         XCTAssertEqual(headers.forwarded.first?.by, "203.0.113.43")
     }
 
+    func testAcceptType() throws {
+        var headers = HTTPHeaders()
+
+        // Simple accept type
+        do {
+            headers.replaceOrAdd(name: .accept, value: "text/html")
```
0xTim, line 116: Just for complete sanity and to avoid future regressions could you add an assert to these tests that ensures that the accepted media types match the ones in the header. E.g.

```swift
XCTAssertTrue(headers.accept.mediaTypes.contains(.html))
```

0xTim (changes requested):
Thanks for doing this! Just a couple of changes around shoring up the tests and we're good to go!

Tests/VaporTests/HTTPHeaderTests.swift
schwmi: Thanks for the feedback. I've added some cases + removed the catchall from the second, more complex test to allow testing for concrete types…

0xTim:
Original comment:

Fixes #2283

This PR fixes an issue that parsing more complex accept headers like text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8 failed.
Reason was that during directive parsing the semicolon was parsed in greedy way, preferred over comma. (.e.g parsing the mentioned header would lead to "text/html,application/xhtml+xml,application/xml;q=0.9" as single directive). I fixed that by also considering commas when parsing the next semicolon.
As this directive parsing code is used also for other header types I am not sure if it suits all other needs, or if some other logic relies on the former behaviour - all existing unit tests still pass (but the test coverage isn't that high)

When adding unit tests for those more complex accept headers I realized that also the logic in HTTPMediaTypePreference.swift for comparing two media types is incorrect. I fixed this along the way and also added a test for this

0xTim (approved):
Thanks!

PR #2609, opened 2021-04-26
Adds optional completion handler for `FileIO.streamFile`

trametheka:
This adds an optional completion handler to `FileIO`'s `streamFile(at:chunkSize:mediaType:onCompleted:)` which is called when the stream completes writing the response. (#2609)


0xTim (approved):
LGTM thanks!

0xTim:
@trametheka could you update your branch (for some reason GH won't let be auto update the branch) and then we're good to merge!

trametheka:
@0xTim try now

PR #2611, opened 2021-04-27
Fix warnings from Swift 5.4 compiler

kemchenj:
Fixes the following warnings now produced by the Swift 5.4 compiler:

```
warning: using 'class' keyword for protocol inheritance is deprecated; use 'AnyObject' instead
public protocol Extendable: class {
                            ^~~~~
                            AnyObject
```


0xTim (approved):
Thanks!

PR #2613, opened 2021-04-28
Add support for async/await

0xTim:
Adds support for async/await to Vapor. This initially adds async APIS for:

* `Client`
* `RoutesBuilder`
* `ResponseEncodable`

We'll add more to Vapor if they make sense - please suggest them in the `#development` channel on Discord.

This PR will be merged when async/await functionality lands in Swift and NIO and will be kept in the branch until then.


Sources/Development/routes.swift
```diff
@@ -223,6 +223,24 @@ public func routes(_ app: Application) throws {
...
         }
     }
+
+    let asyncRoutes = app.grouped("async").grouped(TestAsyncMiddleware())
+    asyncRoutes.get("client") { req async throws -> String in
+        let response = try await req.client.get("https://www.google.com")
+        guard let body = response.body else {
+            throw Abort(.internalServerError)
+        }
+        return String(decoding: body.readableBytesView, as: UTF8.self)
```
weissi, line 233:
```suggestion
        return String(buffer: body)
```

Sources/Development/routes.swift
```diff
@@ -223,6 +223,24 @@ public func routes(_ app: Application) throws {
...
+    }
+
+    func asyncRouteTester(_ req: Request) async throws -> String {
+        let response = try await req.client.get("https://www.google.com")
+        guard let body = response.body else {
+            throw Abort(.internalServerError)
+        }
+        return String(decoding: body.readableBytesView, as: UTF8.self)
```
weissi, line 241:
```suggestion
        return String(buffer: body)
```

lmcd:
Is it worth holding off on this for now? I and others I know are using Vapor (client and server features) on Apple platforms in which several years of OS support are required.

Vapor adopting async/Swift concurrency features will prevent us from using the latest Vapor releases for another year or two.

See: https://forums.swift.org/t/will-swift-concurrency-deploy-back-to-older-oss/49370/160

rleconte:
> Is it worth holding off on this for now? I and others I know are using Vapor (client and server features) on Apple platforms in which several years of OS support are required.
> 
> Vapor adopting async/Swift concurrency features will prevent us from using the latest Vapor releases for another year or two.
> 
> See: https://forums.swift.org/t/will-swift-concurrency-deploy-back-to-older-oss/49370/160

**As far as I understand**, this PR **adds** async/await support, but doesn't remove the current way of doing things (much like in Apple own APIs, an `async` *variant* of APIs is added on top of existing ones). So Vapor would make the async APIs available if your current swift version is adequate, and not bother you if it is not.  
But I may very well be completely wrong 😅 

The thing I'm not sure to understand is if this will be merged into Vapor 4 in an upcoming minor version, of be merged in an upcoming Vapor 5 ? 🤔 

With that being said, even if this comes as a breaking change in Vapor 5 and you need to have the new Swift version supporting concurrency to use it, I think it is fine, and people who want to deploy to older macOS versions can still use Vapor 4, and should not prevent people deploying on macOS 12, Linux (or in docker for that matter) to have these improvements as soon as possible rather than in a few years. Especially as Swift structured concurrency looks to be a pretty big improvement for server side swift.

0xTim:
@lmcd In short absolutely not.

First, these APIs are additional and gated behind availability checks. This won't stop you using Vapor on older OSes and Vapor 4 will work just as before.

Second, async/await is the most important change in server-side Swift since it was open sourced. Holding off on adopting would be the death of Vapor as people want to migrate to it and will choose other frameworks. We need to do what's best for the community as a whole and it's not feasible to hold back when the majority of users want and can use async/await

@rleconte the aim is to do this in Vapor 4

davdroman:
@0xTim do you mind merging `main` again please? :)

0xTim:
@davdroman Done!

Issue #2492, opened 2020-09-11
HEAD requests to routes that require query parameters can't be correctly responded to


owainhunt:
When a HEAD request is made to a route, it is transformed to a GET request by [DefaultResponder](https://github.com/vapor/vapor/blob/66f412bcb276c09ad3a5df263f159cee9fecc310/Sources/Vapor/Responder/DefaultResponder.swift#L78), and passed to the relevant route handler, if one exists. This is an issue if the route handler contains logic to validate query parameters. In that scenario, a HEAD request with no query parameters will fail, as the validation in the route handler will fail. 

### Steps to reproduce

Given the following route handler:
```
app.get("hello") { req -> Response in
    // validate query string contains a param
    guard let hello = try req.query.decode(Hello.self) else {
        throw Abort(.badRequest)
    }
    return .init(status: .ok)
}
```

Make a HEAD request to `/hello`

### Expected behavior

Should return `.ok`

### Actual behavior

Returns `.badRequest`

Note this behaviour also makes the following code non-functional for responding to HEAD requests, as the request will no longer be a HEAD by the time it gets to the route handler:
```
app.on(.HEAD, "hello") { req -> Response in
    .init(status: .ok)
}
````

I'd expect two things here: 
1. the HEAD request to the GET path should return .ok
2. if a HEAD route handler is explicitly defined, it should respond to a HEAD request

### Environment

* Vapor Framework version: 4.29.1
* OS version: Catalina 10.15.5, Ubuntu 18.04

PR #2614, opened 2021-04-28
Create separate HEAD routes instead of forwarding to GET route handler

owainhunt:
Instead of converting HEAD requests to GET requests, register a separate HEAD route for routes with all-constant paths that returns a 200 (#2492)


owainhunt:
@0xTim FYI, this is a recreation of https://github.com/vapor/vapor/pull/2496 that was caught up in a combination of the branch renaming script and me deleting the original branch.

0xTim (approved):
LGTM Thanks!

PR #2617, opened 2021-05-06
Fix warning when registering HEAD routes for GET requests

0xTim:
Fixes an issue where the `DefaultResponder` would add HEAD routes for all requests instead of just GET requests, resulting in `[Routing] Warning: Overriding route output at: HEAD/some/route` if you registered routes at the same path with different methods.


Andrewangeta (approved):
👍🏽

calebkleveter (approved):

PR #2619, opened 2021-05-07
Fix Array `beforeEncode`/`afterDecode` for `Content`

0xTim:
This fixes a bug where `beforeEncode` or `afterDecode` wouldn't be called on an array of `Content`.

However to make this work for value types, it requires replacing each element in the array which could be a significant slow down. Thoughts?

### DIFF
--- a/Sources/Vapor/Content/Content.swift
+++ b/Sources/Vapor/Content/Content.swift
@@ -123,6 +123,18 @@ extension Array: Content, ResponseEncodable, RequestDecodable where Element: Con
     public static var defaultContentType: HTTPMediaType {
         return .json
     }
+
+    public mutating func beforeEncode() throws {
+        for var element in self {
+            try element.beforeEncode()
+        }
+    }
+
+    public mutating func afterDecode() throws {
+        for var element in self {
+            try element.afterDecode()
+        }
+    }
 }
 extension Dictionary: Content, ResponseEncodable, RequestDecodable where Key == String, Value: Content {
     public static var defaultContentType: HTTPMediaType {
--- a/Sources/Vapor/Content/Content.swift
+++ b/Sources/Vapor/Content/Content.swift
@@ -125,14 +125,20 @@ extension Array: Content, ResponseEncodable, RequestDecodable where Element: Con
     }
 
     public mutating func beforeEncode() throws {
-        for var element in self {
+        for index in 0..<self.count {
+            var element = self[index]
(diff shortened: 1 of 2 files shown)


Sources/Vapor/Content/Content.swift
```diff
@@ -123,6 +123,24 @@ extension Array: Content, ResponseEncodable, RequestDecodable where Element: Con
...
+            self.insert(element, at: index)
+        }
+    }
+
+    public mutating func afterDecode() throws {
+        for index in 0..<self.count {
+            var element = self[index]
+            try element.afterDecode()
+            self.remove(at: index)
+            self.insert(element, at: index)
```
madsodgaard, line 141: Can't this be simplified to `try self[index].afterDecode()`? The tests still pass.

Same for encoding

0xTim: Good shout - so the question becomes is it worth iterating through every array when most people (I'm assuming) don't use this

calebkleveter (approved):

siemensikkema (approved):
@0xTim The performance problem you mentioned seems to be resolved now?

0xTim:
@siemensikkema The worst of it correct. However it still introduces a penalty as we'd not loop through every element in an array. For large arrays that could be quite a hit. Ideally we'd gate this behind a flag in my mind

siemensikkema:
@0xTim I guess the concern is mostly for cases where the hooks are not actually implemented and it's just iterating for no good reason. Would the compiler be able to optimize it out?
If not, without having a way to know if `Element` has an explicit implementation for these hooks we'd need some extra (static) variable on `Content` where we can control this behavior, or do you have something better in mind?
To complicate matters even more, one could say we should give `Dictionary<String, Content>` the same treatment if we're doing this.
Alternatively we leave it up the to developer to add the relevant extension for Arrays or Dictionaries where they need this behavior.

0xTim:
Yeah I've tried
```swift
extension Array where Element == SomeData {
    mutating func beforeEncode() throws {
        for index in 0..<self.count {
            try self[index].beforeEncode()
        }
    }
}

extension Array where Element: Content {
    mutating func beforeEncode() throws {
        for index in 0..<self.count {
            try self[index].beforeEncode()
        }
    }
}
```

Both don't get triggered

0xTim:
Closing as I don't think that the tradeoff is worth it

PR #2624, opened 2021-05-13
Fix building Vapor for tvOS and watchOS

Peter-Schorn:
Fixes building Vapor for tvOS and watchOS by setting the minimum platform versions to match Swift Crypto (#2624)


0xTim (approved):
Thanks!

0xTim:
Original comment:

PR #2625, opened 2021-05-19
Fix issue when using `:` in HTTP Basic Authentication passwords

raphaelcruzeiro:
This fixes a bug where using a password with `:` would not be able to authenticate using Vapor's Basic Authorisation. Using colons in a password was previously leading to an edge case as the authentication string was being split on any ":" making it impossible for users who registered with such passwords to ever authenticate. This allows for colons (`:`) to be valid characters in passwords (e.g. "This:is_myPassword123!").


0xTim (approved):
Thanks!

PR #2626, opened 2021-05-20
Add basic docstrings for `Application` and `Request`

kmahar:
This adds inline documentation for `Application` and `Request`. This helps when trying to integrate Jazzy (https://github.com/realm/jazzy/issues/936)

### DIFF
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -1,5 +1,6 @@
 import Backtrace
 
+/// Core type representing a Vapor application.
 public final class Application {
     public var environment: Environment
     public let eventLoopGroupProvider: EventLoopGroupProvider
--- a/Sources/Vapor/Request/Request.swift
+++ b/Sources/Vapor/Request/Request.swift
@@ -1,5 +1,6 @@
 import NIO
 
+/// Represents an HTTP request in an application.
 public final class Request: CustomStringConvertible {
     public let application: Application
 


0xTim (approved):
Thank you!

0xTim:
@kmahar I'm assuming this needs an actual release so you can pull in the updated dependency with the docs?

0xTim:
Original comment:
> I'm trying to include documentation for some extensions I've added to these types in API docs generated with Jazzy, and unfortunately due to https://github.com/realm/jazzy/issues/936 they won't show up unless if these types have their own docstrings to begin with.
>
> I can hack around it in my docs generation script, but it seemed like a better fix to at least provide some very basic docstrings on these types in Vapor itself. (The best fix would probably be fixing the Jazzy bug, but I don't know Ruby at all 🙂) 
>
> In this basic form I guess the docstrings don't provide a lot of value, but they seem like a fine starting point to provide more useful documentation in the future, and since both extending these types and using Jazzy are fairly common, I figured others might benefit from this change as well.

kmahar:
@0xTim Yeah, we would need a release. That said, I can work around it in the short term by manually editing the generated docs to add in the content, so it's not super urgent, just nice to have for the future 🙂

0xTim:
Cool I'll do a release. Yay for automation!

kmahar:
Awesome, thank you! 🎉

PR #2630, opened 2021-05-27
Fix issues with date parsing in HTTP headers

0xTim:
This fixes an issue introduced by #2603 where dates in HTTP headers would break the parsing logic. Any field that has a date would not get parsed correctly. This includes breaking the `set-cookie` header as HTTP header dates are in the format `Date: <day-name>, <day> <month> <year> <hour>:<minute>:<second> GMT`.

### DIFF
--- a/Sources/Vapor/HTTP/Headers/HTTPHeaders+Directive.swift
+++ b/Sources/Vapor/HTTP/Headers/HTTPHeaders+Directive.swift
@@ -25,13 +25,24 @@ extension HTTPHeaders {
     func parseDirectives(name: Name) -> [[Directive]] {
         let headers = self[name]
         var values: [[Directive]] = []
+        let separatorCharacters = getSeparatorCharacters(for: name)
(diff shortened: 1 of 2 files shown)


0xTim:
We likely need to go through the different headers as some can have valid `,` in and most can't. I've added a few exceptions for now but we should do a review and expand the tests to account for others. If anyone can point to an RFC spec that has the different ones that would be great!

siemensikkema:
@0xTim this seems to me like it would solve most of the issues occurring in the wild. A maximally compliant implementation would also take into account [quoted strings](https://datatracker.ietf.org/doc/html/rfc2822#section-3.2.5) in which commas are also allowed.

0xTim:
@siemensikkema yeah we'd need to flesh out a more formal, spec compliant parser but this fixes the immediate need. Want to give it the green tick?

siemensikkema (approved):

PR #2631, opened 2021-05-28
Fix repeated match of switch case

TheHarcker:
Fixed that a switch in HTTPHeaders+Directive.swift tried to match the .lastModified case twice.


0xTim (approved):
D'oh thanks!

PR #2633, opened 2021-05-31
Add support for tracing

0xTim:
This PR adds initial support for distributed tracing to Vapor. We'll be maintaining the fork and keeping it up to date as our tracing work continues with the eventual goal of merging any work back in to the upstream.

**Important:** there may be some breaking changes as this fork evolves so it shouldn't be considered API stable and use it at your own risk. The biggest change will come once Swift 5.5 is released and we switch all the `LoggingContext` stuff over to Task Local Values. We are however using this fork in production for tracing with our services so we consider it 'production ready'.

### DIFF
--- a/Sources/Development/routes.swift
+++ b/Sources/Development/routes.swift
@@ -1,4 +1,5 @@
 import Vapor
+import Baggage
 
 struct Creds: Content {
     var email: String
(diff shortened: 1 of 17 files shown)


0xTim:
@wolverian nope, we're waiting on a 1.0.0 release of https://github.com/apple/swift-distributed-tracing

Sources/Development/routes.swift
```diff
@@ -142,7 +143,7 @@ public func routes(_ app: Application) throws {
     }
 
     app.get("client") { req in
-        return req.client.get("http://httpbin.org/status/201").map { $0.description }
+        return req.client.get("http://httpbin.org/status/201", context: req).map { $0.description }
```
Joannis, line 146: Can't we pass in the context (baggage) in the `.client` property? If it's computed, we should already be able to. Otherwise, we can make it computed.

Sources/Vapor/Client/Client.swift
```diff
@@ -22,50 +24,51 @@ extension Client {
 }
 
 extension Client {
-    public func get(_ url: URI, headers: HTTPHeaders = [:], beforeSend: (inout ClientRequest) throws -> () = { _ in }) -> EventLoopFuture<ClientResponse> {
-        return self.send(.GET, headers: headers, to: url, beforeSend: beforeSend)
+    public func get(_ url: URI, headers: HTTPHeaders = [:], context: LoggingContext, beforeSend: (inout ClientRequest) throws -> () = { _ in }) -> EventLoopFuture<ClientResponse> {
+        return self.send(.GET, headers: headers, to: url, context: context, beforeSend: beforeSend)
```
Joannis, line 28: I'd prefer having this context available passively, so we don't need to consciously add support every time we want to use it.

0xTim:
I'm going to close this out. This was a POC before task locals were a thing and the final implementation will be significantly different and not force the passing of context everywhere

Sources/Vapor/Client/Client.swift
0xTim: Welcome to tracing in other languages 😆

Sources/Development/routes.swift
0xTim: Unfortunately not because if people make multiple requests in a request then they'd need different contexts to ensure the child spans are correct (and if they create their own spans). Since the client is cached on the request this would work. Isn't an issue with TLVs however

PR #2634, opened 2021-06-03
Make thread pool configurable and set default size to 64 threads.

madsodgaard:
Allows the default thread pool to be overriden and updates the default thread pool size to 64 threads. (#2634)

```swift
app.threadPool = NIOThreadPool(numberOfThreads: 100)
```

Fixes: #2308

### DIFF
--- a/Sources/Vapor/Core/Core.swift
+++ b/Sources/Vapor/Core/Core.swift
@@ -18,10 +18,16 @@ extension Application {
     /// ```
     ///
     /// If overriden, Vapor will take ownership of the thread pool and automatically start it and shut it down when needed.
-    /// - Warning: Should only be set during application setup/initialization.
+    ///
+    /// - Warning: Can only be set during application setup/initialization.
     public var threadPool: NIOThreadPool {
         get { self.core.storage.threadPool }
         set {
+            guard !self.isBooted else {
+                self.logger.warning("Cannot replace thread pool after application has booted.")
+                return
+            }
+            
             try! self.core.storage.threadPool.syncShutdownGracefully()
             self.core.storage.threadPool = newValue
             self.core.storage.threadPool.start()
--- a/Sources/Vapor/Core/Core.swift
+++ b/Sources/Vapor/Core/Core.swift
@@ -24,8 +24,7 @@ extension Application {
         get { self.core.storage.threadPool }
         set {
             guard !self.isBooted else {
-                self.logger.warning("Cannot replace thread pool after application has booted.")
-                return
+                fatalError("Cannot replace thread pool after application has booted")
             }
             
             try! self.core.storage.threadPool.syncShutdownGracefully()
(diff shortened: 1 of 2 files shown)


siemensikkema:
@madsodgaard do you think there is a lot of overhead when someone wants to use a different thread pool in terms of creating, shutting down, and creating again on each application startup? We could also have it both ways if we allow to pass a thread pool into the initializer as well as your current implementation.

madsodgaard:
@siemensikkema I don't know a lot about the computational cost of creating a new thread pool, but as this is a part of the startup process and should really only happen once per process, I don't think it would any where close to noticeable. Most people won't touch the `main.swift` file, which is why I chose this approach instead.

0xTim:
Overall +1 on this. The only thing is we might want to add some checks to make sure it's not changed after application start. I think the way you've configured it is the right way to do it as it fits in with the rest of Vapor. I'm not concerned by performance of initialising the thread pool as it's done at start up so shouldn't be a problem

madsodgaard:
@0xTim I added a check to ensure the application has not booted yet

Sources/Vapor/Core/Core.swift
```diff
@@ -9,9 +9,31 @@ extension Application {
...
+    ///
+    /// If overriden, Vapor will take ownership of the thread pool and automatically start it and shut it down when needed.
+    ///
+    /// - Warning: Can only be set during application setup/initialization.
     public var threadPool: NIOThreadPool {
-        self.core.storage.threadPool
+        get { self.core.storage.threadPool }
+        set {
+            guard !self.isBooted else {
+                self.logger.warning("Cannot replace thread pool after application has booted.")
```
0xTim, line 27: This should be a debug because of the Swift log guidelines. I'm actually tempted to just fatal error here

madsodgaard: Sure. I drew inspiration from here, but I agree that a fatal error would probably be better, to prevent hard-to-debug issues.

https://github.com/vapor/vapor/blob/73d57959a96ceca719558e70c320b2f729d2b25b/Sources/Vapor/HTTP/Client/Application%2BHTTP%2BClient.swift#L43

Sources/Vapor/Core/Core.swift
```diff
@@ -9,9 +9,30 @@ extension Application {
...
+    ///
+    /// If overriden, Vapor will take ownership of the thread pool and automatically start it and shut it down when needed.
+    ///
+    /// - Warning: Can only be set during application setup/initialization.
     public var threadPool: NIOThreadPool {
-        self.core.storage.threadPool
+        get { self.core.storage.threadPool }
+        set {
+            guard !self.isBooted else {
+                fatalError("Cannot replace thread pool after application has booted")
```
0xTim, line 27: Final nit - can you log a critical error here then we're good to go

0xTim (approved):

PR #2637, opened 2021-06-06
Add a built in `PlaintextDecoder`

### LINKED ISSUE #2607: Missing PlaintextDecoder means users need to fall back to req.body to parse plain text http apis
**Is your feature request related to a problem? Please describe.**
As per https://discordapp.com/channels/431917998102675485/435928766812913664/833695763296813107 http apis which return content with Content-Type plain/text hit an Abort.httpDecoder error: Unsupported Media Type when decoding.

e.g.

```
return req.client.get(address).flatMapThrowing { res in
    try res.content.decode(UInt8.self)
}.map { n in
    /// do stuff
}
```

I believe it would be in keeping with the rest of the API surface of vapor if this "just worked" (possibly with a requirement to specify the encoder)

I _think_ that means that I'd want a `PlaintextDecoder` (there's already a `PlaintextEncoder`) selected based on Content-Type plain/text

**Describe alternatives you've considered**
In my application I'm instead using `res.body!.getString(at: 0, length: res.body!.readableBytes)!` which feels strictly inferior (albeit hardly a lot of extra typing)

**Additional context**
None

dirtyhabits97:
Adds a new `PlaintextDecoder` to easily decode strings in requests and responses. `PlaintextDecoder` is the default decoder for `Content-Type: plain/text`. Resolves #2607

### DIFF
--- a/Tests/VaporTests/ContentTests.swift
+++ b/Tests/VaporTests/ContentTests.swift
@@ -369,9 +369,21 @@ final class ContentTests: XCTestCase {
             return res
         }
 
+        app.routes.get("empty-plaintext") { (req) -> Response in
+            let res = Response()
+            try res.content.encode("", as: .plainText)
+            return res
+        }
+
         try app.testable().test(.GET, "/plaintext") { res in
             XCTAssertEqual(res.status, .ok)
             XCTAssertEqual(try res.content.decode(UInt8.self), 255)
+            XCTAssertEqual(try res.content.decode(String.self), "255")
+        }
+
+        try app.testable().test(.GET, "/empty-plaintext") { res in
+            XCTAssertEqual(res.status, .ok)
+            XCTAssertEqual(try res.content.decode(String.self), "")
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Tests/VaporTests/ContentTests.swift
```diff
@@ -357,6 +357,23 @@ final class ContentTests: XCTestCase {
...
+
+        app.routes.get("plaintext") { (req) -> Response in
+            let res = Response()
+            try res.content.encode(data, as: .plainText)
+            return res
+        }
+
+        try app.testable().test(.GET, "/plaintext") { res in
+            XCTAssertEqual(res.status, .ok)
+            XCTAssertEqual(try res.content.decode(UInt8.self), 255)
```
0xTim, line 374: Could you add a test that decodes a string as well? Otherwise LGTM

0xTim (changes requested):

Tests/VaporTests/ContentTests.swift
dirtyhabits97: Done. Also added a test case for empty string as the response's body.

0xTim (approved):
Thanks! LGTM

0xTim:
Original comment:

> * Created PlaintextDecoder to easily decode raw strings
> * Registered PlaintextDecoder as the default decoder for Content-Type: plain/text
> This fixes #2607

PR #2647, opened 2021-07-01
Add ability to prepend middleware

jaredh159:
Allows middleware to be inserted at the start of the responder chain. This is useful when working with things like the `CORSMiddleware`.

Usage: 

```swift
app.middleware.use(corsMiddleware, at: .beginning)
```

The default behaviour remains appending:

```swift
app.middleware.use(someOtherMiddleware)
```

### DIFF
--- a/Sources/Vapor/Middleware/MiddlewareConfiguration.swift
+++ b/Sources/Vapor/Middleware/MiddlewareConfiguration.swift
@@ -4,6 +4,12 @@ public struct Middlewares {
     /// The configured middleware.
     private var storage: [Middleware]
 
+  
+    public enum Position {
+      case beginning
+      case end
+    }
+  
     /// Create a new, empty `MiddlewareConfig`.
     public init() {
         self.storage = []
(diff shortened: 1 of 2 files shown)


0xTim (changes requested):
Overall plus one on this - will need some tests however

jaredh159:
@0xTim updated with a test 👍

0xTim (approved):
LGTM

jaredh159:
@0xTim is there anything else I need to do to get this merged? Not sure what the workflow is... I assume you have merge privileges, right? Does someone else need to see this and approve as well? Thanks!

0xTim:
Original comment:

Because CORS middleware needs to go _before_ the error middleware (see #2295), that issue and the docs suggest that if you want to use CORS, you need to clear out the default middleware and then add it all back in piece by piece to get the desired order.

Since CORS seems like a really common requirement for many apps, and the current documented method seems like a bit of a workaround, it struck me that what might be nicer was to allow attaching middleware at the beginning of the stack, like so:

```swift
app.middleware.use(corsMiddleware, at: .beginning)
```

This PR implements that simple idea.  The default behavior will still be _appending_, so you can still do

```swift
app.middleware.use(someOtherMiddleware)
```

If the general idea seems acceptable, there might be a better name for this parameter.  I thought about `location: .prepend` or some other ideas too...

0xTim:
@jaredh159 nope nothing to do other than wait for my todo list to be whittled down 😅

PR #2652, opened 2021-07-13
Fix issue where Plaintext Decoder would crash

0xTim:
This resolves an issue where sending a request to a route handler that decodes a `Content` type with a plaintext header would cause a crash. 

This also fixes a warning by removing an extension on `MultipartError` as that's no longer used.


gwynne (approved):
Great catch! A beautiful example of a need to review `fatalError()` usage across the board, as well.

PR #2655, opened 2021-07-14
Remove Vapor 3 Shims

### LINKED ISSUE #2654: Remove _Vapor3  module
In apple/swift-nio#1897 we removed the NIO1 API shims. This is because recent versions of NIO no longer support Swift versions before 5.2, and NIO2 was released before 5.2. As this module is gone, Vapor will not build against recent NIO versions due to the missing dependency.

I recommend removing the module.

0xTim:
This removes the Vapor 3 shims. Vapor 3 is no longer supported and NIO has also removed the NIO 1 shims so this won't compile anyway.

Resolves #2654


gwynne (approved):
Go for it!

PR #2665, opened 2021-08-10
Improve documentation of ResponseCodable

Taenerys:
This PR further improves the documentation of Vapor's `ResponseCodable` to make it more convenient to see inline docs with `ResponseCodable` and the Vapor types that conform to it.

This PR includes:
- Added documentation comments to `ResponseCodable` protocol and its static function `decodeRequest`.

### DIFF
--- a/Sources/Vapor/Response/ResponseCodable.swift
+++ b/Sources/Vapor/Response/ResponseCodable.swift
@@ -10,7 +10,15 @@ public protocol ResponseEncodable {
     func encodeResponse(for request: Request) -> EventLoopFuture<Response>
 }
 
+/// Can convert `Request` to a `Self`.
+///
+/// Types that conform to this protocol can be returned in route closures.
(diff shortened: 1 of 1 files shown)


Sources/Vapor/Response/ResponseCodable.swift
```diff
@@ -10,7 +10,15 @@ public protocol ResponseEncodable {
     func encodeResponse(for request: Request) -> EventLoopFuture<Response>
 }
 
+/// Can convert `Request` to a `Self`.
+///
+/// Types that conform to this protocol can be returned in route closures.
```
0xTim, line 15: This is wrong - it should be 
```suggestion
/// Types that conform to this protocol can decode requests to their type.
```

0xTim (changes requested):
Small tech nit

0xTim (approved):
Thanks!

PR #2670, opened 2021-08-18
Fix `comparePreference` function when only one media type specified

### LINKED ISSUE #2668: `comparePreference` returns wrong results if only one media type is specified in the accept header
### Describe the bug

I want to write a API that offers different image formats. Therefore I have this in my code for testing:

```swift
switch req.headers.accept.comparePreference(for: .png, to: .gif) {
case .orderedAscending:
    print("prefer gif")
case .orderedDescending:
    print("prefer png")
case .orderedSame:
    print("same")
}
```

This works fine when the accept header contains both media types:
`image/png;q=0.5,image/gif;q=0.1` -> prefer png
`image/png;q=0.5,image/gif;q=0.8` -> prefer gif
`image/png;q=0.5,image/*;q=0.3` -> prefer png

The problem appears if the accept header misses one type:
`image/png;q=0.5` => prefer gif

I checked the `comparePreference` function, and think there is a issue with the cases when only one media type can be found in the header.

### To Reproduce

Register a new endpoint:
```swift
func test(req: Request) throws -> String {
    switch req.headers.accept.comparePreference(for: .png, to: .gif) {
    case .orderedAscending:
        return "prefer gif"
    case .orderedDescending:
        return "prefer png"
    case .orderedSame:
        return "same"
    }
}
```

Then send a request with the accept header `image/png;q=0.5` to the new endpoint. The response should be "prefer png" instead of "prefer gif".

### Expected behavior

The `comparePreference` function should prefer the accepted media type.

### Environment

* Vapor Framework version: 4.48.3
* OS version: macOS Big Sur Version 11.4

AZm87:
This fixes the `comparePreference` function which returns wrong results if only one media type is specified in the accept header.

Resolves #2668

### DIFF
--- a/Sources/Vapor/HTTP/Headers/HTTPMediaTypePreference.swift
+++ b/Sources/Vapor/HTTP/Headers/HTTPMediaTypePreference.swift
@@ -49,10 +49,10 @@ extension Array where Element == HTTPMediaTypePreference {
             }
         case (.none, .some):
             // there is not a value for a, no way it can be preferred
-            return .orderedDescending
+            return .orderedAscending
         case (.some, .none):
             // there is not a value for b, a is preferred by default
-            return .orderedAscending
+            return .orderedDescending
         case (.none, .none):
             // there is no value for either, neither is preferred
             return .orderedSame
--- a/Tests/VaporTests/HTTPHeaderTests.swift
+++ b/Tests/VaporTests/HTTPHeaderTests.swift
@@ -130,6 +130,15 @@ final class HTTPHeaderTests: XCTestCase {
             XCTAssertEqual(headers.accept.first(where: { $0.mediaType == .png })?.q, 0.8)
             XCTAssertTrue(headers.accept.comparePreference(for: .xml, to: .png) == .orderedDescending)
         }
+
+        // #2668 Preference should be consistent: present types are preferred to missing types
+        do {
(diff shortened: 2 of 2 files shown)


vzsg:
Could you add a test for this change, just to be future proof?

For example, this would fit well in `HTTPHeadersTests.testAcceptType()`:

```swift
// #2668 Preference should be consistent: present types are preferred to missing types
do {
    headers.replaceOrAdd(name: .accept, value: "image/png;q=0.5")
    XCTAssertEqual(headers.accept.comparePreference(for: .png, to: .gif), .orderedDescending)
    XCTAssertEqual(headers.accept.comparePreference(for: .gif, to: .png), .orderedAscending)
    XCTAssertEqual(headers.accept.comparePreference(for: .png, to: .png), .orderedSame)
    XCTAssertEqual(headers.accept.comparePreference(for: .gif, to: .gif), .orderedSame)
}
```

(Yeah, I was playing with this issue last night, but you beat me to the PR 🙂)

AZm87:
Done, I hope it's ok that I just copied your example. There are some failed tests, I'm not sure if that is related to this PR.

0xTim:
Tests are due to websocket issues, a PR to fix those is up

0xTim:
@vzsg happy to approve this?

0xTim (approved):

PR #2672, opened 2021-08-18
Use new concurrency verbage

hyerra:
This just updates the `async-await` branch to use the new name `completeWithTask` rather than the old `completeWithAsync`.


0xTim (approved):
Thanks!

Issue #2650, opened 2021-07-12
Multipart Errors are reported as internal server errors


0xTim:
When sending a request body that decodes a `Content` type. If the request is sent as multipart with the right header set, but the body is empty, a 500 error is returned. This can cause problems with apps that have automatic reporting for internal server errors as it should be reported as a bad request

JakeTiritilli:
I’ll work on trying to fix this one.

jareyesda:
@0xTim @BennyDeBock, Is this still an issue? Was looking into it but it seems like it was resolved.

The private `_ContentContainer` struct for a `Request` object has its `decode` method throw `Abort(.unprocessableEntity)` when `body` is `nil`. Line 76 inside `Request.swift`.

Here's a snipper of the `decode` method:

```
func decode<D>(_ decodable: D.Type, using decoder: ContentDecoder) throws -> D where D : Decodable {
   guard let body = self.request.body.data else {
       self.request.logger.debug("Request body is empty. If you're trying to stream the body, decoding streaming bodies not supported")
       throw Abort(.unprocessableEntity)
   }
   return try decoder.decode(D.self, from: body, headers: self.request.headers)
}
```

paunik:
@jareyesda  I have opened MR that covers that some while ago, now i have rebased and ran the tests, seems that issue is still present. If You can check, do I need to add some more things or we should close this MR.
Best

MahdiBM:
I came across this issue when i was looking at [one of the boards](https://github.com/orgs/vapor/projects/10).
I could not reproduce this issue. What the issue expects is if you run this [existing test](https://github.com/vapor/vapor/blob/cd91a66c7dd2d8658c9fc43ee9d96d7a259bbc5c/Tests/VaporTests/ContentTests.swift#L147), but with empty body (set to `nil` or `.init()`), it should throw a 5xx error. That's not the case and it will instead throw a un-processable entity error. Non-empty bad bodies also just throw a bad request error.

BennyDeBock:
@jareyesda as far as i'm aware, this hasn't been solved yet if it's still on the board. 

@MahdiBM maybe you could ask @0xTim for more details as to how to reproduce the issue.

MahdiBM:
@BennyDeBock I'll be happy to have a reproducible example on my hands but to me seems that what @jareyesda is correct. There is an explicit line in the decoder to throw a un-processable entity error in case the body is empty.
[...]

PR #2674, opened 2021-08-22
Resolve issue #2650

paunik:
In order to prevent 5xx internal error we try to:

- validate number of parameters in FormDataDecoder so we can detect when only boundary is present.


Sources/Vapor/Multipart/FormDataDecoder+Content.swift
```diff
@@ -5,6 +5,9 @@ extension FormDataDecoder: ContentDecoder {
         guard let boundary = headers.contentType?.parameters["boundary"] else {
             throw Abort(.unsupportedMediaType)
         }
+        if headers.contentType?.parameters.count ?? 1 == 1  {
```
0xTim, line 8: I'd prefer this to be a guard statement to match the rest of the code and better indicate an early exit. And we can chain the statements to avoid defaulting to 1

0xTim (changes requested):
Just need to switch the if statement to a guard and we'll be good to go

Sources/Vapor/Multipart/FormDataDecoder+Content.swift
paunik: Cool, seems that I had naive assumption that this will be ok, as test was failing when form data are present.
Now I changed if -> guard and moved check lower so we can check if we have anything else apart boundary. If only that in body then we throw.

0xTim (approved):
Thanks! @paunik if you could fix the conflicts this can be merged

gwynne (approved):

PR #2678, opened 2021-09-08
Adds default file support to FileMiddleware

NeedleInAJayStack:
Many HTTP servers support specifying a default file when a request hits a public directory. For example, a request to `http://server.com/path/` may automatically serve the file at `path/index.html`.

This PR adds this functionality to the FileMiddleware via an initialization parameter. It defaults to disabled but can be customized to whatever default file name is desired. This default file name is also served automatically for all subdirectory queries. If the default file does not exist, the request is passed on to the next Responder.

A usage example:

```swift
// Default file serving is disabled
app.middleware.use(FileMiddleware(publicDirectory: "public/"))

// Serves `index.html` (if it exists) to a request to `http://server.com/`, as well as any subdirectories
app.middleware.use(FileMiddleware(publicDirectory: "public/", defaultFile: "index.html"))
```

Resolves: https://github.com/vapor/vapor/issues/2455


NeedleInAJayStack:
Hey @0xTim or @gwynne, do you mind giving this a look?

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -30,14 +35,34 @@ public final class FileMiddleware: Middleware {
...
+            let res = request.fileio.streamFile(at: filePath)
+            return request.eventLoop.makeSucceededFuture(res)
+        } else {
+            // Check for the default file
+            guard let defaultFileName = defaultFile else {
+                return next.respond(to: request)
+            }
+            
+            var defaultFilePath = filePath.hasSuffix("/") ? filePath : filePath + "/"
+            defaultFilePath = defaultFilePath + defaultFileName
```
stevapple, line 57: Ideally we should explicitly handle the cases where `defaultFileName` starts with `/`.
In those cases, `defaultFilePath` should be `publicDirectory + defaultFileName`.

Tests/VaporTests/FileTests.swift
```diff
@@ -106,4 +106,29 @@ final class FileTests: XCTestCase {
...
+
+        let path = #file.split(separator: "/").dropLast().joined(separator: "/")
+        app.middleware.use(FileMiddleware(publicDirectory: "/" + path, defaultFile: "index.html"))
+
+        try app.test(.GET, "Utilities/") { res in
+            XCTAssertEqual(res.status, .ok)
+            XCTAssertEqual(res.body.string, "<h1>Default</h1>\n")
+        }
+    }
+
```
stevapple, line 122: And there should be a `testDefaultFileFromRoot`.

stevapple (changes requested):
Cases where `defaultFile` starts with `/` should be addressed.
We should either handle such cases explicitly or prohibit them with assertions.

Sources/Vapor/Middleware/FileMiddleware.swift
NeedleInAJayStack: Great suggestion! I've added this.

Tests/VaporTests/FileTests.swift
NeedleInAJayStack: Sure, I've added a testDefaultFileAbsolute method.

Sources/Vapor/Middleware/FileMiddleware.swift
stevapple: I think we should allow fallback to an absolute default file whenever the requested path doesn't exist? It doesn't rely on an existing directory as the root.

NeedleInAJayStack: > I think we should allow fallback to an absolute default file whenever the requested path doesn't exist?

I considered this and my only hesitation is that file existence is the only means for us to pass through FileMiddleware. If we make an absolute defaultFile and it matches even on paths that don't exist, we are basically saying that the server should ONLY ever serve the public files or the default file, regardless of any other routing definitions.

For example, consider the following server:

```swift
app.middleware.use(FileMiddleware(publicDirectory: "/Public", defaultFile: "/index.html"))
app.routes.get("hello") { req in
    return "hello there".encodeResponse(for: req)
}
```

We would never be able to reach the `hello` route because it would be caught in the FileMiddleware and `index.html` would be served.

Perhaps this is okay or there's a better way to do this, but I don't have enough experience with Vapor's Middleware system to know for sure. I'd be really interested in your opinion!

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -18,26 +24,67 @@ public final class FileMiddleware: Middleware {
...
+    /// Determines if input path is absolute based on a leading slash
+    func isAbsolute() -> Bool {
+        return self.hasPrefix("/")
+    }
+
+    /// Makes a path relative by removing all leading slashes
+    func removeLeadingSlashes() -> String {
+        var newPath = self
+        while newPath.hasPrefix("/") {
+            newPath = String(newPath.dropFirst())
```
stevapple, line 77:
```suggestion
            newPath.removeFirst()
```

stevapple (approved):

stevapple:
I’m okay with the change, cc @0xTim

0xTim (approved):

PR #2684, opened 2021-09-15
Fixes #2683 `HTTPHeaders+ContentDisposition` init

### LINKED ISSUE #2683: Multipart won't accept `File` while `Data` works
### Describe the bug

Multipart won't parse `File` if there are additional directives in `Content-Disposition` header.

Postman app adds `filename*` directive and it fails `File` parsing.

Screenshot 2021-09-15 at 23 08 10

### To Reproduce

1. Create simple POST request handler and use simple `Content` struct with one `File` field.

```swift
struct UpoadTestPayload: Content {
    let image: File
}
```

2. Use Postman app to send multipart request with one `image` file field

Screenshot 2021-09-15 at 23 18 20

3. See error `Could not convert value at "image" to type File from multipart part. for key image`

### Expected behavior

It should just parse uploaded File without any errors.

### Environment

* Vapor Framework version: 4.48.5
* OS version: macOS 11.1

MihaelIsaev:
Multipart won't parse `File` if there are additional directives in `Content-Disposition` header.

Postman app adds `filename*` directive and it fails `File` parsing.

Screenshot 2021-09-15 at 23 08 10

This fix allows to init `ContentDisposition` with additional directives in `Content-Disposition` header.

### DIFF
--- a/Sources/Vapor/HTTP/Headers/HTTPHeaders+ContentDisposition.swift
+++ b/Sources/Vapor/HTTP/Headers/HTTPHeaders+ContentDisposition.swift
@@ -54,7 +54,7 @@ extension HTTPHeaders {
                 case "filename":
                     self.filename = .init(parameter)
                 default:
-                    return nil
+                    continue
                 }
             }
         }


0xTim (commented):
LGTM though a test would be great. I'll leave the final review to @siemensikkema as he's been reworking all the multipart stuff recently

siemensikkema:
@MihaelIsaev This seems like a sensible and safe change to me. It doesn't impact any of the multipart-kit changes and vice-versa. Nevertheless, I agree with @0xTim that we should have a test for this change.
Besides that, could you rephrase the body of the PR a bit so that it reads more like release notes and describes the problem in more general terms not directly tied to multipart or postman? I'm thinking something along the lines of "Ignore unknown directives in `Content-Disposition` header."

PR #2685, opened 2021-09-15
Correctly pass `SocketAddress` through to `Request` convenience initialiser

davdroman:
Fixes an issue where the convenience initializer for `Request` has a `remoteAddress: SocketAddress?` parameter that wasn't being forwarded into the main initializer call (#2685)


0xTim (approved):
Good spot, thanks!

PR #2689, opened 2021-09-21
Handle null key correctly in validation when not required

0xTim:
Fixes an issue where a key was marked as not required but passing the value `null` in JSON would cause the validation to fail. 

E.g.

```json
{
  "number": null
}
```
would not pass

```swift
validations.add("number", as: Int.self, required: false)
```

### DIFF
--- a/Tests/VaporTests/ValidationTests.swift
+++ b/Tests/VaporTests/ValidationTests.swift
@@ -589,6 +589,12 @@ class ValidationTests: XCTestCase {
         }
         """
         XCTAssertNoThrow(try Site.validate(json: valid))
+
+        let valid2 = """
+        {
+        }
+        """
+        XCTAssertNoThrow(try Site.validate(json: valid2))
     }
 
     override class func setUp() {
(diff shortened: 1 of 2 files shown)


Tests/VaporTests/ValidationTests.swift
```diff
@@ -574,6 +574,92 @@ class ValidationTests: XCTestCase {
         })
     }
 
+    func testValidateNullWhenNotRequired() throws {
```
Mordil, line 577: Do we have a test case for when it is required and that JSON value is provided? I only ask since I'm not too familiar with this code area

Mordil (approved):

Tests/VaporTests/ValidationTests.swift
0xTim: Yes by default a validation is required so this behaviour is covered by the other tests

Mordil: Got it, thanks

gwynne (approved):

PR #2692, opened 2021-09-27
Expand Crypto dependencies

tonyarnold:
Expand the versions of Swift Crypto Vapor depends on to allow versions of Swift Crypto beyond 1.0.0.

⚠️ **WARNING:** this release may contain a breaking change if you're relying on Swift Crypto without specifying it as a dependency. If you're switching on `CryptoError` you'll need to account for the new case or specify Swift Crypto as a dependency and choose the version you want ⚠️


0xTim:
I'm in favour of this but it is a breaking change. There's a discussion here https://forums.swift.org/t/vapor-4-and-swift-crypto-2/52384

0xTim (approved):

PR #2695, opened 2021-10-01
Removing redundant call to beforeEncode() hook.

Strobocop:
When encoding content with the default content type, there was a redundant call to the `beforeEncode()` hook. This removes that unnecessary call.


0xTim:
@Strobocop I haven't forgotten this it's on my todo list. Need to look into why the tests are failing for a simple change

PR #2698, opened 2021-10-07
Allow URLEncodedForms to decode 'on' value als Bool => true

### LINKED ISSUE #2444: Content decode Bool type?
vapor: 4.25.0
xcode: 11.5

issue: when using request.content.decode(... function.

example.leaf
---
```
<form action="/example" method="POST">
    <div>
        <input type="checkbox" name="isActivated">
    </div>
    <div>
        <input type="submit" value="Submit">
   </div>
</form> 
```
example.swift
---
```
app.post("example") { request -> Response in
    struct Context: Content {
        var isActivated: Bool
    }
    let context = try request.content.decode(Context.self)
    return request.redirect(to: "/")
}
```

Result
---
- checkbox unchecked:  no issue
- checkbox checked: **{"error":true,"reason":"Value of type 'Bool' required for key 'isActivated'."}**

bobvoorneveld:
When a HTML form uses a checkbox to check something as a Bool, it sends the 'on' value if checked.
The URLEncodedForms decoder should be able to handle that.

```
<form method="post">
 <input type="checkbox" name="isActive" checked />
 </form>

struct FormData: Decodable {
    let isActive: Bool
}

let formData = try req.content.decode(FormData.self)
```

fixes #2444


0xTim:
@bobvoorneveld Thanks for this! Just a note to say it hasn't been forgotten, I just to spend some time to sit down and read the form spec etc to make sure we're all good. But wanted you to know this wasn't getting lost in a void!

bobvoorneveld:
NP, keep up the good work!

bobvoorneveld:
NP, keep up the good work!

bobvoorneveld:
Reading some documentation on input checkboxes via https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

```
If the value attribute was omitted, the default value for the checkbox is on, so the submitted data in that case would be subscribe=on.
```

I can solve this by adding a value to 1 in the input type. But would be nice to support this behaviour? Cannot find a direct reference to official HTML spec though.

Joannis (approved):
LGTM, spec looks correct

Joannis:
@0xTim or @gwynne any objection to merging this?

Joannis:
I also don't know if this is a semver-minor or semver-patch

PR #2725, opened 2021-11-01
Add a new test method to XCTApplicationTester returns a response directly

tonyarnold:
I'm finding that there are a few places where having my tests return an XCTHTTPResponse directly — rather than via a closure — feels more ergonomic in use (especially inline with async/await code). 

This PR proposes to add a method that does that.

I'm not sold on the name I've given this method (`performTest(…)`), so please feel free to suggest alternatives.


0xTim:
Why not `sendRequest`?

tonyarnold:
Thanks, Tim - I'll propose that in the PR. I'm not familiar with naming standards in Vapor, so I appreciate the suggestion!

PR #2728, opened 2021-11-03
Add missing public access to HTTPHeaders.Link initializer

gwynne:
Without a public initializer, the Link HTTP header support is effectively read-only, which was not intended.

A bit ridiculously, requires a semver-minor release due to adding a "new" public API.


jdmcd (approved):

0xTim:
Eh patch release as it was a bug 😜

gwynne:
Overriding spurious test failure to merge.

PR #2729, opened 2021-11-05
Xcode 13.2 doesn't like TSAN right now

gwynne:


gwynne:
This doesn't fix the issue, closing out.

PR #2730, opened 2021-11-06
Fix empty password in basic authorization parsing

Frizlab:
Currently if the password is empty in a Basic Authorization (the decoded base64 string ends with a semicolon), Vapor will fail to parse it.
It should succeed with an empty password.

This case happens often when dealing with OAuth2: client authentication should be done via Basic auth, and most OAuth clients do not have an app secret, which is represented with an empty password in Basic auth.


0xTim (approved):
Thanks!

PR #2731, opened 2021-11-08
Remove async content protocols pre-concurrency

siemensikkema:
This change attempts to correct a breaking change from Vapor 4.50.0 where changing the definition of `Content` broke existing code.

To illustrate:
```swift
struct S<T> {
  let t: T
}
extension S: Codable where T: Codable {}
extension S: Content, RequestDecodable, ResponseEncodable where T: Content {}
```
Started failing to compile with 4.50.0 because conformance to `AsyncRequestDecodable` and `AsyncResponseEncodable` could not be implied.

With this proposed change, the above problem is at least resolved when the condition `#if compiler(>=5.5) && canImport(_Concurrency)` is not met. For Swift 5.5+ with concurrency, the redefinition of `Content` is still a breaking change that will need to be dealt with.

### DIFF
--- a/Sources/Vapor/Concurrency/ResponseCodable+Concurrency.swift
+++ b/Sources/Vapor/Concurrency/ResponseCodable+Concurrency.swift
@@ -1,3 +1,4 @@
+#if compiler(>=5.5) && canImport(_Concurrency)
 import NIOCore
 
 /// Can convert `self` to a `Response`.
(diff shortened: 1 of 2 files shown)


gwynne (approved):
LGTM so far.

0xTim (approved):
LGTM - I don't know if we want to roll this into a release that supports concurrency on older OSes or that release would superscede this?

0xTim:
@siemensikkema do you still want this merged?

siemensikkema:
If it doesn't apply to 5.5+ then I don't see the point of merging it tbh.

0xTim:
Closing then, we can reopen in the future if needed

PR #2732, opened 2021-11-09
fixed Request Body description

MahdiBM:
This fixes an error in Request Body description.
Currently the description is set to return an empty string at all times.


0xTim (approved):

0xTim:
Thanks!

PR #2733, opened 2021-11-10
HTTP Authorization header scheme is case-insensitive

fizker:
Fixes an issue where Vapor would reject valid requests, if the Authorization token was not cased exactly like Vapor wanted it. The RFC dictates that the token should be case-insensitive.


0xTim:
For reference https://datatracker.ietf.org/doc/html/rfc7617#section-2

0xTim (approved):

PR #2737, opened 2021-11-14
Filled in missing documentation in commonly used APIs

Joannis:
I've added documentation for some of the most commonly used Vapor APIs that were still missing doc comments.

### DIFF
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -97,7 +97,7 @@ public final class Application {
     /// Starts the Application using the `start()` command, then waits for any running tasks to complete
     /// If your application is started without arguments, the default argument is used.
     ///
-    /// Under normal circumstances, `run `will start the shut down and wait for the web server to (manually) shut down before exiting the application.
+    /// Under normal circumstances, `run()` begin start the shutdown, then wait for the web server to (manually) shut down before returning.
     public func run() throws {
         do {
             try self.start()
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -94,6 +94,10 @@ public final class Application {
         DotEnvFile.load(for: environment, on: .shared(self.eventLoopGroup), fileio: self.fileio, logger: self.logger)
     }
     
+    /// Starts the Application using the `start()` command, then waits for any running tasks to complete
+    /// If your application is started without arguments, the default argument is used.
+    ///
+    /// Under normal circumstances, `run `will start the shut down and wait for the web server to (manually) shut down before exiting the application.
     public func run() throws {
         do {
             try self.start()
@@ -104,6 +108,11 @@ public final class Application {
         }
     }
     
+    /// When called, this will execute the startup command provided through an argument. If no startup command is provided, the default is used.
+    /// Under normal circumstances, this will start running Vapor's webserver.
+    ///
+    /// If you `start` Vapor through this method, you'll need to prevent your Swift Executable from closing yourself.
+    /// If you want to run your Application indefinitely, or until your code shuts the application down, use `run()` instead.
     public func start() throws {
         try self.boot()
         let command = self.commands.group()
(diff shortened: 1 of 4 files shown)


Sources/Vapor/Application.swift
```diff
@@ -94,6 +94,10 @@ public final class Application {
         DotEnvFile.load(for: environment, on: .shared(self.eventLoopGroup), fileio: self.fileio, logger: self.logger)
     }
     
+    /// Starts the Application using the `start()` command, then waits for any running tasks to complete
```
gwynne, line 97: I would say "method" rather than "command" here.

Sources/Vapor/Application.swift
```diff
@@ -94,6 +94,10 @@ public final class Application {
...
+    /// If your application is started without arguments, the default argument is used.
+    ///
+    /// Under normal circumstances, `run `will start the shut down and wait for the web server to (manually) shut down before exiting the application.
```
gwynne, line 100:
```suggestion
    /// Under normal circumstances, `run() `will begin the shutdown, then wait for the web server to (manually) shut down before returning.
```
(Yes, the use of both "shut down" and "shutdown" is on purpose 😅)

Sources/Vapor/Request/Request.swift
```diff
@@ -99,6 +99,8 @@ public final class Request: CustomStringConvertible {
         }
     }
 
+    /// This container is used to read your Decodable type using a ContentDecoder implementation.
+    /// If no ContentDecoder is provided, a supported Decoder will be selected based on the Request's Content-Type header.
```
gwynne, line 103:
```suggestion
    /// This container is used to read your `Decodable` type using a `ContentDecoder` implementation.
    /// If no `ContentDecoder` is provided, the `Request`'s `Content-Type` header is used to look up a decoder registered for that type.
```

Sources/Vapor/Request/Request.swift
```diff
@@ -142,12 +146,21 @@ public final class Request: CustomStringConvertible {
...
     }
 
+    /// The address from which this HTTP request was received by SwiftNIO.
+    /// This address may not represent the original address of the peer, especially if Vapor receives its requests through a reverse-proxy such as nginx.
     public let remoteAddress: SocketAddress?
     
+    /// The EventLoop upon which Vapor calls the middlewares and route handler for this request.
+    /// While respond to this request asynchronously, you **MUST** ensure the resulting EventLoopFuture is bound to this EventLoop.
+    ///
+    /// If you cannot ensure that a library returns on this EventLoop, you can use `eventLoopFuture.hop(to: request.eventLoop)`
```
gwynne, line 156:
```suggestion
    /// The `EventLoop` which is handling this request. The route handler and any relevant middleware are invoked on this event loop.
    /// 
    /// - Warning: A futures-based route handler **MUST** return an `EventLoopFuture` bound to this event loop.
    ///   If this is difficult or awkward to guarantee, use `EventLoopFuture.hop(to:)` to jump to this event loop.
```

gwynne (changes requested):
Just some grammatical and phrasing nits. I make no warranty that any of the suggestions are indisputably improvements in any given person's eyes, not even my own 😉.

Sources/Vapor/Client/Client.swift
```diff
@@ -14,7 +14,7 @@ extension Client {
 extension Client {
     public func get(_ url: URI, headers: HTTPHeaders = [:], beforeSend: (inout ClientRequest) throws -> () = { _ in }) -> EventLoopFuture<ClientResponse> {
         return self.send(.GET, headers: headers, to: url, beforeSend: beforeSend)
-    }
+    }`
```
0xTim, line 17: Random backtick 
```suggestion
    }
```

Sources/Vapor/Application.swift
0xTim: Agree, we don't have a start command so that could be seen as confusing

0xTim (changes requested):
Couple of minor nits then we're good

0xTim (approved):

PR #2741, opened 2021-11-22
Fix Backpressure

### LINKED ISSUE #2738: Backpressure not working
### Describe the bug

Memory is ballooning out of control when a client posts a huge upload and it can't be handled quick enough.

### To Reproduce

Simply run the example from the `Development` directory and observer:

```
// ( echo -e 'POST /slow-stream HTTP/1.1\r\nContent-Length: 1000000000\r\n\r\n'; dd if=/dev/zero; ) | nc localhost 8080
app.on(.POST, "slow-stream", body: .stream) { req -> EventLoopFuture<String> in
    let done = req.eventLoop.makePromise(of: String.self)

    var total = 0
    req.body.drain { result in
        let promise = req.eventLoop.makePromise(of: Void.self)

        switch result {
        case .buffer(let buffer):
            req.eventLoop.scheduleTask(in: .milliseconds(1000)) {
                total += buffer.readableBytes
                promise.succeed(())
            }
        case .error(let error):
            done.fail(error)
        case .end:
            promise.succeed(())
            done.succeed(total.description)
        }

        // manually return pre-completed future
        // this should balloon in memory
        // return req.eventLoop.makeSucceededFuture(())
        
        // return real future that indicates bytes were handled
        // this should use very little memory
        return promise.futureResult
    }

    return done.futureResult
}
```

### Expected behavior

Very little memory used. In reality, memory just instantly balloons.

### Environment

* Vapor Framework version:  4.48.4
* Vapor Toolbox version:
* OS version: macOS 11.4 (20F71)

### LINKED ISSUE #2739: Bottleneck in HTTPServerRequestDecoder?
Noticing that backpressure isn't working (https://github.com/vapor/vapor/issues/2738) made me also realise that NIO is taking in massive numbers of chunks super quick and efficiently, but they're arriving into Vapor at a much slower clip, even if you're not even doing anything with them.

Profiling a large upload, it looks like `HTTPServerRequestDecoder` is actually quite slow relative to how quick the rest of NIO is.
Probably around 30%+ of all time is spent dealing in `HTTPServerRequestDecoder`/`CircularBuffer`, half of that being ARC traffic.

With ARC, the whole NIO system is insanely efficient - over a 2s period, I'm seeing 10ms of retain traffic, and 9ms release.
Vapor on the other hand is seeing 109ms retain, and 92ms release.

I get a strong sense that there are some big optimisation opportunities here. I'm still trying to understand how this all works, but I might have a crack at hacking on this.

Someone correct me if I'm wrong and/or misinterpreting profiling data on this one.

lmcd:
Fixes #2738, fixes #2739.

The `read` method that implemented back pressure was in the wrong class and so would never get called. DoSing any streaming body request was trivial.

Before this fix, a 500MB upload in my application could consume 2GB of memory and take about 10-12 seconds (debug build).
The same request now consumes 50MB and takes ~6 seconds.


lmcd:
Also I'll add that this touches critical code paths in Vapor so should be thoroughly tested before merge.

lmcd:
Also, this doesn't *fully* address the issue I described with `CircularBuffer`, but it basically makes it a non-issue.

0xTim:
Thanks for this! Given the big changes, I'll rope a few people in to review and I'll need to set aside some dedicated time to go through it

lmcd:
Perfect!
Just to add some clarity, there is nothing new actually added here. I've just merged two classes into one.

There's one super minor change, that adds an `initialCapacity` of `4` to a `CircularBuffer`. This is because I've never seen this buffer growing beyond 4 with this change in place, so should reduce the impact of copies when the initial capacity was at it's default value of 16.

gwynne:
@lmcd Can you look into the crashing test, appears to occur only on Linux in the test `testEarlyExitStreamingRequest()`

lmcd:
@gwynne - Hmm I don't have a Linux test environment readily available here. From looking at the log, the assertions aren't failing, but the server isn't shutting down gracefully right?

Was this test succeeding before? Vapor has never shutdown gracefully for me on macOS - it's always been broken. See here: https://github.com/vapor/vapor/issues/2502

gwynne:
Yes, this was definitely working as expected before; it passes correctly without this PR and fails with it. I'm not particularly well-versed in NIO channel management, but offhand I'd guess that combining the handlers either changed the ordering of how certain operations are handled or increased the retained lifetime of something, preventing server shutdown with a streaming request in progress and thus causing the lifecycle failure.

gwynne:
And Vapor has always shut down gracefully just fine on macOS for me, except when run in Xcode with LLDB attached due to a bug in signals management.

lmcd:
Ah got it - I don't think I've ever attempted a manual shutdown without the Xcode debugger attached.
My current understanding of NIO pipelines, is that this wouldn't change the order of anything, and it should be behaving identically.

The issue is probably in the implementation of back pressure (initial commit for this here: https://github.com/vapor/vapor/pull/1979). This can be disabled by just fully commenting out the `read` method in this PR and reattempting the test suite on Linux.

Even then though, I think this might be a deeper NIO bug, considering this is working fine on some platforms but not others.

lmcd:
I don't have the resources/time to get a Linux environment going to investigate test suite issues. So I'm leaving this now and just moving forward with a Vapor fork.

I think this should be addressed with urgency though. Vapor is broken imo and should not be used in production without this issue being fixed.

VineFiner:
Any progress?

0xTim:
@VineFiner  this is fixed in [4.54.1](https://github.com/vapor/vapor/releases/tag/4.54.1)

PR #2761, opened 2022-01-16
Make `HTTPServer.init` and `Middleware.resolve` public

samisuteria:
Make `Middlewares.resolve()` public to allow users to create their own `Responders` (#2762)
Add a public initializer to `HTTPServer` to allow users to create heir own instance (#2761)

### DIFF
--- a/Sources/Vapor/HTTP/Server/HTTPServer.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServer.swift
@@ -231,7 +231,7 @@ public final class HTTPServer: Server {
 
     private var application: Application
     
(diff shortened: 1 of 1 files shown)


0xTim:
@samisuteria what's the use case for needing to create an `HTTPServer` yourself?

samisuteria:
I'm running both a `GRPCServer` and `HTTPServer` at the same time on different ports. I created my own server that implements `Vapor.Server` protocol and that class needs access to `HTTPServer.init`.

0xTim (approved):

PR #2762, opened 2022-01-16
Make Middlewares.resolve public

samisuteria:
Makes the resolve function public so users of Vapor can use it when making their own Responder.


samisuteria:
My use case: I'm making a custom `Responder` that behaves differently than `Vapor.DefaultResponder` and I want to be able to chain middleware to the route's responder. But in order to get access to that I need to make the resolve function public.

0xTim (approved):
Agree with opening this up for other responders

0xTim:
Will roll this into a release with #2761

PR #2765, opened 2022-01-19
Update ContentContainer.swift

linus-hologram:
fixed what was likely a copy/paste error in the documentation for the two get() methods of the content container.


0xTim (approved):
Thanks!

0xTim:
This will be rolled into the next release rather than cutting a release just for this

PR #2768, opened 2022-01-22
Fix request decompression (#2766)

### LINKED ISSUE #2766: Vapor 4.54.1 broke decompressing of certain gzipped requests
It seems that Vapor 4.54.1 broke decompressing of requests which are gzipped (not all of them, but some).
In my testing, with Vapor 4.54.0 decompressing works correctly for all gzip requests that I send to it. But in 4.54.1, I am running into this issue with certain gzip files:

```
[ ERROR ] /sync - Error while parsing the request: dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid JSON.", underlyingError: Optional(Foundation.JSONError.unexpectedEndOfFile))), error.localizedDescription = The operation could not be completed. The data isn’t in the correct format..
```

Here's how I am setting up support for decompressing requests in my configure.swift:
`app.http.server.configuration.requestDecompression = .enabled(limit: .size(20_000_000))`

(It has worked correctly for me for the last year+, up until yesterday when I updated my Vapor app on the server which pulled Vapor 4.54.1)

I'm attaching the gzip file for which it is failing:
[problem-file.gz](https://github.com/vapor/vapor/files/7916553/problem-file.gz).

Joannis:
Fix #2766.

Might need some extra changes to ensure decompression doesn't result in a larger-than-acceptable contiguous buffer living in-memory. The old implementations ensured that small payloads were held contiguously, and large payloads streamed. Since we can't assume the (suspected) end of a body is the end, I've made it so that the `awaitingEnd` case will accept more body data. Therefore, we need to prevent potential abuse in that case as well. I hope someone has a good solution here.

### DIFF
--- a/Sources/Vapor/HTTP/Server/HTTPServerRequestDecoder.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServerRequestDecoder.swift
@@ -55,10 +55,17 @@ final class HTTPServerRequestDecoder: ChannelDuplexHandler, RemovableChannelHand
                 self.requestState = .awaitingBody(request)
             default: assertionFailure("Unexpected state: \(self.requestState)")
             }
-        case .body(let buffer):
+        case .body(var buffer):
             switch self.requestState {
-            case .ready, .awaitingEnd: 
+            case .ready:
                 assertionFailure("Unexpected state: \(self.requestState)")
+            case .awaitingEnd(let request, var body):
+                // We cannot assume that a request's content-length represents the length of all of the body
+                // because when a request is g-zipped, content-length refers to the gzipped length.
+                // Therefore, we can receive data after our expected end-of-request
+                // When decompressing data, more bytes come out than came in, so content-length does not represent the maximum length
+                body.writeBuffer(&buffer)
+                self.requestState = .awaitingEnd(request, body)
             case .awaitingBody(let request):
                 if request.headers.first(name: .contentLength).flatMap(Int.init) ?? .max <= buffer.readableBytes {
                     self.requestState = .awaitingEnd(request, buffer)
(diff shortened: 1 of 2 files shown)


Sources/Vapor/HTTP/Server/HTTPServerRequestDecoder.swift
```diff
@@ -55,10 +55,17 @@ final class HTTPServerRequestDecoder: ChannelDuplexHandler, RemovableChannelHand
...
                 assertionFailure("Unexpected state: \(self.requestState)")
+            case .awaitingEnd(let request, var body):
+                // We cannot assume that a request's content-length represents the length of all of the body
+                // because when a request is g-zipped, content-length refers to the gzipped length.
+                // Therefore, we can receive data after our expected end-of-request
+                // When decompressing data, more bytes come out than came in, so content-length does not represent the maximum length
+                body.writeBuffer(&buffer)
+                self.requestState = .awaitingEnd(request, body)
             case .awaitingBody(let request):
                 if request.headers.first(name: .contentLength).flatMap(Int.init) ?? .max <= buffer.readableBytes {
```
gwynne, line 70: Let's correct this as well:
```suggestion
                if request.headers.first(name: .contentLength).flatMap(Int.init) == buffer.readableBytes {
```
I honestly do think ditching the string comparison it was doing before is a good thing, but if you'd rather revert entirely, I'm hardly of a mind to argue the point.

gwynne (commented):
I'm not sure whether it's better to do what you've done here or to just revert my mistaken change altogether. Best I can figure, it mostly depends on whether the original state logic or this version is more readable. Either way it should definitely be commented!

gwynne:
Figures, as soon as I submit my comments, another thought occurs 😆

I think reverting my change (as per the suggestion above) _and_ adding this logic would be the best way to go; it makes the state machine more resilient against a client sending a bogus `Content-Length` header.

gwynne (approved):

gwynne:
Whoops, forgot to tag the PR with a release label. Doing it manually.

gwynne:
These changes are now available in [4.54.2](https://github.com/vapor/vapor/releases/tag/4.54.2)

PR #2770, opened 2022-01-27
Add convenience methods for sending `Client` requests with a `Content` body

owainhunt:
Adds `post`, `put` and `patch` convenience functions for sending `Client` requests with a `Content` body.

E.g.

```swift
struct MyData: Content {
  let title: String
  let count: Int
}

let myData = MyData(...)

let response = try await req.client.post(url, myData)
```


0xTim (approved):
Thanks!

PR #2797, opened 2022-03-27
Fix ambiguous error with Swift nightlies

SusanDoggie:
Fix a compilation error in the latest Swift nightlies due to using `String.lazy.split(separator:)`, which is not needed


0xTim:
@SusanDoggie this appears to be breaking Vapor's JWT library

SusanDoggie:
> @SusanDoggie this appears to be breaking Vapor's JWT library

I run the test in local with the original lazy.split version, it seems that this fix not breaking it.

0xTim:
Original comment:

This line of code results in an ambiguous compilation error, and there is no point in using `lazy.split(separator:)`. It should be removed.

see: https://github.com/SusanDoggie/DoggieDB/runs/5708716271?check_suite_focus=true

PR #2804, opened 2022-04-14
Added custom failure descriptions to Validation Errors

AndreYonadam:
This adds support for custom failure descriptions when running validations. This allows you to customise the any unreadable keys and allows you to provide localized failure descriptions for each Validation.

### DIFF
--- a/Sources/Vapor/Validation/Validation.swift
+++ b/Sources/Vapor/Validation/Validation.swift
@@ -1,7 +1,7 @@
 public struct Validation {
     let run: (KeyedDecodingContainer<ValidationKey>) -> ValidationResult
 
-    init<T>(key: ValidationKey, required: Bool, validator: Validator<T>) {
+    init<T>(key: ValidationKey, required: Bool, validator: Validator<T>, customFailureDescription: String?) {
         self.init { container in
             let result: ValidatorResult
             do {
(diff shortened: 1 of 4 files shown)


Sources/Vapor/Validation/Validation.swift
```diff
@@ -84,6 +85,7 @@ public struct Validation {
 public struct ValidationResult {
     public let key: ValidationKey
     public let result: ValidatorResult
+    public let customFailureDescription: String?
```
0xTim, line 88: We're changing the public API here so will need a new `init()` that matches the existing behaviour to avoid any breaking changes

0xTim (changes requested):
Thanks for this, looks great! One change needed to avoid a breaking change then good to go

Sources/Vapor/Validation/Validation.swift
AndreYonadam: Done! Let me know if there are any other changes you'd like.

0xTim (approved):
LGTM!

0xTim:
Original comment:
This PR adds support for custom failure descriptions. This solves two major issues, unreadable keys, and the ability to provide localized failure descriptions for each Validation.

PR #2805, opened 2022-04-14
Provide async closures in XCTVapor

AndreYonadam:
This adds the ability to perform asynchronous work in the `beforeRequest` and `afterResponse` closures in XCTVapor. 

E.g.

```swift
try await app.test(.GET, "/hello", beforeRequest: { req async throws in
  let currentUsersCount = try await User.query(on: app.db).count()
}, afterResponse: { res async throws in
  let newUsersCount = try await User.query(on: app.db).count()
})
```

### DIFF
--- a/Sources/XCTVapor/XCTApplication.swift
+++ b/Sources/XCTVapor/XCTApplication.swift
@@ -103,6 +103,30 @@ public protocol XCTApplicationTester {
 }
 
 extension XCTApplicationTester {
+    #if compiler(>=5.5) && canImport(_Concurrency)
+    @discardableResult
+    public func test(
+        _ method: HTTPMethod,
+        _ path: String,
+        headers: HTTPHeaders = [:],
+        body: ByteBuffer? = nil,
+        file: StaticString = #file,
+        line: UInt = #line,
+        afterResponse: (XCTHTTPResponse) async throws -> ()
+    ) async throws -> XCTApplicationTester {
+        try await self.test(
+            method,
+            path,
+            headers: headers,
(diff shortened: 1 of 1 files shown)


0xTim (changes requested):
So whilst I agree these changes make sense in an async world this is going to be confusing to people porting code over, not to mention the compiler (did you have to change the existing tests to get it to compile?). So these functions should all return `XCTApplicationTester` like the non-async versions, unless you make an async version of `sendRequest`. The `beforeRequest` and `afterResponse` closures should just become async

AndreYonadam:
> So whilst I agree these changes make sense in an async world this is going to be confusing to people porting code over, not to mention the compiler (did you have to change the existing tests to get it to compile?). So these functions should all return `XCTApplicationTester` like the non-async versions, unless you make an async version of `sendRequest`. The `beforeRequest` and `afterResponse` closures should just become async

I didn't have to change the existing tests to get it to compile. I only did so to ensure that the new code was being tested/used somewhere. I am fine with either providing an async version of `sendRequest` or simply just changing the `beforeRequest` and `afterResponse` closures should just become async instead of this change. I feel like you might prefer the latter?

0xTim:
So I think we need both essentially. There are 3 helper functions in XCTVapor including a `sendRequest` - https://github.com/vapor/vapor/blob/main/Sources/XCTVapor/XCTApplication.swift#L157

So essentially just 3 additional functions that match these but with closures that can run async code - make sense?

0xTim (approved):
Thanks!

PR #2808, opened 2022-04-18
Do not conform UUID to LosslessStringConvertible

stephencelis:
Vapor should not extend a type it does not own with a protocol it does not own. Doing so can lead to conflicts with other code.


0xTim:
I agree that this bit of code should never have been added and Swift makes it difficult to add these enhancements. However removing this would cause breaking changes, both in code compilation and behaviour across Vapor so we can't merge this. At least not for Vapor 4. I'll close this but we can ensure that it's fixed for Vapor 5

PR #2820, opened 2022-05-07
Allow perform test when another testable application is in Live.

AFutureD:
When we build some microservices in Vapor, one common thing is that we request other services. 

For example, we have a URL provided by an authentication server to verify a Request and return the user information.

Now, the `performTest` method in XCTVapor, when XCTApplicationTester is `Live`, will start the app and that shutdown right away. So, this is not allowed that a testable application in Live exists when another app is testing.

To make two applications exists at same time, I provide new method.

For demo, see `testLiveApplication` in  [ApplicationTests.swift](Tests/VaporTests/ApplicationTests.swift).

### DIFF
--- a/Sources/XCTVapor/XCTApplication.swift
+++ b/Sources/XCTVapor/XCTApplication.swift
@@ -1,6 +1,10 @@
 extension Application: XCTApplicationTester {
     public func performTest(request: XCTHTTPRequest) throws -> XCTHTTPResponse {
-         try self.testable().performTest(request: request)
+        try self.testable().performTest(request: request)
+    }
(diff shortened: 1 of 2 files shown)


0xTim:
I'm not sure I 100% see the benefit in adding this. Is your intended usage to spin up `Service1Application` and `Service2Application` as complete applications to be able to send requests from one to the other in a test? Is one of these just a test service or are they both the real service?

AFutureD:
One is the real service and one is the test service.
But the real service may not available when testing, so we have to create a temperate one to simulate it.

PR #2821, opened 2022-05-11
Adds `cache.delete(_:)`

### LINKED ISSUE #2659: Removing value from cache requires explicit generic type
To reset a value in the cache you should be able to do:
```swift
cache.set("foo", to: nil)
```
However, due to the generic constraint that code would fail with: `Generic parameter 'T' could not be inferred`. You must do something like:
```swift
cache.set("foo", to: String?.none)
```

BennyDeBock:
closes #2659 

Adds a function called `delete` to remove a cached value. I wasn't able to override the method with `ExpressableByNilLiteral?` instead of `T?` without making breaking changes

### DIFF
--- a/Sources/Vapor/Cache/Cache.swift
+++ b/Sources/Vapor/Cache/Cache.swift
@@ -9,12 +9,12 @@ public protocol Cache {
     func set<T>(_ key: String, to value: T?) -> EventLoopFuture<Void>
         where T: Encodable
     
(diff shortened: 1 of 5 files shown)


0xTim (approved):
LGTM

0xTim:
Original comment:

Adds a function called `delete` to remove a cached value. I wasn't able to override the method with `ExpressableByNilLiteral?` instead of `T?` without making breaking changes

PR #2822, opened 2022-05-11
Support for `peerAddress` on `Request`

### LINKED ISSUE #2736: Support for peerAddress on Request
When checking a request's `remoteAddress`, it's usually the address from which NIO received the request. While that's technically correct, Vapor servers are often hosted behind a reverse-proxy such as nginx. Therefore the original peer's address is often lost. To resolve this, it's not uncommon to set the original peer's address in a header such as `X-Forwarded-For`.

I'd like Vapor to add a property on Request, such as `peerAddress`, which reads the headers to find the peer's address first.

paunik:
Adds a new computed property `peerAddress` on `Request` that parses headers commonly used by proxies and load balancers to pass through originating client address

Resolves #2736


Sources/Vapor/Request/Request.swift
```diff
@@ -32,6 +33,26 @@ public final class Request: CustomStringConvertible {
...
 
+    /// We try to determine true peer address if load balacer or reversed proxy provided info in headers
+    ///
+    /// Priority of getting value from headers is as following:
+    ///
+    /// 1. try the "Forwarded" header (e.g. for=192.0.2.60; proto=http; by=203.0.113.43)
+    /// 2. try the "X-Forwarded-For" header (e.g. client_IP, proxy1_IP, proxy2_IP)
+    /// 3. fallback to the socket's remote address provided by SwiftNIO ( e.g. 192.0.2.60:62934)
+    ///
+    public var peerAddress: String? {
```
Joannis, line 44: I like your docs! But it also made me realise that the different formats for these headers should be mixed carefully. I think it's wiser to expose `peerAddress` as a `SocketAddress`, and try to parse that address from these headers. Possibly providing a fallback for the port, and if there are multiple IP addresses, dealing with that in a later PR or returning a SocketAddress based on the client_IP from your examples.

Joannis:
@0xTim can you read my above comment?

Sources/Vapor/Request/Request.swift
paunik: Hey, tnx for the reply, while I was working i had exact taught, that maybe should be "unified" how this address is returned, rather then just some string :)

From my side makes 100% sense to have this as `SocketAddress`

paunik: @Joannis If I want to use `SocketAddress` as return, and from forwarded header and from xForwardedFor I can only get the address and not the port, what would be ok value to use for `SocketAddress(address: String, port: Int)` init method?

0xTim: I agree `SocketAddress` makes sense, we want to try and avoid `String` where possible. For the port a sensible fallback I guess is port 80

paunik: 👍

Joannis: Sorry that I missed your reply last week @paunik . I was going to suggest exactly what Tim did

paunik: Done, only doubt is about this test: `testRequestPeerAddressRemoteAddres` where i dont check that the port itself is different then 80, but not sure if we really need that test?

Joannis (approved):

Joannis:
Thanks for the PR, and your work to improve upon the feedback! LGTM!

Joannis:
I'll await @0xTim 's feedback as well, so he can merge it.

0xTim:
Original comment

MR for https://github.com/vapor/vapor/issues/2736

- Added computed property peerAddress on Request class that parses headers commonly used by proxies and load balancers to pass through originating client address
- Added tests for peerAddress computed property

Sources/Vapor/Request/Request.swift
```diff
@@ -32,6 +33,26 @@ public final class Request: CustomStringConvertible {
...
+    ///
+    /// Priority of getting value from headers is as following:
+    ///
+    /// 1. try the "Forwarded" header (e.g. for=192.0.2.60; proto=http; by=203.0.113.43)
+    /// 2. try the "X-Forwarded-For" header (e.g. client_IP, proxy1_IP, proxy2_IP)
+    /// 3. fallback to the socket's remote address provided by SwiftNIO ( e.g. 192.0.2.60:62934)
+    /// in 1. and 2. will use port 80 as default port, and  3. will have port number provided by NIO if any
+    public var peerAddress: SocketAddress? {
+        if let clientAddress = headers.forwarded.first?.for {
+            return try! SocketAddress.init(ipAddress: clientAddress, port: 80)
```
0xTim, line 46: We should not be force trying these. Since `peerAddress` is optional, make these `try?`

Sources/Vapor/Request/Request.swift
```diff
@@ -32,6 +33,26 @@ public final class Request: CustomStringConvertible {
...
+        }
+
+        if let xForwardedFor = headers.first(name: .xForwardedFor) {
+            return try! SocketAddress.init(ipAddress: xForwardedFor, port: 80)
```
0xTim, line 50: As above

0xTim (changes requested):
We can't force try anywhere in the framework, we should change this

Sources/Vapor/Request/Request.swift
paunik: Done, tnx.

Sources/Vapor/Request/Request.swift
paunik: Done

Sources/Vapor/Request/Request.swift
```diff
@@ -32,6 +33,26 @@ public final class Request: CustomStringConvertible {
...
+    ///
+    /// Priority of getting value from headers is as following:
+    ///
+    /// 1. try the "Forwarded" header (e.g. for=192.0.2.60; proto=http; by=203.0.113.43)
+    /// 2. try the "X-Forwarded-For" header (e.g. client_IP, proxy1_IP, proxy2_IP)
+    /// 3. fallback to the socket's remote address provided by SwiftNIO ( e.g. 192.0.2.60:62934)
+    /// in 1. and 2. will use port 80 as default port, and  3. will have port number provided by NIO if any
+    public var peerAddress: SocketAddress? {
+        if let clientAddress = headers.forwarded.first?.for {
+            return try? SocketAddress.init(ipAddress: clientAddress, port: 80)
```
0xTim, line 46: Actually, we don't need to provide a port here since it's optional so we can probably just leave it out. That should result with the peer address just being an IP address right?

paunik: Thing is that SocketAddress init has port as Int, and not Int? so it cant be omitted. Guess we are stuck with default port, and in this case 80 as suggested.

Joannis: I think we can merge, right @0xTim ?

0xTim: I was looking at the `SocketAddress` in Soto 🤦 

Yes we're good to merge, will bundle it in with a few other PRs

0xTim (approved):

Issue #2488, opened 2020-09-02
DefaultResponder metrics calls cause excessive CPU/RAM usage increases


owainhunt:
If your app has a large number of routes, or a single route that accepts a variable (`/products/:productId`, for example), the standard metric gathering in `DefaultResponder` can lead to dramatic increases in CPU and RAM usage. Because the metrics for `http_requests_total` and `http_request_duration_seconds` are based on path, a new counter and timer will be created for each individual path requested from the app, which over time leads to CPU/RAM usage and the size of the metrics payload increasing, to the point that in a Docker environment your app will get killed.

### Steps to reproduce

- Create a new app with a single route that accepts a variable as part of its path.
- Make a large number of requests to the app with a different value for the variable each time. 
- Observe CPU usage increasing in Xcode
- Optional: add a metrics package (I'm using Prometheus) and a metrics route, and make a request to a metrics path as below. You'll see the CPU spike dramatically when the path is called.
```
app.get("metrics") { req -> EventLoopFuture<String> in

        guard let prometheus = try? MetricsSystem.prometheus() else {
            return req.eventLoop.makeFailedFuture(Abort(.internalServerError))
        }
        
        let promise = req.eventLoop.makePromise(of: String.self)
        prometheus.collect(into: promise)
        return promise.futureResult
    }
```

### Expected behavior

CPU usage shouldn't dramatically increase.

### Actual behavior

For each request that comes in that doesn't match a previous request, a new Counter and Timer are created. This adds to the values stored in memory, and also increases the payload size for the metrics endpoint (ours hits 1MB within a few minutes and continues to increase). CPU and RAM usage increase, to the point that the app is killed by (in this case) Kubernetes, and the whole cycle starts again.

### Environment

* Vapor Framework version: 4.29.1
* OS version: Observed on Catalina and Ubuntu 18.04

I'm not sure what the sensible way to handle this is. We've gone the route of adding a custom `Responder` that's identical to `DefaultResponder`, but with the call to `updateMetrics()` removed. Would it be sensible to have these metrics be opt-in (or at least make it possible to disable them) on `DefaultResponder`? Or should the Counter/Timer use something other than the full path as a label, to avoid the issue altogether?

PR #2827, opened 2022-05-16
Add flag to enable/disable metrics

valerianb:
Adds a flag that defaults to true to enable / disable reporting metrics via the default handler.

Relates to:
https://github.com/vapor/vapor/issues/2171
https://github.com/vapor/vapor/issues/2488


0xTim (approved):
Thanks!

PR #2829, opened 2022-05-17
Add international email validator

### LINKED ISSUE #2815: Email validator doesn't support ß
Thank you for a great framework!

### Describe the bug

Disclaimer: this is a very regional issue...

The built in email validator doesn't support ß in the email name nor the email domain. This has been allowed in German emails since 2010 and hence can cause issues for German users.
[Reference to (German) article describing this issue](https://www.derwesten.de/leben/digital/e-mail-adressen-jetzt-mit-ss-moeglich-id3951009.html)

### To Reproduce

Create a request:
```swift
struct CheckEmailRequest: Codable, Validatable {
    let email: String

        validations.add("email", as: String.self, is: .email)
    }
}
```

In controller:
```swift
func checkEmailAvailability(request: Request) throws -> HTTPResponseStatus {
    try CheckEmailRequest.validate(query: request)
    let email = try request.query.decode(CheckEmailRequest.self).email
    return .noContent
}
```

Write a test file with the following tests:
```swift
// MARK: Test valid email with ß

let emailWithSpecialCharacterInName = "ß@b.com"
try app.test(
    .GET,
    "\(checkEmailPath)?email=\(emailWithSpecialCharacterInName)"
) { response in
    XCTAssertEqual(
        response.status,
        .noContent,
        "Expected `noContent` to be returned since the ß is allowed in an email name"
    )
}

let emailWithSpecialCharacterInDomain = "a@ß.com"
try app.test(
    .GET,
    "\(checkEmailPath)?email=\(emailWithSpecialCharacterInDomain)"
) { response in
    XCTAssertEqual(
        response.status,
        .noContent,
        "Expected `noContent` to be returned since the ß is allowed in an email domain"
    )
}
```
Both tests will return a validation error at the moment

### Expected behavior

I would the expect the built in email validator to not return an error for valid emails with ß in either the name or domain part.

### Environment

* Vapor Framework version: Vapor 4.55 + newest version
* Vapor Toolbox version: 18.3.6
* OS version: MacOS Monterey 12.3.1 (also on Linux server)

BennyDeBock:
Since it's hard to get a regex that works for both unicode character emails and utf8 emails, add a extra validator for international emails.
Adds an international email validator conforming to [this regex](http://jsfiddle.net/aossikine/qCLVH/3/) that works for unicode characters.

Also adds tests.
Increase the max characters of an email to 255

closes #2815

### DIFF
--- a/Sources/Vapor/Validation/Validators/Email.swift
+++ b/Sources/Vapor/Validation/Validators/Email.swift
@@ -6,7 +6,22 @@ extension Validator where T == String {
                 let range = $0.range(of: regex, options: [.regularExpression]),
                 range.lowerBound == $0.startIndex && range.upperBound == $0.endIndex,
                 // FIXME: these numbers are incorrect and too restrictive
-                $0.count <= 80, // total length
+                $0.count <= 255, // total length
+                $0.split(separator: "@")[0].count <= 64 // length before `@`
+            else {
+                return ValidatorResults.Email(isValidEmail: false)
+            }
+            return ValidatorResults.Email(isValidEmail: true)
+        }
+    }
+    
+    public static var internationalEmail: Validator<T> {
+        .init {
+            guard
+                let range = $0.range(of: regexInt, options: [.regularExpression]),
+                range.lowerBound == $0.startIndex && range.upperBound == $0.endIndex,
+                // FIXME: these numbers are incorrect and too restrictive
+                $0.count <= 255, // total length
                 $0.split(separator: "@")[0].count <= 64 // length before `@`
             else {
                 return ValidatorResults.Email(isValidEmail: false)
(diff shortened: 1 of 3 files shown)


Sources/Vapor/Validation/Validators/Email.swift
```diff
@@ -6,7 +6,22 @@ extension Validator where T == String {
                 let range = $0.range(of: regex, options: [.regularExpression]),
                 range.lowerBound == $0.startIndex && range.upperBound == $0.endIndex,
                 // FIXME: these numbers are incorrect and too restrictive
```
0xTim, line 8: If the numbers are now correct, we can remove this. Would be good to link to https://emailregex.com/email-validation-summary/ as the reason why we have these numbers

Sources/Vapor/Validation/Validators/Email.swift
```diff
@@ -48,3 +63,7 @@ z0-9-]*[a-z0-9])?\\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\\[(?:(?:25[0-5\
 9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21\
 -\\x5a\\x53-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])+)\\])
 """
+
+private let regexInt: String = """
```
0xTim, line 67: Rename for clarity

```suggestion
private let regexInternationEmail: String = """
```

0xTim (changes requested):
Couple of small nits

0xTim:
@heidipuk does this work for you?

heidipuk:
@0xTim yes, that should work for our use case 😄

0xTim (approved):

Issue #2526, opened 2020-11-07
Validation .nil is not working for Float? type.


seriyvolk83:
It works for `String?.self`, but not for `Float?.self`

```
validations.add("weight", as: Float?.self, is: .nil || .range(0...), required: false)
```

```
$ curl -X GET 'http://localhost:8080/hello?name=123&date=2020-01-20&age=13&weight=null'

{"error":true,"reason":"weight is not a(n) Float"}
```

0xTim:
@seriyvolk83 this doesn't work for `String?.self` either, the only reason it works for your case is because `&weight=null` satisfies `.range(0...)`. You shouldn't pass the `weight` through in the query if it's null

(Sorry it's taken so long to look at this!)

0xTim:
I'm going to close this for now. Feel free to reopen if you still think it's an issue!

PR #2830, opened 2022-05-18
Fix issue where validation on nil with empty parameter fails

BennyDeBock:
Fixes issue #2526


BennyDeBock:
When setting up the validators, it sees the nil validation, but it never goes into the init to do the validation

Putting a breakpoint here stops the program
https://github.com/vapor/vapor/blob/main/Sources/Vapor/Validation/Validators/Nil.swift#L4

Putting a breakpoint a line lower, into the init doesn't cause the program to stop at the breakpoint meaning the argument inside the init function does not get triggered

PR #2832, opened 2022-05-21
Clean up use of C targets in Vapor

gwynne:
This PR:

- Removes `COperatingSystem` entirely (unneeded)
- Replaces `CBase32` with a pure-Swift implementation of Base32 and Base64 (BaseN, really).
- Manually vendors `CURLParser` as `CVaporURLParser`.
- Manually vendors `CBcrypt` as `CVaporBcrypt`.

The aim of these changes is to reduce the dependency on additional unnecessary C-based targets and to avoid naming conflicts with other packages which also embed, for example, Bcrypt.


0xTim (approved):
LGTM - I don't see any tests for BaseN though? Also have we verified any performance changes in the new base64 code? I'm assuming it's been lifted from the swift extras org but there's no attribution

0xTim:
Also this should be a semver-patch right because there are no new public APIs?

gwynne:
> LGTM - I don't see any tests for BaseN though? Also have we verified any performance changes in the new base64 code? I'm assuming it's been lifted from the swift extras org but there's no attribution

Nope, entirely hand-written by me. I haven't run any performance tests yet, planning to add those (and more unit tests) next.

gwynne:
> Also this should be a semver-patch right because there are no new public APIs?

I'd rather make it semver-minor with so many changes to the package setup - some of the changes are publicly visible, such as the renamed modules; I started this PR because of a name collision with the `CBcrypt` target in another package, even though neither package provides the target as a product (see https://github.com/orlandos-nl/Citadel/pull/11)

Joannis (approved):

gwynne:
Marking as draft to avoid this accidentally getting merged as is; need to finish fixing performance regressions.

gwynne:
The Base32 and Base64 implementations now included match or (usually) improve upon the C-based implementation (for Base32) or Foundation's (for Base64), on both macOS and Linux.

0xTim (approved):
LGTM

0xTim:
@gwynne hold off merging for now, we have several server-minor PRs in flight so I'll roll them in to one

PR #2833, opened 2022-05-23
Add logs for failing parameters `require()`

### LINKED ISSUE #2720: Improve debugging for errors thrown when missing required Parameters
We had a typo in our code that led to this error: https://github.com/vapor/vapor/blob/main/Sources/Vapor/Routing/Parameters%2BRequire.swift#L23. It was hard to debug in part because of the vague `Abort(.internalServerError)`. I suggest adding a `reason` value to the `Abort`, unless that would be considered a breaking change. I can submit a PR if you're interested.

BennyDeBock:
Adds a reason to the `Abort` statement as well as adding a log statement for debugging

closes #2720


0xTim (changes requested):
One minor change

Sources/Vapor/Routing/Parameters+Require.swift
```diff
@@ -20,11 +20,13 @@ extension Parameters {
         where T: LosslessStringConvertible
     {
         guard let stringValue: String = get(name) else {
-            throw Abort(.internalServerError)
+            self.logger.debug("The parameter provided does not exist")
```
0xTim, line 23: Let's log the name here
```suggestion
            self.logger.debug("The parameter \(name) does not exist")
```

Sources/Vapor/Routing/Parameters+Require.swift
```diff
@@ -20,11 +20,13 @@ extension Parameters {
...
+            throw Abort(.internalServerError, reason: "The parameter provided does not exist")
         }
 
         guard let value = T.init(stringValue) else {
-            throw Abort(.unprocessableEntity)
+            self.logger.debug("The parameter value could not be converted to the required type")
```
0xTim, line 28:
```suggestion
            self.logger.debug("The parameter \(stringValue) could not be converted to \(T.Type)")
```

(You'll need to check that locally, I think T.Type should compile but worth checking)

0xTim (changes requested):

0xTim (approved):

PR #2840, opened 2022-06-01
Fix issue with tail content-range headers

### LINKED ISSUE #2836: `Content-Range` with tail sent is handled wrongly on media stream
### Describe the bug

When doing a request to a route that stream media like a video or file, when doing a `curl -sv -r -700 http://localhost:8080/hello` it returns a response with `content-range: 700-.../...`

This means that the http request wasn't handled properly as it returned more than the last 700 bytes of the file.

### To Reproduce

1. To begin with, clone this repository for testing purposes:
 `https://github.com/BennyDeBock/videoTest.git`
2. Fill in the path to the files in routes.swift
3. Run the application
4. Run `curl -sv -r -700 http://localhost:8080/hello` in a bash cli

### Expected behavior

The returning request gives the last 700 bytes of the file back

### Environment

*Vapor Framework version: 4.60.1
*Vapor Toolbox version: 18.3.6
*OS version: 12.4

BennyDeBock:
Adds testing for the different content-range options (start, within and tail)

Fix the bytecount calculation of the content-range headers

Add tests that should fail.

closes #2836

### DIFF
--- a/Sources/Vapor/HTTP/Headers/HTTPHeaders+ContentRange.swift
+++ b/Sources/Vapor/HTTP/Headers/HTTPHeaders+ContentRange.swift
@@ -249,13 +249,29 @@ extension HTTPHeaders.ContentRange {
 extension HTTPHeaders.Range.Value {
     
     ///Converts this `HTTPHeaders.Range.Value` to a `HTTPHeaders.ContentRange.Value` with the given `limit`.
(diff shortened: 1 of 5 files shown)


0xTim (changes requested):
(Adding review so GH reminds me when it's updated) - this looks great but we need to test edge cases more and ensure videos still play back etc if you could do a manual test as well)

PR #2848, opened 2022-06-10
Conform `Bool` to `Content`

josercc:
Conform `Bool` to `Content` to allow `Bool` types to be returned to the top level

```swift
app.get("isOK") { req in
    return true
}
```


0xTim:
@josercc can you add a test to ensure this works? Thanks!

josercc:
Yes, I will add a test case to test whether it is normal

0xTim (approved):
Thanks!

PR #2853, opened 2022-06-21
AsyncCommand has changed. You can use the async/await it command in vapor.

josercc:
```swift
struct TestCommandSignature: CommandSignature {
    @Argument(name: "age")
    var age: Int
}
struct TestCommand: AsyncCommand {
    typealias Signature = TestCommandSignature

    var help: String { "print age number" }

    func run(usingAsync context: CommandContext, signature: TestCommandSignature) async throws {
        context.console.output("I'm \(signature.age) years old".consoleText(color: .green))
    }
}

app.commands.use(TestCommand(), name: "test")
```


0xTim (changes requested):
There's already an `AsyncCommand` we just need to make it conform to `Command`

josercc:
> There's already an `AsyncCommand` we just need to make it conform to `Command`

I have considered that `AsyncCommand` conforms to the `Command` scheme, but when I want to implement the `Command` protocol mode,`Promise` cannot be created in the `ConsoleKit ` library to support the async method.

PR #2856, opened 2022-07-06
Add support for regular expression validations

K1NXZ:
Validate a regular expression pattern

Example:
```swift
struct TestContent: Codable, Validatable {
        validations.add("numbersOnly", as: String.self, is: .pattern("^[0-9]*$"))
    }
    
    let numbersOnly: String
    
    init(numbersOnly: String) {
        self.numbersOnly = numbersOnly
    }
}
```


0xTim (approved):
Thanks!

PR #2857, opened 2022-07-15
Add documentation comments and a defaulting subscript to `Storage`.

gwynne:
The new subscript simplifies "provider" implementations that extend `Application` and use its `Storage` instance without complex initialization requirements:

```swift
extension Application {
    public struct Foo {
        final class Storage { /* content which needs no special initialization */ }
        struct Key: StorageKey { typealias Value = Storage }
        let application: Application

        // Before:
        var storage: Storage {
            if self.application.storage[Key.self] == nil { self.initialize() }
            return self.application.storage[Key.self]!
        }

        func initialize() { self.application.storage[Key.self] = .init() }

        // After:
        var storage: Storage { self.application.storage[Key.self, default: .init()] }
```

### DIFF
--- a/Sources/Vapor/Utilities/Storage.swift
+++ b/Sources/Vapor/Utilities/Storage.swift
@@ -18,7 +18,7 @@ public struct Storage {
         }
     }
     
-    /// The logger provied to shutdown closures.
+    /// The logger provided to shutdown closures.
     let logger: Logger
 
     /// Create a new ``Storage`` container using the given logger.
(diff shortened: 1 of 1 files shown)


Sources/Vapor/Utilities/Storage.swift
```diff
@@ -34,10 +44,26 @@ public struct Storage {
...
+    /// value to the storage if the key does not already exist. Similar to
+    /// ``Swift/Dictionary/subscript(key:default:)``. The `defaultValue` autoclosure
+    /// is evaluated only when the key does not already exist in the container.
+    public subscript<Key>(_ key: Key.Type, default defaultValue: @autoclosure () -> Key.Value) -> Key.Value
+        where Key: StorageKey
+    {
+        mutating get {
+            if let existing = self[key] { return existing }
+            self.set(Key.self, to: defaultValue())
+            return self[key]!
```
0xTim, line 57: I mean I'm not enamored with the force unwrap even though it's safe. One option would be

```suggestion
            let new = defaultValue()
            self.set(Key.self, to: new)
            return new
```

0xTim (approved):
LGTM, one potential clean up

vzsg (approved):
I agree with Tim about removing that force unwrap, otherwise LGTM 👍🏻

PR #2859, opened 2022-07-19
fix: validate each not taking required parameter into account

BasPeter:
- Required parameter in add(each) now taken into account for validation
- Add tests for required false test case


0xTim (approved):
Thanks!

PR #2865, opened 2022-08-04
Implement support for custom verify callbacks

Lukasa:
Motivation

When using NIOSSL it is sometimes necessary to completely take over the
certificate verification logic. NIOSSL exposes a callback for this, but
it's currently hidden from Vapor users. We should let them get access to
this callback.

Modifications

- Added the callback to the HTTPServer configuration struct.
- Plumbed the callback through.
- Added some invalid test certs to the resources for the tests.
- Added a test to confirm the override functions correctly.

Result

Users can override client cert validation.


0xTim (approved):
Thanks!

PR #2866, opened 2022-08-06
Redirect keep path redirected from

iamtuxmeister:
This is a non breaking change that add a property appendNext: Bool to the redirectMiddleware() method.
This will allow capturing the next get parameter from the url to redirect back to the source of the authentication request.

### DIFF
--- a/Sources/Vapor/Authentication/RedirectMiddleware.swift
+++ b/Sources/Vapor/Authentication/RedirectMiddleware.swift
@@ -3,16 +3,24 @@ extension Authenticatable {
     ///
     /// - parameters:
     ///    - path: The path to redirect to if the request is not authenticated
-    public static func redirectMiddleware(path: String) -> Middleware {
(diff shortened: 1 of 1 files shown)


0xTim:
This isn't needed because the closure for the `RedirectMiddleware` allows you to specify this in a flexible way:

```swift
let redirectMiddleware = User.redirectMiddleware { req -> String in
  return "/login?authRequired=true&next=\(req.url.path)"
}
```

iamtuxmeister:
Thank you, that is a far more elegant solution than my hammer.
I am a perl programmer, working toward joining the 21st century with swift.
The swift magic that can tell the difference between the 6 different things
named the same, and know where to pass the closure is still a little
mysterious to me.

Would you update the documentation on the website?

Thank you

0xTim:
Tracking at https://github.com/vapor/docs/issues/699

PR #2870, opened 2022-08-12
Add support for using AsyncCommand and booting in fully async context

gwynne:
Support for `AsyncCommand` goes through a new `asyncCommands` property on `Application`, which supersedes `commands`. Anything added to `commands` - including Vapor's butilin default commands - is automatically included in the `asyncCommands` (assuming the async entry point methods are called).

To correctly boot Vapor from an `async` entry point, call `try await app.run()` instead of `try app.run()`; it's a subtle change, but makes a significant difference.

This is an alternative to #2853 and vapor/console-kit#175; this PR should not be merged if those are, and vice versa.


Joannis:
@gwynne is this PR ready for review? If so, can you update the branch and tag me as reviewer?

gwynne:
This PR is defunct, I didn't mean to leave it open.

PR #2873, opened 2022-08-16
Async Next Steps

0xTim:
This PR has a number of aims:

* Allow easier adoption of `Sendable` in Vapor
* Allow user's to back-deploy async/await to older OSes
* Ensure that we're safe in an async/await world.

To achieve this, the PR bumps the minimum supported Swift version to 5.6. This allows us to adopt `Sendable` and backdeploy Concurrency without needing a ton of duplication all over the code base.

This PR also starts deprecating `Storage` to move to an actor based `AsyncStorage` to ensure we can handle reads and writes from multiple concurrent contexts. This is going to be a far-reaching change as almost everything is underpinned by `Storage` so migrating away is going to require some tricks (like semaphores to ensure we can synchronously wait in non-async contexts), shuffling data between legacy storage systems and new storage systems and forcibly nudging users to use the new APIs

### DIFF
--- a/Sources/Vapor/Sessions/AsyncSessions.swift
+++ b/Sources/Vapor/Sessions/AsyncSessions.swift
@@ -29,4 +29,8 @@ public final actor AsyncSession {
     public func destroy() {
         self.isValid = false
     }
+    
+    public func set(_ key: String, to value: String?) {
+        self.data[key] = value
+    }
 }
--- a/Sources/Vapor/Sessions/AsyncSessions.swift
+++ b/Sources/Vapor/Sessions/AsyncSessions.swift
@@ -7,7 +7,7 @@
 /// This is the version of ``Session`` for use with `async`/`await`
 public final actor AsyncSession {
     /// This session's unique identifier. Usually a cookie value.
-    public var id: SessionID?
+    public let id: SessionID?
 
     /// This session's data.
     public var data: SessionData
@@ -17,7 +17,7 @@ public final actor AsyncSession {
 
     /// Create a new `Session`.
     ///
-    /// Normally you will use `Request.session()` to do this.
+    /// Normally you will use ``Request.asyncSession`` to do this.
     public init(id: SessionID? = nil, data: SessionData = .init()) {
         self.id = id
         self.data = data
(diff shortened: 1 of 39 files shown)


Sources/Vapor/Sessions/AsyncSessions.swift
```diff
@@ -0,0 +1,41 @@
...
+        self.isValid = false
+    }
+    
+    
+    /// Set the underlying session data in a safe way
+    /// - Parameters:
+    ///   - key: The key for that piece of session data
+    ///   - value: The value of the session data
+    public func set(_ key: String, to value: String?) {
+        self.data[key] = value
```
jdmcd, line 39: Just out of curiosity - should setting this data be guarded against `isValid`?

clayellis:
I'd like to see the `async*` prefix that was introduced to parameter names (not necessarily type names) dropped from the API. It starts to get very noisy to see `await req.asyncSession.foo` or `await req.auth.asyncLogin`. The presence of the `await` keyword should be indication enough that you're using the async variant. That said — I know there are valid reasons for introducing the `async*` variant since most of the time they're using different types. But maybe there's an elegant solution here?

Sources/Vapor/Sessions/AsyncSessions.swift
0xTim: Potentially - or at least setting it to true

0xTim:
@clayellis I agree the `async*` prefixes are laborious to say the least. But we don't want to break the public API until Swift 6 which means we can't just change all the property types and async-ness without

clayellis:
@0xTim makes perfect sense — and I'm glad to hear that the plan is (or at least sounds like) to drop that prefix with Vapor 5/Swift 6

0xTim:
@clayellis yes to be clear - when Vapor 5 is released when Swift 6 is out all the `async*` prefixes will disappear, as will any mention of `EventLoop`s!

Kyle-Ye:
Since Swift 5.7 is officially landing, is there any plan to make progress on this PR? Or are there any other blocking items?

0xTim:
@Kyle-Ye just time - I'm slowly working through it alongside a few other things

Gerzer:
Is there any movement on this? It’s currently very difficult to achieve full concurrency safety with Vapor because of the lack of `Sendable` conformance for many types that should be thread-safe (_e.g._, with `NIOLock`) but don’t explicitly adopt `Sendable`.

0xTim:
Closing in favour of #3000

PR #2874, opened 2022-08-17
Add missing protocol `ExpressibleByStringLiteral` to `HTTPHeaders.Name`

grahamburgsma:
Add missing protocol `ExpressibleByStringLiteral` to `HTTPHeaders.Name`. The implementation `init(stringLiteral:)` was there, but the actual protocol was missing.


0xTim (approved):
Thanks!

PR #2875, opened 2022-08-18
Fix Leaf Downstream CI

0xTim:
Adds the required env var for Leaf to ensure the downstream tests still work


gwynne (changes requested):
I prefer to use the `env` property rather than setting env direction in the shell: https://github.com/vapor/ci/blob/reusable-workflows/.github/workflows/run-unit-tests.yml#L92-L93

PR #2876, opened 2022-08-21
Update comment in DebuggableError.swift

sidepelican:
Fix incorrect comment. Now default logLevel is `.warning`( changed in #2598  )


tonyarnold (approved):
Good catch!

0xTim (approved):
Thanks!

PR #2764, opened 2022-01-18
Update Server Configuration after start

0xTim:
This ensures that the `Application`'s configuration is updated correctly when the server starts. This allows you to retrieve correct hostname and ports from the `Application`.

Resolves #2755

PR #2882, opened 2022-09-05
Fixing issue #2755: updating config storage inside HTTPServer

### LINKED ISSUE #2755: http server configuration has wrong hostname and port if they are set through cli ( -H hostname -p port )
### Describe the bug

It looks like that even if the application has been started setting custom hostname and port via CLI, the hostname and port properties of app.http.server.configuration return the default 127.0.0.1:8080.
Some other properties may be incorrect too, I didn't test those.

### To Reproduce

1. Add a line to print hostname and port.
```swift
app.logger.log(level: .warning, "\(app.http.server.configuration.hostname):\(app.http.server.configuration.port)"
```
2. Run the application setting custom hostname and port.
```shell
vapor run serve --hostname 192.168.1.14 --port 8081
```
3. Output is
```
[ WARNING ] 127.0.0.1:8080
[ NOTICE ] Server starting on http://192.168.1.14:8081
```

### Expected behavior

I think that configuration should reflect the real hostname and port that the server is running on.
```
[ WARNING ] 192.168.1.14:8081
[ NOTICE ] Server starting on http://192.168.1.14:8081
```

### Environment

* Vapor Framework version: 4.54.0
* Vapor Toolbox version: 18.3.3
* OS version: macOS Monterey 12.1

### Additional context

If hostname and port are changed programmatically in configure.swift configure method then they are successfully set to the correct values.

rkreutz:
We can update application.storage from within HTTPServer, this way we can keep any changes that happen to the configuration internally up-to-date with the application storage.

This possibly has the least changes and less surface of potential flaws, since we are only adding an extra param and working on top of it. However, now we are setting the application storage from within `HTTPServer`, there is no issue with that, is just that now we have 2 places changing the storage for the config.

Resolves #2755

### DIFF
--- a/Sources/Vapor/HTTP/Server/HTTPServer.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServer.swift
@@ -232,8 +232,23 @@ public final class HTTPServer: Server {
         return connection.channel.closeFuture
     }
 
+    public var configuration: Configuration {
+        get { _configuration }
+        set {
+            guard !didStart else {
+                _configuration.logger.warning("Cannot modify server configuration after server has been started.")
+                return
+            }
+            _configuration = newValue
+        }
+    }
+
     private let responder: Responder
-    private var configuration: Configuration
+    private var _configuration: Configuration {
+        willSet {
+            self.application.storage[Application.HTTP.Server.ConfigurationKey.self] = newValue
+        }
+    }
     private let eventLoopGroup: EventLoopGroup
     
     private var connection: HTTPServerConnection?
@@ -250,7 +265,7 @@ public final class HTTPServer: Server {
     ) {
         self.application = application
         self.responder = responder
-        self.configuration = configuration
+        self._configuration = configuration
         self.eventLoopGroup = eventLoopGroup
         self.didStart = false
         self.didShutdown = false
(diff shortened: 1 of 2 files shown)


rkreutz:
this is an alternative to https://github.com/vapor/vapor/pull/2881

0xTim:
This should have been solved in #2764 - does that match with what you were seeing?

rkreutz:
In that PR you have reflected changes in configuration back to HTTPServer, which is just part of the fix, we also need to update `application.storage[Application.HTTP.Server.ConfigKey]` with the new configuration otherwise accessing `application.http.server.configuration` will be outdated

0xTim:
@rkreutz ah I see - do you want to update the PR so the test reflects the missing config? Thanks!

rkreutz:
Should we go with this approach then rather than #2881 ?

0xTim:
Yeah this feels like a better change, keeping it as a struct if possible is desirable as we adopt `Sendable` throughout the codebase

rkreutz:
@0xTim added the unit tests there [d0c6778](https://github.com/vapor/vapor/pull/2882/commits/d0c67785aeff7bcf67bf571754d294c82945b426), I've reverted the changes and ran the tests and confirmed they were failing, and after reapplying the change the tests pass

rkreutz:
Hey @0xTim anything else here you'd like to have added?

0xTim (approved):
Apologies, got lost in the queue! This is good to go when CI passes

PR #2883, opened 2022-09-06
Reducing minimum target for Concurrency related code

rkreutz:
Since Xcode 13.2, Swift Concurrency support has been extended to macOS 10.15, iOS 13, tvOS 13 and watchOS 6 therefore we can safely lower minimum target version on Concurrency related code

### DIFF
--- a/Sources/Development/routes.swift
+++ b/Sources/Development/routes.swift
@@ -232,7 +232,7 @@ public func routes(_ app: Application) throws {
     }
 
     #if compiler(>=5.5) && canImport(_Concurrency)
-    if #available(macOS 12, *) {
+    if #available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) {
         let asyncRoutes = app.grouped("async").grouped(TestAsyncMiddleware(number: 1))
         asyncRoutes.get("client") { req async throws -> String in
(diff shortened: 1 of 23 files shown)


0xTim:
Unfortunately it's not a simple as this. Whilst it works for most people,  anyone on older OSes with a new Xcode won't be able to compile and this doesn't work with Xcode 13.0 and 13.1. To fix this we'd need to duplicate all the concurrency API for old Xcode and new Xcode (see [NIO as an example](https://github.com/apple/swift-nio/blob/main/Sources/NIOCore/AsyncAwaitSupport+OldXcodes.swift). The maintenance burden is too great for us to commit to (and test and keep up to date etc).

We do have plans to solve this by dropping Swift 5.4 and 5.5 with the next chunk of async work - see https://blog.vapor.codes/posts/async-next-steps/

0xTim:
Unfortunately that would be a breaking change, which we can't do. If someone has 5.5.0 either via Xcode or in Docker and run an update they'd get compilation errors, which we can't do I'm afraid. The best workaround is probably VSCode and remote dev containers

rkreutz:
I'll just run my code on my forks, but hopefully this will get sorted eventually, thanks anyway

PR #2887, opened 2022-09-12
Migrate tests from NIOAtomics to Atomics

MahdiBM:
NIOAtomics has been deprecated in favor of Atomics. We should move everything to Atomics.


0xTim (approved):
Thanks!

PR #2888, opened 2022-09-13
Remove dangerous and problematic re-export of XCTest

gwynne:
This removes the re-export of the `XCTest` module from the `XCTVapor` module. This incorrect, ill-advised export causes problems when building with different SDKs on macOS and is generally a nuisance without being a help to anyone. While this would normally be a semver-major break, the fact that it applies only to a testing-related module, and that most unit test files end up having an `import XCTest` statement by default in any event, sufficiently justifies the source compatibility break in light of the difficulties the improper re-exporting causes.

### DIFF
--- a/Sources/XCTVapor/Exports.swift
+++ b/Sources/XCTVapor/Exports.swift
@@ -1,2 +1,1 @@
-@_exported import XCTest
 @_exported import Vapor


jdmcd (approved):

Joannis (approved):

0xTim:
What's this in response to? I know several projects that this will break the build for so I'd say the bar will be pretty high for a source breaking change

0xTim:
Not to mention [every project built using the template](https://github.com/vapor/template/blob/main/Tests/AppTests/AppTests.swift)

dmonagle:
This appears to cause a build issue in the now released Xcode 14. 

While this may break builds for anybody not importing XCTest, it certainly DOES break builds for anybody using an up-to-date toolchain. It's a tough situation.

I can't confirm this error also appears building from the command line, but I'm happy to follow that up (and one would suspect that it will also fail). 

image

0xTim:
(Pressed the wrong button but might as well keep it - feel free to ask for it to be reopen if you want!)

gwynne:
Yeah, it turned out you had to be in an unusual situation to hit that error. The export is still hilariously wrong, but it's not worth the disruption.

PR #2893, opened 2022-10-06
Update FileIO streamFile to allow async / await in whenComplete closure

omrd:
Currently the 'whenComplete' is called, it does not support async / await. Trying to resolve this issue as I would like to await some updates once completed.

```swift
        let r = request.fileio.streamFile(at: f.path(study: study)) { result in
            switch result {
            case .success(): 
               doSomeSuccessProcessing() //works fine
               await doSomeSuccessProcessing() //doesn't work
            case .failure(let err):
               doSomeProcessing() //works fine
               await doSomeProcessing() //doesn't work
            }
       }
```


0xTim:
@omrd yeah we need full async/await support for FileIO - if you want to implement it that would be great!

PR #2894, opened 2022-10-06
Prevent vapor from crashing during a crash in third-party code, obfuscating the real problem

Joannis:
If a third-party library or user defined code crashed a Vapor app, the Vapor Application deinit will crash the app before the real issue pops up. This leads to frustrating debug sessions

### DIFF
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -161,7 +161,8 @@ public final class Application {
     deinit {
         self.logger.trace("Application deinitialized, goodbye!")
         if !self.didShutdown {
-            assertionFailure("Application.shutdown() was not called before Application deinitialized.")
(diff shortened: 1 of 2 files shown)


gwynne (approved):
Yeah, I gotta come down in favor of this, ultimately. The importance of calling out the misbehavior is not enough to justify the massive penalties to debugging and usability that people have been dealing with.

0xTim:
@Joannis ready for this to be merged?

0xTim:
I'm wondering if we should add a test for this

Joannis:
I'll add a test, because the simple condition is "the test doesn't crash"

Joannis:
> They should be fixed by #2897 

Let's merge after tests succeed then

PR #2839, opened 2022-05-31
Fix for #2574 Missing quote from value

paunik:
- now quotes are always enforced in `DirectiveSerializer.serialize`
- adjusted tests

Closes #2574

PR #2895, opened 2022-10-10
Conforms `Request.Body` to  `AsyncSequence`

mcritz:
This PR wraps `Request.Body.drain()` as a modern Swift `AsyncSequence<ByteBuffer, Error>`.

This is useful to _stream_ bytes from request rather than collecting them in memory. Example: A route could handle a multigigbyte file upload like this:

```swift
do {
    let nioFileHandle = try NIOFileHandle(path: filePath, mode: .write)
    var offset: Int64 = 0
    
    for try await bytes in req.body {
        try await req.application.fileio.write(fileHandle: nioFileHandle,
                                               toOffset: offset,
                                               buffer: bytes,
                                               eventLoop: req.eventLoop).get()
        offset += Int64(bytes.readableBytes)
        try nioFileHandle.close()
    }
} catch {
   ...
}
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -53,9 +53,6 @@ let package = Package(
         // Swift collection algorithms
         .package(url: "https://github.com/apple/swift-algorithms.git", from: "1.0.0"),
         
-        // Advanced swift async/await
-        .package(url: "https://github.com/apple/swift-async-algorithms", from: "0.0.3"),
-        
         // WebSocket client library built on SwiftNIO
         .package(url: "https://github.com/vapor/websocket-kit.git", from: "2.0.0"),
         
@@ -89,7 +86,6 @@ let package = Package(
             .product(name: "NIOWebSocket", package: "swift-nio"),
             .product(name: "Crypto", package: "swift-crypto"),
             .product(name: "Algorithms", package: "swift-algorithms"),
-            .product(name: "AsyncAlgorithms", package: "swift-async-algorithms"),
             .product(name: "RoutingKit", package: "routing-kit"),
             .product(name: "WebSocketKit", package: "websocket-kit"),
             .product(name: "MultipartKit", package: "multipart-kit"),
--- a/Package.swift
+++ b/Package.swift
@@ -52,7 +52,7 @@ let package = Package(
         
         // Swift collection algorithms
         .package(url: "https://github.com/apple/swift-algorithms.git", from: "1.0.0"),
-        
+
         // WebSocket client library built on SwiftNIO
         .package(url: "https://github.com/vapor/websocket-kit.git", from: "2.0.0"),
         
--- a/Package.swift
+++ b/Package.swift
@@ -33,7 +33,7 @@ let package = Package(
         .package(url: "https://github.com/swift-server/swift-backtrace.git", from: "1.1.1"),
         
         // Event-driven network application framework for high performance protocol servers & clients, non-blocking.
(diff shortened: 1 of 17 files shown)


mcritz:
Hello! I’m starting this Draft PR as a conversation starter for adding `AsyncThrowingStream` API to Vapor. If this work is considered worthwhile by the maintainers, then I can perform any changes and write tests.

Cheers

0xTim:
Just to carry over from Discord - this looks great! Will need some tests though

mcritz:
I have refactored this PR to support backpressure with AsyncThrowingChannel.

Adds a dependency on `swift-async-algorithms`.

The call site API changes a little bit:

```
do {
    let nioFileHandle = try NIOFileHandle(path: filePath, mode: .write)
    var offset: Int64 = 0
    let channel = req.asyncThrowingChannel
    for try await byteBuffer in channel {
        try await req.application.fileio.write(fileHandle: nioFileHandle,
                                               toOffset: offset,
                                               buffer: byteBuffer,
                                               eventLoop: req.eventLoop).get()
        offset += Int64(byteBuffer.readableBytes)
    }
    try nioFileHandle.close()
} catch {
    ...
}
```

mcritz:
Hi maintainers!

#2839 is ready for review. I am open to any changes.

In addition to the included test, I have built a [test project](https://github.com/mcritz/VaporUploads/tree/feature/async) that proves out the use case. I ran a Swift Concurrency Profile run against the test project target while uploading files and it met my expectations for concurrent tasks.

Cheers!

Package.swift
```diff
@@ -52,7 +52,7 @@ let package = Package(
         
         // Swift collection algorithms
         .package(url: "https://github.com/apple/swift-algorithms.git", from: "1.0.0"),
-
+
```
Joannis, line 55: Could we get rid of this change? Just for cleanliness

Sources/Vapor/Concurrency/Request+Concurrency.swift
```diff
@@ -0,0 +1,171 @@
...
+                let result = source.yield(buffer)
+                // Inspect the source view and handle outcomes
+                switch result {
+                case .dropped:
+                    // The consumer dropped the sequence.
+                    // Inform the producer that we don't want more data
+                    // by returning an error in the future.
+                    return request.eventLoop.makeFailedFuture(CancellationError())
+                case .stopProducing:
+                    // The consumer is consuming fast enough for us.
```
Joannis, line 147: Isn't it the opposite, the consumer is **not** consuming fast enough?

Joannis (approved):
LGTM, just two nits

Package.swift
mcritz: done

mcritz:
@Joannis Updated! Thanks for review.

0xTim (approved):
This looks great thanks!

0xTim:
@gwynne any final thoughts before merging

gwynne (approved):
I could nitpick a few lines over different ways to write exactly the same code that make no actual difference whatsoever... 😆 But nah, LGTM!

gwynne:
Although the CI failures should be addressed as appropriate - I probably should've finished scrolling before commenting 😅

0xTim:
@mcritz we either need to gate this behind 5.7 only or rewrite to allow it to work on older versions of Swift

mcritz:
@0xTim On it. 

@gwynne Please nitpick. I'd like to learn.

PR #2898, opened 2022-10-13
Ensure response body is collected on the EventLoop

0xTim:
This ensures that the request body is collected on the correct `EventLoop` no matter where if was called from. This avoids hitting an assertion failure when calling 
this from a Swift Concurrency context.


0xTim:
cc @weissi

Sources/Vapor/HTTP/BodyStream.swift
```diff
@@ -45,7 +45,13 @@ public protocol BodyStreamWriter {
 extension BodyStreamWriter {
     public func write(_ result: BodyStreamResult) -> EventLoopFuture<Void> {
         let promise = self.eventLoop.makePromise(of: Void.self)
-        self.write(result, promise: promise)
+        if self.eventLoop.inEventLoop {
```
weissi, line 48: @0xTim this isn't 100% correct: `inEventLoop` is "best effort": if it says we're on the EventLoop that's true but the contrary isn't necessarily true. Some EventLoop implementations might not be able to reliably determine if you're on them.
So this'll work just fine for Vapor today which always uses `MultiThreadedEventLoopGroup` which can reliably determine this but I'd rewrite it similar to this

```
    func write(_ result: BodyStreamResult, promise: EventLoopPromise<Void>?) {
        func write0() {
            switch result {
            case .buffer(var buffer):
                self.buffer.writeBuffer(&buffer)
            case .error(let error):
                self.promise.fail(error)
            case .end:
                self.promise.succeed(self.buffer)
            }
            promise?.succeed(())
        }

        if self.eventLoop.inEventLoop {
            write0()
        } else {
            self.eventLoop.execute { write0() }
        }
    }
```

which works regardless and just hits the "slow path" if the EventLoop can't reliably determine that we're on the EL already.

gwynne (dismissed):
What @weissi said 😅

Sources/Vapor/HTTP/BodyStream.swift
0xTim: @weissi just to be clear is the difference that the promise is fulfilled on the event loop in your example but not in mine? Updated!

weissi: doens't matter where the promise is fulfilled. The issue is that this some other EventLoop implementations (none that I would be aware of in the real world right now) might always return `false` from `eventLoop.inEventLoop`. And I think your original code would then up in an infinite loop if that were to happen.

weissi (approved):
Thank you, LGTM

0xTim:
Added to #3055

PR #2900, opened 2022-10-17
Adds action to update README on new backer

ptoffy:
Adds a GitHub Action to automatically update the `README.md` when someone sponsors the organisation

### DIFF
--- a/.github/workflows/sponsors.yml
+++ b/.github/workflows/sponsors.yml
@@ -1,8 +1,5 @@
 name: Generate Sponsors README
-on:
-  workflow_dispatch:
-  schedule:
-    - cron: 30 15 * * 0-6
+on: workflow_dispatch
 jobs:
   deploy:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/sponsors.yml
```diff
@@ -0,0 +1,29 @@
...
+          marker: "backers"
+          organization: true
+
+      # Here we can eventually add sponsors too rather than just backers
+
+      - name: Deploy to GitHub Pages 🚀
+        uses: JamesIves/github-pages-deploy-action@v4
+        with:
+          branch: main
+          folder: "."
```
0xTim, line 29: We don't need this as we're not deploying to Pages right?

0xTim (commented):
So a few questions - does this actually commit the changes (I'm assuming we're going to have to do that manually) and is there a way to sort the backers? Just had a quick look and I don't think so which is fine

.github/workflows/sponsors.yml
ptoffy: I updated that part of the action to just commit the new changes without deploying them to pages.
Backers aren't sorted

0xTim (approved):

PR #2901, opened 2022-10-20
Make Services in Vapor Usable

0xTim:
This adds real support to Vapor to make it easy to integrate different services that can be tested.

For example, if you have a service defined as:

```swift
protocol MyService {
    func `for`(_ request: Request) -> MyService
    func doSomething() -> String
}
```

You may then have a real implementation:

```swift

import Vapor

struct MyRealService: MyService {
    let logger: Logger
    let eventLoop: EventLoop
    
    func `for`(_ request: Vapor.Request) -> MyService {
        return MyRealService(logger: request.logger, eventLoop: request.eventLoop)
    }
    
    func doSomething() -> String {
        return "Tada"
    }
    
}
```

This is a very contrived example, but shows a service that needs a `Logger` and `EventLoop` - things that are normally tied to specific requests. Doing this in a safe and testable way involves a lot of boilerplate. This moves the boilerplate into Vapor to make it easier to do.

Now, you can define your services:

```swift
extension Application.Services {
    var myService: Application.Service<MyService> {
        .init(application: self.application)
    }
}

extension Request.Services {
    var myService: MyService {
        self.request.application.services.myService.service.for(request)
    }
}
```

Configure it in **configure.swift**:

```swift
app.services.myService.use { app in
    MyRealService(logger: app.logger, eventLoop: app.eventLoopGroup.next())
}
```

And then use it in routes:

```swift
app.get("myService") { req -> String in
    let thing = req.services.myService.doSomething()
    return thing
}
```

## Testing

This approach makes it very easy to write test services as well. With the protocol defined above, you can create an implementation for testing:

```swift
struct MyFakeService: MyService {
    let cannedResponse: String
    let eventLoop: EventLoop
    let logger: Logger
    
    func `for`(_ request: Vapor.Request) -> App.MyService {
        return MyFakeService(cannedResponse: self.cannedResponse, eventLoop: request.eventLoop, logger: request.logger)
    }
    
    func doSomething() -> String {
        return cannedResponse
    }   
}
```

And then you can configure your test application and use it:

```swift
let myFakeServicce = MyFakeService(cannedResponse: testString, eventLoop: app.eventLoopGroup.next(), logger: app.logger)        
app.services.myService.use { _ in
    myFakeServicce
}
    
try app.test(.GET, "myService", afterResponse: { res in
    XCTAssertEqual(res.status, .ok)
    XCTAssertEqual(res.body.string, testString)
})
``

### DIFF
--- a/Sources/Vapor/Services/Service.swift
+++ b/Sources/Vapor/Services/Service.swift
@@ -1,0 +1,53 @@
+public extension Application {
+    struct Service<ServiceType> {
+
+        let application: Application
+
+        public init(application: Application) {
+            self.application = application
+        }
+
+        public struct Provider {
+            let run: (Application) -> ()
+
+            public init(_ run: @escaping (Application) -> ()) {
+                self.run = run
+            }
+        }
+
+        final class Storage {
+            var makeService: ((Application) -> ServiceType)?
+            init() { }
+        }
+
+        struct Key: StorageKey {
+            typealias Value = Storage
+        }
+
+        public var service: ServiceType {
+            guard let makeService = self.storage.makeService else {
+                fatalError("No service configured for \(ServiceType.self)")
+            }
+            return makeService(self.application)
+        }
+
+        public func use(_ provider: Provider) {
+            provider.run(self.application)
+        }
+
+        public func use(_ makeService: @escaping (Application) -> ServiceType) {
+            self.storage.makeService = makeService
+        }
+
+        func initialize() {
+            self.application.storage[Key.self] = .init()
+        }
+
+        private var storage: Storage {
+            if self.application.storage[Key.self] == nil {
+                self.initialize()
+            }
+            return self.application.storage[Key.self]!
+        }
+    }
+}
--- a/Sources/Vapor/Services/Service.swift
+++ b/Sources/Vapor/Services/Service.swift
@@ -1,7 +1,7 @@
 import NIOConcurrencyHelpers
 
-public extension Application {
-    struct Service<ServiceType> {
+extension Application {
+    public struct Service<ServiceType> {
 
         let application: Application
 
@@ -43,15 +43,18 @@ public extension Application {
             self.storage.makeService.withLockedValue { $0 = makeService }
         }
 
-        func initialize() {
-            self.application.storage[Key.self] = .init()
+        func initialize() -> Storage {
+            let new = Storage()
+            self.application.storage[Key.self] = new
+            return new
         }
 
         private var storage: Storage {
-            if self.application.storage[Key.self] == nil {
-                self.initialize()
+            if let storage = application.storage[Key.self] {
+                return storage
+            } else {
+                return self.initialize()
             }
-            return self.application.storage[Key.self]!
         }
     }
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by 0xTim** [comment]
One point for comment is it might be worth defaulting to making service access `async` to avoid immediately deprecating everything when #2873 lands

**Review by Andrewangeta — APPROVED** [review]

File: Tests/VaporTests/ServiceTests.swift [hunk 1]
```diff
@@ -4,55 +4,108 @@ final class ServiceTests: XCTestCase {
     func testReadOnly() throws {
         let app = Application(.testing)
         defer { app.shutdown() }
-
+
```
Joannis, line 7: Nit, but there's a lot of whitespace changes.

Sources/Vapor/Services/Service.swift
```diff
@@ -0,0 +1,53 @@
...
+        let application: Application
+
+        public init(application: Application) {
+            self.application = application
+        }
+
+        public struct Provider {
+            let run: (Application) -> ()
+
+            public init(_ run: @escaping (Application) -> ()) {
```
Joannis, line 13: Shouldn't we mark these `@Sendable` from now on?

Sources/Vapor/Services/Service.swift
```diff
@@ -0,0 +1,53 @@
...
+                fatalError("No service configured for \(ServiceType.self)")
+            }
+            return makeService(self.application)
+        }
+
+        public func use(_ provider: Provider) {
+            provider.run(self.application)
+        }
+
+        public func use(_ makeService: @escaping (Application) -> ServiceType) {
```
Joannis, line 38: `@Sendable`?

Sources/Vapor/Services/Service.swift
```diff
@@ -0,0 +1,53 @@
...
+            self.storage.makeService = makeService
+        }
+
+        func initialize() {
+            self.application.storage[Key.self] = .init()
+        }
+
+        private var storage: Storage {
+            if self.application.storage[Key.self] == nil {
+                self.initialize()
```
Joannis, line 48: I know this works fine, but can we make `initialize()` return the newly created entity? Then return that here.

Sources/Vapor/Services/Service.swift
```diff
@@ -0,0 +1,53 @@
+public extension Application {
+    struct Service<ServiceType> {
```
Joannis, line 2: `public struct` instead of `public extension`? I tend to be explicit in marking things public in libraries/frameworks, so future changes don't accidentally expose internal API.

Joannis (changes requested):

Joannis:
Great addition btw!

FullQueueDeveloper:
@0xTim @Joannis @Andrewangeta What's the best way I can help get this merged? 

1. I'm using this code in a project
2. Tim has copy-pasted this into Discord at least three times

Joannis:
What's the goal of this PR?

0xTim:
@Joannis what do you mean?

Joannis:
@0xTim sorry for missing your message;
I mean, what's the intention behind this new feature? To make it easier to work with `app.storage`? Or is there another goal?

dimitribouniol (approved):

mynona:
Would it be possible to add this changed approach to the vapor documentation?

0xTim:
@mynona yeah its on the list, PRs would be great!

PR #2904, opened 2022-11-03
Fix 24h timeformat for `expire` and `last-modified` header

patrick-zippenfenig:
`Expire` and `Last-Modified` header were encoding the hour part in 12h format (`hh`) instead of 24h format (`HH`). This results in timestamps being 12 hours off for all afternoon hours. This fixes the format used to follow the [spec](https://httpwg.org/specs/rfc9110.html#field.last-modified) correctly and adds tests to ensure no regressions.


0xTim (approved):
Thanks!

PR #2905, opened 2022-11-08
Fix Stream Callback never being called in close

0xTim:
Fixes a bug where abandoned requests mean that the stream callback is never invoked. This happens when a client sends a request, the server starts processing it but hasn't constructed the `Response` and then the client closes early. The response is discarded, as is the stream callback so it never gets invoked. This fixes that issue


Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
```diff
@@ -50,13 +50,17 @@ final class HTTPServerHandler: ChannelInboundHandler, RemovableChannelHandler {
                     if !keepAlive {
                         context.close(mode: .output, promise: nil)
                     }
+                case .failure(ChannelError.ioOnClosedChannel):
```
weissi, line 53: quick question: why only when `.ioOnClosedChannel`? I would've thought that this needs to be done on any write error. And even if there's no write error but the connection goes away we should notify the user, right?

CC @Lukasa

0xTim: Shouldn't `errorCaught(context:error:)` function be triggered in all cases as well so the error can be bubbled up and handled?

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
```diff
@@ -67,3 +71,11 @@ final class HTTPServerHandler: ChannelInboundHandler, RemovableChannelHandler {
         }
     }
 }
+
+struct ErrorBodyStreamWriter: BodyStreamWriter {
+    var eventLoop: EventLoop
+    var error: Error
+    func write(_ result: BodyStreamResult, promise: EventLoopPromise<Void>?) {
+        promise?.fail(ChannelError.ioOnClosedChannel)
```
Danappelxx, line 79:
```suggestion
        promise?.fail(error)
```

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
weissi: @0xTim the problem at hand isn't that _Vapor_ doesn't know about the error, it's that the _user_ doesn't learn about it, right? So I'm not sure why other errors aren't handled similarly. Not saying it's wrong, just don't quite follow

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
0xTim: Thanks!

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
```diff
@@ -51,12 +51,15 @@ final class HTTPServerHandler: ChannelInboundHandler, RemovableChannelHandler {
                         context.close(mode: .output, promise: nil)
                     }
                 case .failure(let error):
+                    if case .stream(let stream) = response.body.storage {
+                        stream.callback(ErrorBodyStreamWriter(eventLoop: request.eventLoop, error: error))
+                    }
```
0xTim, line 56: @weissi can you see any problem with this? I've taken a look at it and it seems to make sense we should call call `errorCaught` and trigger the stream callback if we have them

gwynne (approved):
LGTM... I think 😅

PR #2907, opened 2022-11-08
Fix unused generator parameter in Array.random(count:using:)

michal-tomlein:
The `generator` parameter in `Array.random(count:using:)` was unused in what appears to be a copy-paste error. This change passes it down to `FixedWidthInteger.random(using:)`, which was the original intention.

### DIFF
--- a/Sources/Vapor/Utilities/Array+Random.swift
+++ b/Sources/Vapor/Utilities/Array+Random.swift
@@ -21,7 +21,7 @@ extension Array where Element: FixedWidthInteger {
         where T: RandomNumberGenerator
     {
         var array: [Element] = .init(repeating: 0, count: count)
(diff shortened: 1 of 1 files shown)


Joannis (approved):

Joannis:
Thanks for spotting that detail! That's gotta be an oversight. Ping @0xTim / @gwynne

0xTim (approved):
Thanks!

0xTim:
Original comment:

The `generator` parameter in `Array.random(count:using:)` was unused in what appears to be a copy-paste error. This change passes it down to `FixedWidthInteger.random(using:)`, which seems to have been the original intention.

If that was not the original intention of the code author, I believe a comment should be added to explain why the parameter is unused.

PR #2914, opened 2022-11-16
Fix encoding/decoding an array of dates with URL Encoding

### LINKED ISSUE #2909: Array of dates are not encoded correctly as a query parameter
### Describe the bug

Attempting to encode an array of dates as a query param ignores the `dateEncodingStrategy` set.

### To Reproduce

Very basic example extracted from my codebase

```swift

func test(client: Client) async throws {
    // Breaking example
    let dates: [Date] = [
        Date(),
    ]
    try await client.get("example.com/multi-dates") { req in
        let encoder = URLEncodedFormEncoder(
            configuration: URLEncodedFormEncoder.Configuration(
                arrayEncoding: .values, 
                dateEncodingStrategy: .iso8601
            )
        )
        try req.query.encode([
            "dates": dates
        ], using: encoder)
        // This shows that the date is actually still seconds formatted instead of iso8601
        print("query: \(req.query)")
    }

    // Passing example
    let singleDate = Date()
    try await client.get("example.com/single-date") { req in
        let encoder = URLEncodedFormEncoder(
            configuration: URLEncodedFormEncoder.Configuration(
                arrayEncoding: .values, 
                dateEncodingStrategy: .iso8601
            )
        )
        try req.query.encode([
            "dates": singleDate
        ], using: encoder)
        // Corrected formatted date this time
        print("query: \(req.query)")
    }
}
```

1. Add Vapor package
2. Send run above block
3. Look at console when it is printed

### Expected behavior

Breaking example should output something like:
```
example.com/multi-dates?dates=2022-11-11&dates=2022-11-12 //...
```

But it actually outputs something like:
```
example.com/multi-dates?dates=12345&dates=23456 //...
```

### Environment

* Vapor Framework version: 4.67.1
* Vapor Toolbox version: 18.5.1 (but integrated into larger project via SPM)
* OS version: macOS Ventura 13.0 (22A380)

### Additional context

~ nil

Joannis:
Fixes a bug where an Array of Dates wouldn't be encoded or decoded when using URL encoding.

### DIFF
--- a/Sources/Vapor/URLEncodedForm/URLEncodedFormDecoder.swift
+++ b/Sources/Vapor/URLEncodedForm/URLEncodedFormDecoder.swift
@@ -332,6 +332,11 @@ private struct _Decoder: Decoder {
                 return try T(from: decoder)
             } else {
                 let value = self.values[self.currentIndex]
+                // Check if we received a date. We need the decode with the appropriate format.
+                guard !(T.self is Date.Type) else {
+                    return try configuration.decodeDate(from: value, codingPath: codingPath, forKey: nil) as! T
+                }
+                
                 if let convertible = T.self as? URLQueryFragmentConvertible.Type {
                     if let result = convertible.init(urlQueryFragmentValue: value) {
                         return result as! T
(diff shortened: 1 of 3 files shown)


0xTim (approved):
Thanks! Can you update the PR description for the release notes? Then feel free to merge

Issue #2906, opened 2022-11-08
Collecting the request body sometimes hangs


tancred:
### Describe the bug

A route that `.collect`s a request body hangs under certain conditions.

### To Reproduce

Download the attached Vapor project [hanging-collector.zip](https://github.com/vapor/vapor/files/9968465/hanging-collector.zip) and unpack it. Then follow the instructions in`readme.md` that's found in the root of the project. I've copied most of the instructions here.

#### Prepare
 
1. Start the app:

```
$ swift run Run serve --port 7687
```

2. Upload the sample image:

```
$ curl -i -T ./sample.jpg -X PUT http://127.0.0.1:7687/hanger
```

This should work out ok: there's a copy of `sample.jpg` in `/tmp/pix/hanger`.

#### Reproduce the hang

1. Start the app

```
$ swift run Run serve --port 7687
```

2. GET the image

```
curl -i http://127.0.0.1:7687/hanger
```

3. PUT the image

```
$ curl -i -T ./sample.jpg -X PUT http://127.0.0.1:7687/hanger
```

Often this will be enough. The vapor app has logged the PUT request but `curl` will now hang, and running `lsof -i -nP` will show the open connection:

```
$ lsof -i -nP
COMMAND     PID  USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
...
Run       22462 malte   12u  IPv4 0xd3a42b18da250499      0t0  TCP 127.0.0.1:7687 (LISTEN)
Run       22462 malte   13u  IPv4 0xd3a42b18da24e4d1      0t0  TCP 127.0.0.1:7687->127.0.0.1:55955 (ESTABLISHED)
curl      22503 malte    6u  IPv4 0xd3a42b18d947ef69      0t0  TCP 127.0.0.1:55955->127.0.0.1:7687 (ESTABLISHED)
```

Stopping the app now takes some time. After about 10 seconds the app quits with the following message:

```
[ ERROR ] Could not stop HTTP server: Abort.500: Server stop took too long.
ERROR: Cannot schedule tasks on an EventLoop that has already shut down. This will be upgraded to a forced crash in future SwiftNIO versions.
```

The hang is intermittent. I often have to retry from step 1 for the hang to occur. It's important that the app is restarted!

### Expected behavior

The PUT request should never hang.

### Environment

* Vapor Framework version: 4.67.1
* Vapor Toolbox version: 18.6.0
* OS version: macOS 12.6.1 (21G217)

```
$ swift --version
[...]

PR #2917, opened 2022-11-20
Prevent vapor streaming requests from randomly losing a first chunk, stalling the request

Joannis:
Fixes a bug where a streaming body from a request was not processing a chunk, causing the request to stall. This happened when a handler was being set at the exact moment a chunk was being processed, causing the chunk to be added to a now irrelevant array that was just processed while switching to a streaming callback.

See #2906


0xTim:
@Joannis still got test failures

vzsg:
Deadlocks are a b—

Sources/Vapor/Request/Request+BodyStream.swift
```diff
@@ -23,58 +23,65 @@ extension Request {
...
         }
 
         func write(_ chunk: BodyStreamResult, promise: EventLoopPromise<Void>?) {
-            // A lock is needed, because some pipeline users (and known unit tests) rely on this being handled synchronously.
             // See https://github.com/vapor/vapor/issues/2906
-            lock.withLockVoid {
+            if self.eventLoop.inEventLoop {
+                write0(chunk, promise: promise)
+            } else {
+                self.eventLoop.execute {
```
gwynne, line 38: It is my understanding that `EventLoop.execute(_:)` already does the "is in event loop" check internally, which in addition, can return false negatives in the normal course of execution - If the problem is that you were executing on the event loop unconditionally before and are skipping the extra "hop", this isn't a guaranteed solution.

Joannis: It's not supposed to be a guaranteed solution either. The main reason is that I think it's a waste of a hop. And it does _always_ [queue it for the next tick](https://github.com/apple/swift-nio/blob/main/Sources/NIOPosix/SelectableEventLoop.swift#L334). Not only does it fix the test, it also is a good optimisation to have.

Joannis: Ideally, the test's assumption is to be fixed as well of course.

0xTim (approved):
Thanks!

PR #2918, opened 2022-11-21
Add hooks for socket configuration

John-Connolly:
### Description

Some user may need more advanced socket options for Vapors HTTP server.  There is currently no way of specifying these options.  This PR solves the issue by adding hooks for the server bootstrap.  Many socket options such as ```SO_REUSEPORT``` need to be set before the server is started.  Providing users with a configuration handler seems like a logical way of accomplishing this.

Example usage

```swift
 app.http.server.configuration.serverChannelOptionConfiguration = { serverBootstrap in
        serverBootstrap.serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEPORT), value: SocketOptionValue(1))
 }

    app.http.server.configuration.childChannelOptionConfiguration = { serverBootstrap in
        serverBootstrap.childChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEPORT), value: SocketOptionValue(1))
 }

```

### DIFF
--- a/Sources/Vapor/HTTP/Server/HTTPServer.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServer.swift
@@ -155,7 +155,13 @@ public final class HTTPServer: Server {
 
         /// An optional callback that will be called instead of using swift-nio-ssl's regular certificate verification logic.
         public var customCertificateVerifyCallback: NIOSSLCustomVerificationCallback?
+        
+        /// An optional configuration for the setting the servers channels options
+        public var serverChannelOptionConfiguration: (ServerBootstrap) -> (ServerBootstrap)
 
+        /// An optional configuration for the setting the servers child channels options
+        public var childChannelOptionConfiguration: (ServerBootstrap) -> (ServerBootstrap)
+        
         public init(
             hostname: String = Self.defaultHostname,
             port: Int = Self.defaultPort,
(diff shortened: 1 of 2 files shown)


Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -202,7 +212,9 @@ public final class HTTPServer: Server {
             serverName: String? = nil,
             reportMetrics: Bool = true,
             logger: Logger? = nil,
-            shutdownTimeout: TimeAmount = .seconds(10)
+            shutdownTimeout: TimeAmount = .seconds(10),
+            serverChannelOptionConfiguration: @escaping (ServerBootstrap) -> ServerBootstrap = { bootstrap in return bootstrap },
```
Joannis, line 216: I really don't like adding this niche use case to the initializer

Joannis (commented):
I think it's a good addition, but I dislike the way it clutters the initializer for such a niche use case.

John-Connolly:
> I think it's a good addition, but I dislike the way it clutters the initializer for such a niche use case.

@Joannis Thanks for the feedback. I tend to agree, 90 % of vapor users will not have a need to use this option. Thoughts on this being more akin to the ```customCertificateVerifyCallback``` that way the initializer doesn't need to change.

PR #2920, opened 2022-11-25
FileMiddleware: Add convenience support for serving files from a Bundle

MrSkwiggs:
Adds support for instantiating a `FileMiddleware` to serve files from a Bundle. This is necessary when Vapor is used as a dependency in an `Xcodeproj`

### DIFF
--- a/Sources/Vapor/Middleware/FileMiddleware.swift
+++ b/Sources/Vapor/Middleware/FileMiddleware.swift
@@ -16,6 +16,30 @@ public final class FileMiddleware: Middleware {
         self.publicDirectory = publicDirectory.addTrailingSlash()
         self.defaultFile = defaultFile
     }
+    
+    #if canImport(Foundation)
+    /// Creates a new `FileMiddleware` for a server contained in an Xcode Project.
+    ///
+    /// - parameters:
+    ///     - bundle: The Bundle which contains the files to serve.
+    ///     - publicDirectory: The public directory to serve files from.
+    ///     - defaultFile: The name of the default file to look for and serve if a request hits any public directory. Starting with `/` implies
+    ///     an absolute path from the public directory root. If `nil`, no default files are served.
+    ///
+    /// - important: Make sure the root directory you wish to serve files from is included in the `Copy Bundle Resources` build phase of your project
+    public init?(bundle: Bundle = .main, publicDirectory: String = "Public", defaultFile: String? = nil) {
+        let fullPublicDirectoryPath: String
+        if #available(macOS 13.0, *) {
+            guard let bundleResourceURL = bundle.resourceURL?.appending(path: publicDirectory) else { return nil }
+            fullPublicDirectoryPath = bundleResourceURL.path()
+        } else {
+            guard let bundleResourceURL = bundle.resourceURL?.appendingPathComponent(publicDirectory) else { return nil }
+            fullPublicDirectoryPath = bundleResourceURL.path
+        }
+        self.publicDirectory = fullPublicDirectoryPath.addTrailingSlash()
+        self.defaultFile = defaultFile
+    }
+    #endif
 
     public func respond(to request: Request, chainingTo next: Responder) -> EventLoopFuture<Response> {
         // make a copy of the percent-decoded path


Joannis:
Any idea why the tests aren't running?

MrSkwiggs:
@Joannis It says `First-time contributors need a maintainer to approve running workflows`

vzsg:
The changes are not building on macOS 11:

```
/__w/vapor/vapor/vapor/Sources/Vapor/Middleware/FileMiddleware.swift:33:63: error: value of type 'URL' has no member 'appending'

            guard let bundleResourceURL = bundle.resourceURL?.appending(path: publicDirectory) else { return nil }
                                          ~~~~~~~~~~~~~~~~~~~ ^~~~~~~~~
```

That method and syntax is only available on macOS 13, and I don't think Linux supports it either. The `appendingPathComponent(_:)` method should work instead.

vzsg:
That's assuming your resource bundles are somehow preserving internal dictionaries. I haven't seen them do that before (but it might be my lack of reading on the subject).

MrSkwiggs:
@vzsg Thanks for your review & comments 🙌

> That method and syntax is only available on macOS 13 (according to the docs, though the tests suggest that it's available from 12), and I don't think Linux supports it either. The appendingPathComponent(_:) method should work instead.

I have actually used both and explicitly check for `@available(macOS 13.0` where appropriate 🤔 I could also just simply use the older `appendingPathComponent(_:)` and not try for more recent APIs.

> That's assuming your resource bundles are somehow preserving internal dictionaries. I haven't seen them do that before (but it might be my lack of reading on the subject).

I'm not sure I understand this comment? This extension aims to provide convenience for serving actual files that are copied over into an app's `Bundle`

MrSkwiggs:
@Joannis @vzsg I've removed references to unavailable APIs, could I ask for another workflow run? 🙏

vzsg:
> I'm not sure I understand this comment? This extension aims to provide convenience for serving actual files that are copied over into an app's Bundle

As I said in the parentheses, it was my misunderstanding. I'm so used to adding files directly to bundles, that I didn't realize that folder references can be used too – and then the "Public" folder would be preserved.

Godspeed.

vzsg:
TBH, a test would be nice to see it working too.

MrSkwiggs:
@vzsg 
> TBH, a test would be nice to see it working too.

I'll try to get that going, but since it requires an actual `Bundle`, I'm not sure how I can set that up for tests from a Swift Package 🤔

MrSkwiggs:
@vzsg Turns out, it's super straightforward to use Bundle, have added 2 tests 👌

MrSkwiggs:
@vzsg @Joannis I think this PR is ready for final review/workflow checks, could I ask for giving it a spin? 🙏

MrSkwiggs:
@vzsg @Joannis @0xTim Sorry for the bump 🙇  It seems that all tests passed, so this is most likely ready for final review by one of you's 🙏

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -64,6 +64,29 @@ public final class FileMiddleware: Middleware {
...
+    ///
+    /// - parameters:
+    ///     - bundle: The Bundle which contains the files to serve.
+    ///     - publicDirectory: The public directory to serve files from.
+    ///     - defaultFile: The name of the default file to look for and serve if a request hits any public directory. Starting with `/` implies
+    ///     an absolute path from the public directory root. If `nil`, no default files are served.
+    ///
+    /// - important: Make sure the public directory you wish to serve files from is included in the `Copy Bundle Resources` build phase of your project
+    /// - returns: A fully qualified FileMiddleware if the given `publicDirectory` can be served, `nil` otherwise
+    public convenience init?(bundle: Bundle, publicDirectory: String = "Public", defaultFile: String? = nil) {
```
Joannis, line 79: I don't like having a `nillable` API here, as providing the wrong information is an error for (almost) all use cases and would prevent further setup of the app.

I'm also wondering why we need `#if canImport(Foundation)` since it's virtually always present.

Joannis (changes requested):

Sources/Vapor/Middleware/FileMiddleware.swift
MrSkwiggs: Thanks for the review @Joannis 🙌 

> I don't like having a nillable API here, as providing the wrong information is an error for (almost) all use cases and would prevent further setup of the app.

Do you mean, I should use `init(_:) throws` instead? I could do so but then I'd need to create a `FileMiddlewareError` type. I initially did not go that route however, as it also did not follow the existing `FileMiddleware.init(_:)` pattern already in use.

Or do you mean, this should just _not_ be `nil-able` and always succeed, even if the given `publicDirectory` or `Bundle` do not resolve to anything?

If you mean the latter, as a user of this interface, I found this to be somewhat confusing at first when things seemingly compiled and ran just fine, but my files would never actually be served 😕
And in this case, would there be a compromise or a different way to surface a setup issue to the user that you would be more comfortable with?

> I'm also wondering why we need #if canImport(Foundation) since it's virtually always present.
Just wanted to be cautious with imports, as it was not declared at file level and I did not want to assume `Foundation` was available for all targets. 
I'll see if adding this import de-facto breaks unit tests and report back 👌

Joannis: I intended to suggest using `init(_:) throws`, though your alternative could also work. I think I'd prefer a throwing initializer still, because it is indeed confusing.

MrSkwiggs: Gotcha, will do!

MrSkwiggs: @Joannis I've now refactored this init to `throw`, which required me to create a new nested Error type for `FileMiddleware` -> `SetupError`.

I've edited the unit tests to cover for these changes 👌 
Additionally, I've removed the `#if canImport(Foundation)` directive.

Could I ask for another review? 🙏

MrSkwiggs: Hey @Joannis, sorry for the bump, but would like to merge this before the holidays if possible 🙏  Could you give it a final review?

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -5,6 +5,13 @@ public final class FileMiddleware: Middleware {
     /// The public directory. Guaranteed to end with a slash.
     private let publicDirectory: String
     private let defaultFile: String?
+    
+    public enum SetupError: Error {
+        /// Cannot generate Bundle Resource URL
+        case bundleResourceURLIsNil
+        /// Cannot find any actual folder for the given Public Directory
+        case publicDirectoryIsNotAFolder
```
Joannis, line 13: Since this error is a `public enum`, I'd like to make sure this fits with the rest of Vapor's error cases and conventions. We also need to make sure that this enum does not get expanded in the future. And @MrSkwiggs you can probably help theorise what other cases might get added in future PRs.

A way to circumvent this is by making it a `public struct SetupError: Equatable, Error {}`. With that pattern, you can add some static lets like so:

```swift
public static let missingBundleResourceUrl = SetupError(description: "...")
```

That way you can add new cases that users can catch _without_ breaking API. Look at some of SwiftNIO's errors as an example.

@0xTim I'd like you to weigh in.

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -5,6 +5,13 @@ public final class FileMiddleware: Middleware {
     /// The public directory. Guaranteed to end with a slash.
     private let publicDirectory: String
     private let defaultFile: String?
+    
+    public enum SetupError: Error {
```
Joannis, line 9: I also think that `SetupError` should be restricted/renamed to being a `BundleSetupError`. Otherwise, I expect other additions to FileMiddleware to make use of this error enum

Joannis (commented):
Code looks good to me! Just need to verify the error case - unless it becomes a struct.

vzsg:
BTW the tests are failing on Linux because Bundle doesn't seem to have a parameterless constructor there.

Sources/Vapor/Middleware/FileMiddleware.swift
MrSkwiggs: Fair point, will do!

Sources/Vapor/Middleware/FileMiddleware.swift
MrSkwiggs: Ah yes sounds good, will do!

0xTim: Yes we can have an internal enum to model it but the public error should be a static var

MrSkwiggs:
> BTW the tests are failing on Linux because Bundle doesn't seem to have a parameterless constructor there.

@vzsg There's not much I can do other than remove that specific test, then :/ There's no other way to generate a Bundle that does not have a resource URL otherwise (which is what is being tested for the FileMiddleware).

Sources/Vapor/Middleware/FileMiddleware.swift
MrSkwiggs: I've replaced this with a public struct instead, and added static lets as suggested 👌

Sources/Vapor/Middleware/FileMiddleware.swift
MrSkwiggs: Done

Joannis (approved):

MrSkwiggs:
@Joannis Could I ask you to run the workflow, and perhaps MWPS? 🙏

Joannis:
Done @MrSkwiggs

MrSkwiggs:
@Joannis it seems like one [unit test](https://github.com/vapor/vapor/actions/runs/3874251630/jobs/6605251407) is timing out 🤔 Not sure why

MrSkwiggs:
@vzsg @0xTim @Joannis Sorry for the bump 🙇‍♂️ One unit test timed out and I'm not sure why. Could I ask for re-running [this particular test](https://github.com/vapor/vapor/actions/runs/3874251630/jobs/6605251407)? 🙏

MrSkwiggs:
@vzsg @0xTim @Joannis Tests succeeded, ready for merge 🙌

Issue #2921, opened 2022-11-25
Special characters in URL lead to empty response from server 


fred-sch:
### Describe the bug

Invalid URLs containing not %-encoded special characters lead to an empty response and a hang when shutting down the application.

### To Reproduce

1. run `vapor new helloworld-test`
2. `cd helloworld-test`
3. `swift run --log trace`
4. in another terminal execute: `curl "http://127.0.0.1:8080/°" --http1.1 --verbose` 

This will give the following behavior in the curl query:
```bash
⚡55% ✖  curl "http://127.0.0.1:8080/°" --http1.1 --verbose
*   Trying 127.0.0.1:8080...
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> GET /° HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.78.0
> Accept: */*
>
* Empty reply from server
* Closing connection 0
curl: (52) Empty reply from server
```
and the following in the trace log
```bash
⚡53% ➜ .build/debug/Run --log trace
[ codes.vapor.application ] [ DEBUG ] Could not load .env.development file: open(file:oFlag:mode:): No such file or directory (errno: 2) (Vapor/DotEnv.swift:120)
[ codes.vapor.application ] [ TRACE ] Running on shared EventLoopGroup. Not shutting down EventLoopGroup. (Vapor/DotEnv.swift:106)
[ codes.vapor.application ] [ DEBUG ] Could not load .env file: open(file:oFlag:mode:): No such file or directory (errno: 2) (Vapor/DotEnv.swift:120)
[ codes.vapor.application ] [ TRACE ] Running on shared EventLoopGroup. Not shutting down EventLoopGroup. (Vapor/DotEnv.swift:106)
[ codes.vapor.application ] [ TRACE ] Running on shared EventLoopGroup. Not shutting down EventLoopGroup. (Vapor/DotEnv.swift:59)
[ codes.vapor.application ] [ NOTICE ] Server starting on http://127.0.0.1:8080 (Vapor/HTTPServer.swift:296)
[ codes.vapor.application ] [ DEBUG ] Unhandled HTTP server error: invalid URL (Vapor/HTTPServer.swift:459)
[ codes.vapor.application ] [ TRACE ] Unhandled user event: outputClosed (Vapor/HTTPServerRequestDecoder.swift:217)
^C
[ codes.vapor.application ] [ DEBUG ] Application shutting down (Vapor/Application.swift:136)
[...]

PR #2922, opened 2022-11-29
HTTPServerErrorHandler: Improve Error Handling for HTTPParserError

fred-sch:
Invalid HTTP previously just closed the connection, but did not actually handle the errors. 

This PR uses an adapted version of https://github.com/apple/swift-nio/blob/main/Sources/NIOHTTP1/HTTPServerProtocolErrorHandler.swift to properly catch exceptions if they occur.

Closes: https://github.com/vapor/vapor/issues/2921


Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -447,17 +447,47 @@ private final class HTTPServerConnection {
...
-        context.close(mode: .output, promise: nil)
+        guard error is HTTPParserError else {
+            context.fireErrorCaught(error)
+            return
+        }
+
+        // Any HTTPParserError is automatically fatal, and we don't actually need (or want) to
+        // provide that error to the client: we just want to tell it that it screwed up and then
+        // let the rest of the pipeline shut the door in its face. However, we can only send an
+        // HTTP error response if another response hasn't started yet.
```
MrSkwiggs, line 478: I'm not sure we want to use this tone of voice in code comments / docs
```suggestion
        // Any HTTPParserError is automatically fatal, and we don't actually need (or want) to
        // provide that error to the client: we just want to inform them something went wrong and then
        // let the rest of the pipeline close. However, we can only send an
        // HTTP error response if another response hasn't started yet.
```

fred-sch: @MrSkwiggs thanks for having a look at the PR.

I actually took this comment directly from https://github.com/apple/swift-nio/blob/00341c92770e0a7bebdc5fda783f08765eb3ff56/Sources/NIOHTTP1/HTTPServerProtocolErrorHandler.swift#L41, but I think it is perfectly fine to adapt to Vapors code/documentation style.

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -447,17 +447,47 @@ private final class HTTPServerConnection {
...
     
     init(logger: Logger) {
         self.logger = logger
     }
     
     func errorCaught(context: ChannelHandlerContext, error: Error) {
-        self.logger.debug("Unhandled HTTP server error: \(error)")
-        context.close(mode: .output, promise: nil)
+        guard error is HTTPParserError else {
+            context.fireErrorCaught(error)
```
fred-sch, line 471: @MrSkwiggs do you agree that `context.fireErrorCaught(error)` should always be called (to propagate the error through the ChannelPipeline) instead of just closing it with `context.close(mode: .output, promise: nil)`?

Also I was thinking that a line for the logger would be nice in this guard statement:
```swift
self.logger.debug("Catching HTTP server error: \(error)")
```

Sources/Vapor/HTTP/Server/HTTPServer.swift
MrSkwiggs: Huh interesting 😅

Sources/Vapor/HTTP/Server/HTTPServer.swift
MrSkwiggs: I can't speak for Vapor as I was just merely glancing at your PR and am not a maintainer here; but I would think so yes, though take whatever I say with a healthy spoonful of salt :d

fred-sch:
Could you have a look at the proposed changes and trigger the workflow when you have some time @0xTim ?

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -447,17 +447,60 @@ private final class HTTPServerConnection {
...
+            context.close(mode: .output, promise: nil)
+            return
+            // TODO: why can we not make this: ?
+            // otherwise the `PipelineTests.testBadStreamLength` fails
+            // context.fireErrorCaught(error)
+            // return
+        }
+
+        // TODO: why do we need a special case for this this??
+        // Otherwise the "general fix" would brake the `testRequestBodyStreamGetsFinalisedEvenIfClientDisappears`
```
Joannis, line 480: brake -> break?

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -447,17 +447,60 @@ private final class HTTPServerConnection {
...
+/// servers want. This handler does not suppress the parser errors: it allows them to
+/// continue to pass through the pipeline so that other handlers (e.g. logging ones) can
+/// deal with the error.
+/// 
+/// adapted from: https://github.com/apple/swift-nio/blob/00341c92770e0a7bebdc5fda783f08765eb3ff56/Sources/NIOHTTP1/HTTPServerProtocolErrorHandler.swift
 final class HTTPServerErrorHandler: ChannelInboundHandler {
     typealias InboundIn = Never
+    typealias OutboundOut = Response
     let logger: Logger
+    private var hasUnterminatedResponse: Bool = false
```
Joannis, line 463: Is this being set anywhere?

fred-sch: You are absolutely write this was not being set anywhere. It is now correctly being set in `func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?)`.

fred-sch:
I had another look at the code and figured out that I was most likely not adding the `HTTPServerErrorHandler` at the right place between the different `ChannelHandler`s (which lead to some problems).

Now I put the `HTTPServerErrorHandler` always before the  `HTTPServerResponseEncoder`. Also, I added the `NIOCloseOnErrorHandler` at the very end of the `ChannelHandlers`. This one should now take care of the role the `HTTPServerErrorHandler` had previously (closing connections) while the `HTTPServerErrorHandler` takes care of `HTTPParserError`s.

Can one of you possibly take another look at the changes @Joannis @0xTim ?

fred-sch:
For the context again: If a vapor app is supporting HTTP without TLS/SSL and is running behind nginx proxy, a single faulty (invalid HTTP) request can easily bring down all upstream servers  for a time amount of the upstreams `fail_timeout` if the app is under some load, because there is also a bug in nginx that it will ignore the `proxy_next_upstream_tries` directive if under load (compare: https://trac.nginx.org/nginx/ticket/2421). 

Another side note: we have been using this fix on one of our vapor based apis (approx. 20 Million req/day) for over a month now and have not experienced any problems so far. 

Any thoughts or suggestions on this MR are very appreciated!

0xTim (commented):
This looks good to me - @Lukasa @weissi are you able to give this a quick once over to ensure the handler stuff is being used correctly?

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -449,17 +449,77 @@ private final class HTTPServerConnection {
...
+        // provide that error to the client: we just want to tell it that it screwed up and then
+        // let the rest of the pipeline shut the door in its face. However, we can only send an
+        // HTTP error response if another response hasn't started yet.
+        //
+        // A side note here: we cannot block or do any delayed work. ByteToMessageDecoder is going
+        // to come along and close the channel right after we return from this function.
+        if !self.hasUnterminatedResponse {
+            self.logger.debug("Bad Request - Invalid HTTP: \(error)")
+            let headers = HTTPHeaders([("Connection", "close"), ("Content-Length", "0")])
+            let head = HTTPResponseHead(version: .http1_1, status: .badRequest, headers: headers)
```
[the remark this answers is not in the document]
fred-sch: Maybe the easiest way to deal with this would be to have a separate `HTTP1ServerErrorHandler`and `HTTP2ServerErrorHandler` that are added in `addVaporHTTP1Handlers` and `addVaporHTTP2Handlers` respectively?

As far as I understand getting the HTTP version from the current request will not be easily possible because the error is thrown in the `feedInput` method of the `BetterHTTPParser`  when the `HTTPRequestDecoder` is used to transform the InboundIn `ByteBuffer` to the InboundOut `HTTPServerRequestPart` (in the Http1 pipeline). So we never actually get to see the `HTTPServerRequestPart` when an error is thrown.

I am also wondering if this is even possibly happening in the HTTP/2 pipeline, because the `HTTP2FramePayloadToHTTP1ServerCodec` transforms a `HTTP2Frame.FramePayload` to a `HTTPServerRequestPart`. The only errors that are being thrown on the way should be `NIOHTTP2Errors`, so maybe a separate handler would be needed to handle such errors as well?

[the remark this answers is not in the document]
0xTim: Yeah that makes sense. @FranzBusch I'd like to ask your thoughts here as the resident NIO expert! And if you have chance for a once-over in general as well that would be ace. This has been running fine in a prod system so I'm happy with the overall stability but just a sanity check to make sure we're not doing anything silly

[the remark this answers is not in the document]
FranzBusch: This PR is by itself fine and writing out a `badRequest` in response to seeing an `HTTPParserError` is good. As you noted above though you also need to handle the `HTTP2Error`s somewhere. Either in a separate handler or by modifying this one.

Sources/Vapor/HTTP/Server/HTTPServer.swift
Lukasa: What handler is going to deal with these errors?

fred-sch: Currently the `NIOCloseOnErrorHandler` will eventually close the `Channel`

Sources/Vapor/HTTP/Server/HTTPServer.swift
gwynne: While I may be a bit late to the table, NIO is never shy about accepting improvements of this sort, and Vapor isn't either; the suggested rephrasing is a definite improvement.

gwynne (changes requested):
My only comment here is the one where I resurrected someone else's old comment 😅 But other than that, LGTM

Sources/Vapor/HTTP/Server/HTTPServer.swift
fred-sch: I adjusted the comment as suggested

Sources/Vapor/HTTP/Server/HTTPServer.swift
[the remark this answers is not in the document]
fred-sch: Any opinions on my above comment @0xTim @FranzBusch ?

gwynne:
@fred-sch Slightly off-topic, is it intentional that the Vapor CI workflows are being run in your fork? (Such as the API docs CI, although it fails immediately since the credentials secrets obviously aren't available.)

gwynne (approved):
This LGTM at this point, though I wouldn't mind at least one more set of eyes on it.

0xTim (approved):
I think we're good to merge!

PR #2928, opened 2022-12-27
Add mocked client + subclass of XCTestCase

Kithin:
I have created a spy client, in order to stub the response and also being able to verify the request.

To reduce boilerplate code, I created a XCTVaporTestCase. You don't need to configure the application every time. And tearing it down. All is being handled by XCTVaporTestCase.

You just simply do this:

```
final class ExampleTests: XCTVaporTestCase {
     func test_example() throws {
           try client.stubResponse(httpStatus: .ok)
     }
}
```


Sources/XCTVapor/XCTVaporTestCase.swift
```diff
@@ -0,0 +1,32 @@
+import XCTest
```
0xTim, line 1: Let's remove this - we can just add how to use in the docs since most people using it will want to customise it

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
```diff
@@ -0,0 +1,53 @@
+public final class SpyClient: Client {
+    private(set) var requestUrlComponentsUsed: URLComponents = .init()
+    private(set) var requestHttpMethodUsed: HTTPMethod?
+    private(set) var requestHeadersUsed: HTTPHeaders?
+    private(set) var requestBodyUsed: Data?
+    
+    private var stubResponse: ClientResponse!
```
0xTim, line 7: Let's default to an empty 200 OK instead of using an IUO

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
```diff
@@ -0,0 +1,53 @@
...
+    // Conforming to the protocol: Client
+    public func delegating(to eventLoop: EventLoop) -> Client {
+        self.eventLoop = eventLoop
+        
+        return self
+    }
+    
+    // Conforming to the protocol: Client
+    public func send(_ request: ClientRequest) -> EventLoopFuture<ClientResponse> {
+        self.requestUrlComponentsUsed.scheme = request.url.scheme
```
0xTim, line 26: We should definitely store the requests directly in to an array

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
```diff
@@ -0,0 +1,53 @@
+public final class SpyClient: Client {
```
0xTim, line 1: We're likely definitely missing an import on Foundation here

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
```diff
@@ -0,0 +1,53 @@
...
+    private(set) var requestUrlComponentsUsed: URLComponents = .init()
```
0xTim, line 2: I'd like to avoid URLComponents if possible, I don't see any issue just storing the request directly

0xTim (changes requested):
Initial comments - it would also be good to add this to a test to a) make sure it works and b) show how to use it

Sources/XCTVapor/XCTVaporTestCase.swift
Kithin: Just curious, for understanding and learning, why do we want to remove this import XCTest?

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
Kithin: Why do we want to store the request in an array? 

Is it because in case there are multiple request being made. And with an array you know how many request in total was made. And also being able to select a specific request (e.g. the third request).

Is this the reason?

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
Kithin: Good one. I will add this import

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
Kithin: The reason why I decided to use URLComponents was because in situations where there is also a query. With: `request.url.query`, you get back a String. And the problem with it, is that the order of key, value pairs are random.
Meaning the first time you get it like this: `key_1=value_1&key_2=value_2`
and the second time maybe like this: `key_2=value_2&key_1=value_1`
So it becomes impossible to make assertions in your test.

And with: `requestUrlComponentsUsed.query`, you get back an array. So it's easier to make assertions in your tests. By looping over it.

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
Kithin: What does IUO mean?

Kithin: Why do you want to have a default stubResponse of status code 200?

Just curios

Kithin:
@0xTim can you maybe help me, with this compiling error: 
`Undefined symbol: _$s11Development9configureyy5Vapor11ApplicationCKF`

In the sub class I created: XCTVaporTestCase, I have added this code on line 20: 
`try configure(app)`

After that I got this error:
`Cannot find 'configure' in scope`

That's why I added this import: 
`import Development`

But then I got this symbol error:
`Undefined symbol: _$s11Development9configureyy5Vapor11ApplicationCKF`

0xTim:
@Kithin that's because you're trying to use something that links XCTest in an executable which isn't possible. If you use it in a test similar to https://github.com/vapor/vapor/blob/main/Tests/VaporTests/ApplicationTests.swift#L104 (that file is probably a good place for it) it should work

Sources/XCTVapor/XCTVaporTestCase.swift
0xTim: Sorry I meant remove the whole file - those using the fake client will likely have much more complicated set ups that make this class redundant

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
0xTim: This avoids it crashing in case they don't set a response and makes it easier to use in tests where they may not actually test the client stuff but still use it to avoid outbound requests in tests

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
0xTim: Yep exactly, you might have a route that makes 5 requests and you want to assert the order and contents of each request for example

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
0xTim: There's nothing stopping anyone using it from doing this themselves. This reason I want to about `URLComponents` is that adds a dependency on `FoundationNetworking` which we want to avoid on Linux if possible

Kithin:
Aaaah, so you mean, I should move this file in the test folder instead of the source folder?

But is it still accessible for people to use in their own tests?

Kithin:
That something is that the SpyClient I created or because that class is a subclass of XCTestCase, and this super class can only be executed in test?

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
0xTim: (IUO is implicitly unwrapped optional)

0xTim:
The `SpyClient` should stay in XCTVapor but where you try to use it should be in a test case

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
Kithin: Could you explain, why. Just curious and to learn.

Kithin:
> @Kithin that's because you're trying to use something that links XCTest in an executable which isn't possible. If you use it in a test similar to https://github.com/vapor/vapor/blob/main/Tests/VaporTests/ApplicationTests.swift#L104 (that file is probably a good place for it) it should work

I have tried it, but then I get the same error as in my previous comment:

```
@0xTim can you maybe help me, with this compiling error:
Undefined symbol: _$s11Development9configureyy5Vapor11ApplicationCKF

In the sub class I created: XCTVaporTestCase, I have added this code on line 20:
try configure(app)

After that I got this error:
Cannot find 'configure' in scope

That's why I added this import:
import Development

But then I got this symbol error:
Undefined symbol: _$s11Development9configureyy5Vapor11ApplicationCKF
```

Kithin:
> @Kithin that's because you're trying to use something that links XCTest in an executable which isn't possible. If you use it in a test similar to https://github.com/vapor/vapor/blob/main/Tests/VaporTests/ApplicationTests.swift#L104 (that file is probably a good place for it) it should work

I have tried it, but then I get the same error as in my previous comment:

```
@0xTim can you maybe help me, with this compiling error:
Undefined symbol: _$s11Development9configureyy5Vapor11ApplicationCKF

In the sub class I created: XCTVaporTestCase, I have added this code on line 20:
try configure(app)

After that I got this error:
Cannot find 'configure' in scope

That's why I added this import:
import Development

But then I got this symbol error:
Undefined symbol: _$s11Development9configureyy5Vapor11ApplicationCKF
```

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
0xTim: FoundationNetworking means that you need curl which isn't available by default on many systems and can cause issues with linking. So we avoid using it. This also ties nicely in with the upcoming Foundation changes to split things up and best practices for us to make our library as small as possible

Kithin:
> Can you push the current test case up?

@0xTim I have pushed it. It's the latest commit: UPDATE - set mocked client

Tests/VaporTests/ApplicationTests.swift
```diff
@@ -104,6 +104,16 @@ final class ApplicationTests: XCTestCase {
...
+        
+        let client = SpyClient(
+            eventLoop: app.eventLoopGroup.next()
+        )
+        
+        app.clients.use { _ in
+            return client
+        }
+        
+        try configure(app)
```
0xTim, line 116: The tests are currently failing because this function doesn't exist - you can just remove this line for the tests
```suggestion
        
```

Kithin: Does this not mean the mocked client will not be used by the application?

familon16:
hi


[image: Mailtrack]
<https://mailtrack.io?utm_source=gmail&utm_medium=signature&utm_campaign=signaturevirality11&>
Sender
notified by
Mailtrack
<https://mailtrack.io?utm_source=gmail&utm_medium=signature&utm_campaign=signaturevirality11&>
01/12/23,
12:56:50 AM

Tests/VaporTests/ApplicationTests.swift
Kithin: I have removed the line of code, thank you for your help

Kithin:
@0xTim the test keeps failing on GitHub actions, but not locally. Can you perhaps help me?

0xTim:
@Kithin you're using the new generics syntax from Swift 5.7 that isn't available on older Swift versions. You'll need to find a work around

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+public final class SpyClient: Client {
+    private(set) var requestsUsed: [ClientRequest] = []
+    
+    private var stubResponse: ClientResponse = .init(status: .ok)
```
0xTim, line 6: This should be public to make it completely customisable

Kithin: Do you mean stubResponse?

Is the stubResponse() method not enough? 

Could you explain, trying to learn geen your thoughts.

0xTim: Yeah `stubResponse`. The function is good for most people to allow them to set most common cases but it's impossible to predict what people might want to try and test (custom headers, missing bodies, empty bodies etc) so we should let them set the response wholesale if needs be

Kithin:
@0xTim do you have more feedback. And I also wanted to thank you for your help, when I was stuck.

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+public final class SpyClient: Client {
+    private(set) var requestsUsed: [ClientRequest] = []
```
0xTim, line 4:
```suggestion
    private(set) var requestsSeen: [ClientRequest] = []
```

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
```diff
@@ -0,0 +1,42 @@
...
+    // Conforming to the protocol: Client
+    public func delegating(to eventLoop: EventLoop) -> Client {
+        self.eventLoop = eventLoop
+        
+        return self
+    }
+    
+    // Conforming to the protocol: Client
+    public func send(_ request: ClientRequest) -> EventLoopFuture<ClientResponse> {
+        self.requestsUsed.append(request)
```
0xTim, line 25:
```suggestion
        self.requestsSeen.append(request)
```

Sources/XCTVapor/TestDoubles/Spies/SpyClient.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
```
0xTim, line 1:
```suggestion
import Foundation
import Vapor
import NIOCore
```

0xTim (changes requested):
Sorry for the delay - last few changes and I think we're good to go!

Kithin:
@0xTim When I merge main into this feature branch, I don't get that merge conflict. But I have fixed the merge conflicts anyway. But the PR I still get the message: This branch has conflicts that must be resolved

0xTim:
@Kithin It's in the application tests, I can't see the test you added anymore?

Kithin:
> @Kithin It's in the application tests, I can't see the test you added anymore?

I didn't added a new test, I updated this test: testBoilerplate

0xTim:
@Kithin Hmm I see. I can't check the merge conflicts because I don't have access to the fork. So the easiest thing to probably do is close the PR, update your fork, then create a new branch and copy the code across. Let's put the test for the client in its own test case as well

SpaceboatDVLP:
Would it make sense to add a `configure` function within SpyClient in case someone wants to init a stubbed response within a unit test rather than setting up the client within the XCTestCase? For example:

`static func configure<T: Content>(on app: Application, with response: T) throws {
        let client = SpyClient(eventLoop: app.eventLoopGroup.next())
        app.clients.use { _ in
            return client
        }
        try client.stubResponse(httpStatus: .ok, responseData: response)
    }`
    
Then you could use it within a unit test like this:

`let mockResponse = MockResponse(message: "nice")
    try SpyClient.configure(on: app, with: mockResponse)`

Thanks for this by the way.

0xTim:
@SpaceboatDVLP that pattern would be to store the instance of the `SpyClient` on the `XCTestCase`

So in your test case you'd do
```swift
self.spyClient.stubResponse(httpStatus: .ok, responseData: response)
```
That allows you to also inspect the request send to the client etc

There's nothing stopping you adding the configure function in your code of course

AFutureD:
Support stub to a specific request?

I have a [demo](https://github.com/AFutureD/SignInWithAppleVapor/blob/main/Tests/SignInWithAppleVaporTests/Utils/MockClient.swift)
And an [example](https://github.com/AFutureD/SignInWithAppleVapor/blob/3327be40adc5a0c139c7546b4ebc0abdd934e495/Tests/SignInWithAppleVaporTests/AppleApiTests.swift#L82) to use it.

Naveen-C-Ramachandrappa (approved):

PR #2934, opened 2023-01-10
Add 308 Redirect

NeedleInAJayStack:
Adds support for a `308 redirect` to support POST method redirection. This is important because not all user-agents preserve the method and body through a 301 redirect: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301

`redirect(to:type:)` is now deprecated in favour of `redirect(to:redirectType:)`

### DIFF
--- a/Sources/Vapor/Error/Abort.swift
+++ b/Sources/Vapor/Error/Abort.swift
@@ -12,6 +12,10 @@ public struct Abort: AbortError, DebuggableError {
     ///
     /// Set type to '.permanently' to allow caching to automatically redirect from browsers.
     /// Defaulting to non-permanent to prevent unexpected caching.
+    /// - Parameters:
+    ///   - location: The path to redirect to
+    ///   - type: The type of redirect to perform
+    /// - Returns: An abort error that provides a redirect to the specified location
     @available(*, deprecated, renamed: "redirect(to:redirectType:)")
     public static func redirect(to location: String, type: RedirectType) -> Abort {
         var headers: HTTPHeaders = [:]
@@ -25,6 +29,10 @@ public struct Abort: AbortError, DebuggableError {
     ///
     /// Set type to '.permanently' to allow caching to automatically redirect from browsers.
     /// Defaulting to non-permanent to prevent unexpected caching.
+    /// - Parameters:
+    ///   - location: The path to redirect to
+    ///   - redirectType: The type of redirect to perform
+    /// - Returns: An abort error that provides a redirect to the specified location
     public static func redirect(to location: String, redirectType: Redirect = .normal) -> Abort {
         var headers: HTTPHeaders = [:]
         headers.replaceOrAdd(name: .location, value: location)
--- a/Sources/Vapor/Error/Abort.swift
+++ b/Sources/Vapor/Error/Abort.swift
@@ -12,8 +12,8 @@ public struct Abort: AbortError, DebuggableError {
     ///
     /// Set type to '.permanently' to allow caching to automatically redirect from browsers.
     /// Defaulting to non-permanent to prevent unexpected caching.
-    @available(*, deprecated, renamed: "redirectTo")
-    public static func redirect(to location: String, type: RedirectType = .normal) -> Abort {
+    @available(*, deprecated, renamed: "redirect(to:redirectType:)")
+    public static func redirect(to location: String, type: RedirectType) -> Abort {
         var headers: HTTPHeaders = [:]
         headers.replaceOrAdd(name: .location, value: location)
         return .init(type.status, headers: headers)
@@ -21,14 +21,14 @@ public struct Abort: AbortError, DebuggableError {
     
     /// Creates a redirecting `Abort` error.
     ///
-    ///     throw Abort.redirectTo("https://vapor.codes")"
+    ///     throw Abort.redirect(to: "https://vapor.codes")
     ///
     /// Set type to '.permanently' to allow caching to automatically redirect from browsers.
     /// Defaulting to non-permanent to prevent unexpected caching.
-    public static func redirectTo(_ location: String, type: Redirect = .normal) -> Abort {
+    public static func redirect(to location: String, redirectType: Redirect = .normal) -> Abort {
         var headers: HTTPHeaders = [:]
         headers.replaceOrAdd(name: .location, value: location)
-        return .init(type.status, headers: headers)
+        return .init(redirectType.status, headers: headers)
     }
 
     /// See `Debuggable`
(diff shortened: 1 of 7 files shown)


NeedleInAJayStack:
The failing test indicates that this changes the public API. That's true - it adds an additional case to the public `RedirectType` enum.

0xTim:
Yeah this is going to be one we can't accept for now as it's a breaking change. We'll need to find another way around it, like adding a new enum and function and deprecate the old stuff

NeedleInAJayStack:
@0xTim Totally fair point. I've adjusted this so that it introduces a new `Redirect` enum and a `redirectTo` function, and deprecates the existing `RedirectType` enum and `redirect` functions.

Joannis:
Assuming we want to implement this change, simply replacing an enum (because of breaking changes) with a new enum is only going to expand our problems with the API surface. The only real option in my eyes is to make it a `struct` with a `private enum` of some sorts. That way we can expand the API without breaking the API surface of the enum.

NeedleInAJayStack:
@Joannis I like your suggestion. I've adjusted the implementation to use a `struct` that wraps an `enum`.

Tests/VaporTests/ClientTests.swift
```diff
@@ -9,7 +9,7 @@ final class ClientTests: XCTestCase {
         app.http.client.configuration.redirectConfiguration = .disallow
 
         app.get("redirect") {
-            $0.redirect(to: "foo")
```
Joannis, line 12: We can't just change the tests to use the new method. We should be testing the 'old' method in addition to the new one.

Joannis (changes requested):

Tests/VaporTests/ClientTests.swift
Joannis: There are 2 other tests that also need to be reverted/duplicated for the new method.

NeedleInAJayStack: As far as I can tell, these tests are not really testing the functionality of the redirect itself. Rather, they are testing that changes to a Client configuration have the expected impact based on where the changes occur. The redirect is simply an easy way to test these configuration changes. I definitely can create duplicate test functions, but it seems a little strange to do so for a test that is not really directly testing the redirect behavior.

This does surface that we were not directly testing the `Request.redirect` functionality, so I added those tests for the new and deprecated versions.

Joannis (approved):
LGTM, awaiting @0xTim 's review

Joannis:
Ping @0xTim

Sources/Vapor/Error/Abort.swift
```diff
@@ -12,11 +12,24 @@ public struct Abort: AbortError, DebuggableError {
...
         return .init(type.status, headers: headers)
     }
+    
+    /// Creates a redirecting `Abort` error.
+    ///
+    ///     throw Abort.redirectTo("https://vapor.codes")"
+    ///
+    /// Set type to '.permanently' to allow caching to automatically redirect from browsers.
+    /// Defaulting to non-permanent to prevent unexpected caching.
+    public static func redirectTo(_ location: String, type: Redirect = .normal) -> Abort {
```
0xTim, line 28: Have to say I'm not a fan of the new API - how about
```suggestion
    public static func redirect(to location: String, redirectType: Redirect = .normal) -> Abort {
```
Fits better with the Swift API guidelines, looks better and makes the redirect type clearer

0xTim (changes requested):
This looks in pretty good shape. Want to discuss the API though, suggested what I think is a better replacement

Sources/Vapor/Error/Abort.swift
NeedleInAJayStack: I prefer this too, the only issue is that if we keep around the deprecated `redirect(to, type)`, then when using the default `redirectType` argument, the call is ambiguous. That is, calling `request.redirect(to: "/")`, it wouldn't know whether it was the old `redirect(to, type)` or the new `redirect(to, redirectType)`, and will throw a compilation error.

0xTim: We can actually be clever/sneaky here - the old API doesn't need a default value. That migrates everyone to the new API unless they're specifying a redirect type without a breaking change

NeedleInAJayStack: Ohhh, that's so smart! I love it! I've made this change

Sources/Vapor/Error/Abort.swift
```diff
@@ -12,11 +12,24 @@ public struct Abort: AbortError, DebuggableError {
...
         return .init(type.status, headers: headers)
     }
+    
+    /// Creates a redirecting `Abort` error.
+    ///
+    ///     throw Abort.redirect(to: "https://vapor.codes")
+    ///
+    /// Set type to '.permanently' to allow caching to automatically redirect from browsers.
+    /// Defaulting to non-permanent to prevent unexpected caching.
+    public static func redirect(to location: String, redirectType: Redirect = .normal) -> Abort {
```
0xTim, line 28: This needs DocC comments for the return type and parameters

Sources/Vapor/Error/Abort.swift
```diff
@@ -12,11 +12,24 @@ public struct Abort: AbortError, DebuggableError {
     ///
     /// Set type to '.permanently' to allow caching to automatically redirect from browsers.
     /// Defaulting to non-permanent to prevent unexpected caching.
```
0xTim, line 14: If you could tweak this comment to add parameters and return type that would be awesome!

0xTim (changes requested):
Great! Just a couple of documentation changes then we're good!

Sources/Vapor/Request/Redirect.swift
```diff
@@ -7,17 +7,35 @@ extension Request {
     ///
     /// Set type to '.permanently' to allow caching to automatically redirect from browsers.
     /// Defaulting to non-permanent to prevent unexpected caching.
```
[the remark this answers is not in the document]
NeedleInAJayStack: Done

Sources/Vapor/Request/Redirect.swift
```diff
@@ -7,17 +7,35 @@ extension Request {
...
     }
+    
+    /// Creates a redirect `Response`.
+    ///
+    ///     router.get("redirect") { req in
+    ///         return req.redirect(to: "https://vapor.codes")
+    ///     }
+    ///
+    /// Set type to '.permanently' to allow caching to automatically redirect from browsers.
+    /// Defaulting to non-permanent to prevent unexpected caching.
```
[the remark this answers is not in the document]
NeedleInAJayStack: Done

Sources/Vapor/Error/Abort.swift
NeedleInAJayStack: Done

Sources/Vapor/Error/Abort.swift
NeedleInAJayStack: Done

0xTim (approved):
🙌

PR #2935, opened 2023-01-12
FileMiddleware redirect option

NeedleInAJayStack:
Previously, FileMiddleware would just serve the default file when requesting a directory with no trailing slash. This caused rendering issues when the default file loaded additional resources by relative path. This adds an option to either redirect to the url with a trailing slash (fixing rendering issues), or just pass the request through the middleware. The default is set to a redirect to preserve the intended functionality of FileMiddleware


Joannis:
I don't like the use of a boolean for a functionality as specific as this. If it were a `struct` with an `internal enum` inside of it, we could expand the redirect behaviour in the future without making a bunch of extra initialisers. Therefore I'd change the API to:
```swift
FileMiddleware(
  publicDirectory: "/" + path,
  defaultFile: "index.html",
  redirect: .none // or omit this parameter to get `.none` by default
)

FileMiddleware(
  publicDirectory: "/" + path,
  defaultFile: "index.html",
  redirect: .inferDirectories
)
```

vzsg:
This is the second time "struct with internal enum" came up today, but what exactly do you mean? There's no need for an enum at all, when you can just declare the struct along with two static fields for the two strategies.

Joannis:
That's what I mean @vzsg . You could wrap the underlying use cases in an enum, with associated values if needed. Then use the setup you described to wrap it

NeedleInAJayStack:
@Joannis Thanks for the review, and good input - I like how explicit the API becomes with struct properties.

I've updated it to use a `struct` wrapping an `enum` like you requested. I also changed `redirect` to `directoryAction` to better describe the functionality.

Since the new initializer has a default value for `directoryAction`, I'm not sure it breaks the public API. If so, I'm happy to preserve the original initializer and create a new one with `directoryAction`.

vzsg:
Oh, an interesting pattern. Instead of involving a strategy pattern, it's just the enum being hidden by a public struct that exposes the same options. TIL, thanks.

Joannis:
Thanks @NeedleInAJayStack , love the changes. I don't know whether we can merge because of API breakage, but I think we can. Ping @0xTim .

Joannis:
@NeedleInAJayStack I just merged another PR that's conflicting with yours. Can you resolve the conflicts? I'm happy to merge it then.

NeedleInAJayStack:
Sure, they should be resolved now. Thanks!

Joannis:
Thanks for the quick update!

NeedleInAJayStack:
Hey @0xTim, do you have any comments/concerns on this? Just want to note again that the CI API check is failing on a new function argument with a default, which I think is not a public API breakage.

Joannis:
Ping @0xTim

NeedleInAJayStack:
Hey, @Joannis / @0xTim is there anything I can do to help unblock this? I'd love to get it into main if possible...

Joannis:
@NeedleInAJayStack I'm awaiting @0xTim

0xTim:
Will be this week

0xTim (approved):
Thanks!

0xTim:
API Breakage tool is wrong - this is not a breaking change

PR #2943, opened 2023-01-25
Fix: FileIO headers

NeedleInAJayStack:
- Adds `Last-Modified` header to file requests
- Fixes `ETag` header formatting in file requests: According to [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag), the `ETag` header should start and end with double-quotes.

### DIFF
--- a/Sources/Vapor/Utilities/FileIO.swift
+++ b/Sources/Vapor/Utilities/FileIO.swift
@@ -152,7 +152,7 @@ public struct FileIO {
         var headers: HTTPHeaders = [:]
 
         // Generate ETag value, "HEX value of last modified date" + "-" + "file size"
-        let fileETag = "\(modifiedAt.timeIntervalSince1970)-\(fileSize)"
+        let fileETag = "\"\(modifiedAt.timeIntervalSince1970)-\(fileSize)\""
         headers.replaceOrAdd(name: .eTag, value: fileETag)
 
         // Check if file has been cached already and return NotModified response if the etags match
--- a/Sources/Vapor/Utilities/FileIO.swift
+++ b/Sources/Vapor/Utilities/FileIO.swift
@@ -160,7 +160,11 @@ public struct FileIO {
 
         // Check if file has been cached already and return NotModified response if the etags match
         if fileETag == request.headers.first(name: .ifNoneMatch) {
-            return Response(status: .notModified)
+            // Per RFC 9110 here: https://www.rfc-editor.org/rfc/rfc9110.html#status.304
+            // and here: https://www.rfc-editor.org/rfc/rfc9110.html#name-content-encoding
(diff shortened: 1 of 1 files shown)


NeedleInAJayStack:
A followup commit was added to make Vapor's `304: Not Modified` response compliant with [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#status.304).

Joannis:
Is there a potential security concern here? In the sense of leaking unneccessary/unwanted metadata?

NeedleInAJayStack:
@Joannis What header are you specifically concerned about leaking metadata? The `E-Tag` or the `Last-Modified`?

NeedleInAJayStack:
Just to note, Nginx adds these headers onto static files by default, and they are both used in common client-side caching strategies.

I'm not saying that there isn't a potential security concern here - just that usage of them appears to be somewhat common.

Joannis:
I understand @NeedleInAJayStack , but adding more metadata without it being an initial feature could lead to unexpected disclosure. That's all really, but I'd like the informed opinion of someone like @weissi

weissi:
> I understand @NeedleInAJayStack , but adding more metadata without it being an initial feature could lead to unexpected disclosure. That's all really, but I'd like the informed opinion of someone like @weissi

I think @Lukasa knows more about this than I do. My understanding how things "should work" is:

- `ETag` should be a hash of the contents (not just `\(lastModificationTime)-\(fileLength)`)
- I think of `Last-Modified` as a legacy thing, I didn't know that was still necessary today?

But to answer the actual security question: Given that the old behaviour for `ETag` was `\(modifiedAt.timeIntervalSince1970)-\(fileSize)`, adding `Last-Modfied` to `\(modifiedAt.timeIntervalSince1970.unixSecs)` wouldn't make the security situation any worse because the information is already sent to the client today. (Unless I'm missing smth)

Joannis:
I concur @weissi , I didn't think of the fact we're already sending the timestamp through the ETag. Thanks for your feedbacK!

Joannis (approved):
LGTM

Lukasa:
[`Last-Modified` is very much current](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#field.last-modified) and recommended if we can easily compute it, especially if we can appropriately cache subsequent requests.

Last modification time and file length are definitely not sufficiently strong for an Etag, a hash output is more generally safe. Note that if you're going to hash it you may want to find a way to precompute that computation, as computing it will necessarily involve doing all the same file I/O as a regular access.

Joannis:
@Lukasa I agree, pre-caching hashes should also be introduced in the API surface then, which is currently not the case. I do think that this would be a larger scope than was intended for this PR. I'll add an issue for that, good and important catch! Thank you for the keen eye.

PR #2957, opened 2023-02-13
Use YML compatible quotes in the `.spi.yml`

daveverwer:
We're picking these up as part of the text:

Screenshot 2023-02-13 at 12 44 52@2x


Joannis:
@daveverwer does this require a release?

daveverwer:
> @daveverwer does this require a release?

No, we pick this information up from the default branch. As soon as it's merged it'll update within a few hours (next time we poll the repository).

PR #2960, opened 2023-02-15
Implement Server-Sent Events

Joannis:
Has an `sse.html` to test the `GET /sse` route in the Development app. I'll force-push that to be removed once I get a unit test as a replacement.

### DIFF
--- a/Sources/Vapor/SSE/SSE.swift
+++ b/Sources/Vapor/SSE/SSE.swift
@@ -1,0 +1,71 @@
+enum SSEError: Error {
+    case notAString, protocolViolation
+}
+
+public struct SSEValue: ExpressibleByStringLiteral {
+    fileprivate var parts = [Substring]()
+    
+    public var string: String {
+        get {
+            parts.joined(separator: "\n")
+        }
+        set {
+            assert({
+                !newValue.contains("\n") && !newValue.contains("\r")
+            }())
+            
+            parts = newValue.split(omittingEmptySubsequences: true) { character in
+                character == "\r" || character == "\n"
+            }
+        }
+    }
+    
+    public init(string: String) {
+        parts = string.split(omittingEmptySubsequences: true) { character in
+            character == "\r" || character == "\n"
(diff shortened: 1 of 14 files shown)


jhoughjr (approved):
Looks good to me.

Sources/Vapor/SSE/SSE.swift
```diff
@@ -0,0 +1,76 @@
...
+    }
+    
+    internal init(unchecked parts: [String]) {
+        self.parts = parts.map {
+            Substring($0)
+        }
+    }
+}
+
+public struct SSEvent {
```
Steven4294, line 40: Not going to be a stickler, but I suppose this should be named `SSEEvent` instead

jhoughjr: I disagree. Its a Server Side Event. Server Side Event Event would be redundant.

Steven4294: Totally reasonable and fair, you have changed my mind

wendellli99:
Do we know when this will be merged? Trying to use SSE for MacPaw OpenAI's chat stream.

Joannis:
@wendellli99 it can't get reviewed yet, as this PR depends on the unmerged async streams PR

wendellli99:
Ok. Any chance we can get that PR merged? We are eagerly waiting for SSE.

Sherlouk:
Looks like work was merged a couple months ago which could unblock this again. Wondered if there were intentions to revitalise this? 🙂

Joannis:
@Sherlouk merged and reverted

Joannis:
If you can get access to an AsyncSequence<ByteBuffer>, you can now use this: https://github.com/orlandos-nl/SSEKit

awamser:
Can I "sponsor" this so it gets merged?

0xTim:
@awamser have you tried using Joannis' package? That should work fine

PR #2963, opened 2023-02-19
Fix up Imports

0xTim:
Fixes up the imports in Vapor so it will build without exporting everything.

This allows us to avoid exports when building documentation and future proofs us.

Also run this through CI to avoid PRs breaking the docs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,16 +38,12 @@ jobs:
        with_tsan: false
        with_public_api_check: true
   test-exports:
-      name: Test exports
-      runs-on: ubuntu-latest
-      steps:
-        - name: Check out Vapor
-          uses: actions/checkout@v3
-          with:
-            fetch-depth: 0
-        - name: Remove exports
-          run: |
-            rm -rf Sources/Vapor/Exports.swift
-            rm -rf Sources/XCTVapor/Exports.swift
-        - name: Build
-          run: swift build
+    name: Test exports
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Vapor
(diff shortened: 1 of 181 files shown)


.github/workflows/test.yml
```diff
@@ -37,3 +37,13 @@ jobs:
...
+  test-exports:
+    name: Test exports
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Vapor
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Build
+        run: swift build -Xswiftc -DBUILDING_DOCC
```
gwynne, line 49: `-D` flags can't be passed directly to `swift build`? I could've sworn that was supported.

.github/workflows/test.yml
```diff
@@ -37,3 +37,13 @@ jobs:
        with_coverage: false
        with_tsan: false
```
gwynne, line 38: While we're at it, we should either turn off TSan for the provider tests or turn it on for the unit tests.

gwynne (approved):
It's kinda sad just how inaccurate the API breakage checker can be about some things...

0xTim:
@gwynne We should probably raise issues for all these API checker problems... 😅

.github/workflows/test.yml
0xTim: Honestly I didn't actually try! I'll try in the next repo and we can always backport it

.github/workflows/test.yml
0xTim: It should probably be on really. I'll make a note to enable it when I do some Sendable work

PR #2964, opened 2023-02-21
Expose Vapor Request's request-id value, so that it can be passed into logging from other libraries

Joannis:
Exposes the Request's ID as a new property on `Request` as `requestID`. This is useful for any libraries or users who want to be able to trace the request ID through a request handler.


Joannis:
See https://github.com/orlandos-nl/MongoKitten/issues/300

0xTim (approved):
Thanks! This should get reused when it comes to tackling tracing as well

0xTim (changes requested):
Actually I'm going to be a real pain here - can you just add a simple test to make 2 requests and ensure the IDs are different? That'll protect us in the future for any changes and stop silly mistakes

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
```diff
@@ -134,7 +134,7 @@ private final class ChannelResponseBodyStream: BodyStreamWriter {
             self.context.fireUserInboundEventTriggered(HTTPServerResponseEncoder.ResponseEndSentEvent())
             self.context.writeAndFlush(self.handler.wrapOutboundOut(.end(nil)), promise: promise)
             self.promise?.fail(error)
-        }
+            }
```
0xTim, line 137: Indentation
```suggestion
        }
```

0xTim:
Original comment:
> Libraries such as Fluent, MySQL, Postgres, MongoKitten & Redis currently have their own logging set up using swift-log. However, it can be useful to see the request-id from which a query was launched. Therefore I propose exposing the request-id as a property on `Request` so that end-users can set the metadata in other libraries' loggers as well. Therefore improving the debugability using these logs.

Issue #2965, opened 2023-02-21
FloatingPoint.nan does not fail range validation


fred-sch:
### Describe the bug

`FloatingPoint.nan` does not fail range validation.

I would have expected that the following validation is a failure if the latitude can be decoded to `Float.nan` but instead it passes the validation.

```swift
struct MyQuery: Codable {
    let lat: Float
    let lon: Float
}

extension MyQuery: Vapor.Validatable {
    public static func validations(_ validations: inout Validations) {
        validations.add("lat", as: Float.self, is: .range(-90...90), required: true)
    }
}
```

PR #2966, opened 2023-02-21
Correctly handle invalid numbers in range validations

fred-sch:
Previously trying to perform a range validation check on `Float.nan` would cause a `fatalError` - this correctly handles it and throws an error instead.

Closes https://github.com/vapor/vapor/issues/2965.


fred-sch:
I think there is unfortunately no real way to avoid the api breakage. Admittedly, it is kind of counterintuitive that there should be a `case notComparable` for something that operates on the `Comparable` protocol, but `FloatingPoint.nan` is exactly this case as it will always yield `false` in any comparison (even comparing it with itself).

0xTim:
Just skimming this - since the initialiser is internal we could mark it as throwing and throw an error if you try to compare something that isn't comparable and then handle that up the stack. Just a thought and not sure if higher up in the stack can handle it easily. What do you think?

fred-sch:
That is a good idea! I did as you asked in https://github.com/vapor/vapor/pull/2966/commits/f87494b25bd7955beca7d9d61e2ea804c7fac668

0xTim (approved):
Thanks!

PR #2969, opened 2023-03-01
Better `DecodingError` description

MahdiBM:
The `DecodingError`'s description, when returned from HTTP requests, was missing some important info and sometimes was not clear enough. Hopefully this PR can resolve that issue.


MahdiBM:
I'll take care of the tests if the changes seem OK.

0xTim (changes requested):
Makes sense but would like to see some tests for the 3 different error paths

Tests/VaporTests/ContentTests.swift
```diff
@@ -432,7 +432,85 @@ final class ContentTests: XCTestCase {
...
+            headersNoUpdate: ["Content-Type": "application/json"],
+            collectedBody: ByteBuffer(string: #"{"badJson: "Key doesn't have a trailing quote"}"#),
+            on: app.eventLoopGroup.next()
+        )
+        
+        struct DecodeModel: Content {
+            let badJson: String
+        }
+        XCTAssertThrowsError(try req.content.decode(DecodeModel.self)) { error in
+            XCTAssertContains(
```
MahdiBM, line 457: Had to use "Contains"  check instead of "Equal", because the errors could differ on macOS vs on Linux.

Sources/Vapor/Error/AbortError.swift
```diff
@@ -69,20 +69,40 @@ extension DecodingError: AbortError {
...
+
+private extension DecodingError.Context {
+    var debugDescriptionAndUnderlyingError: String {
+        "\(self.debugDescriptionNoTrailingDot)\(self.underlyingErrorDescription)."
+    }
+    
+    /// `debugDescription` sometimes has a trailing dot, and sometimes not.
+    private var debugDescriptionNoTrailingDot: String {
+        if self.debugDescription.isEmpty {
+            return ""
```
MahdiBM, line 93: didn't mention it in the documentation above the variable but i saw somewhere in JSONDecoder there was an empty `debugDescription` passed into the context.

Sources/Vapor/Error/AbortError.swift
```diff
@@ -69,20 +69,40 @@ extension DecodingError: AbortError {
     public var reason: String {
         switch self {
         case .dataCorrupted(let ctx):
```
MahdiBM, line 71: I took a look at the JSONDecoder and Codable implementations and these descriptions are what i came up

Tests/VaporTests/ContentTests.swift
```diff
@@ -432,7 +432,85 @@ final class ContentTests: XCTestCase {
         XCTAssertThrowsError(try req.content.decode(PostInput.self)) { error in
             XCTAssertEqual(
                 (error as? AbortError)?.reason,
-                "Value required for key 'is_free'."
+                #"Value required for key at path 'is_free'. No value associated with key CodingKeys(stringValue: "is_free", intValue: nil) ("is_free")."#
+            )
+        }
+    }
+
+    func testDataCorruptionError() throws {
```
MahdiBM, line 440: tests added for the 3 missing cases. 1 case already had a test (the one above)

Tests/VaporTests/ContentTests.swift
```diff
@@ -432,7 +432,85 @@ final class ContentTests: XCTestCase {
         XCTAssertThrowsError(try req.content.decode(PostInput.self)) { error in
             XCTAssertEqual(
                 (error as? AbortError)?.reason,
-                "Value required for key 'is_free'."
+                #"Value required for key at path 'is_free'. No value associated with key CodingKeys(stringValue: "is_free", intValue: nil) ("is_free")."#
```
MahdiBM, line 435: The explanation basically repeats itself twice. This is the case with Foundation's `JSONDecoder` errors but i thought maybe someone uses another decoder so better to keep the info

MahdiBM:
had to cancel 2 macos tests because it was taking 18+ minutes. unlikely to be related to this pr

0xTim (approved):

0xTim:
Thanks!

PR #2971, opened 2023-03-02
Add support for Cache-Control header 'immutable' attribute

Palleas:
Improve parsing of Cache-Control header to parse the `immutable` attribute.


0xTim (approved):
Thanks!

PR #2973, opened 2023-03-06
Revise api doc

makhocheung:
revise the doc of async version of FileIO.collectFile

### DIFF
--- a/Sources/Vapor/Concurrency/FileIO+Concurrency.swift
+++ b/Sources/Vapor/Concurrency/FileIO+Concurrency.swift
@@ -5,7 +5,7 @@ import NIOCore
 extension FileIO {
     /// Reads the contents of a file at the supplied path.
     ///
-    ///     let data = try await req.fileio().read(file: "/path/to/file.txt")
+    ///     let data = try await req.fileio.collectFile(file: "/path/to/file.txt")
     ///     print(data) // file data
     ///
     /// - parameters:


makhocheung:
```
    /// Reads the contents of a file at the supplied path in chunks.
    ///
    ///     try await req.fileio().readChunked(file: "/path/to/file.txt") { chunk in
    ///         print("chunk: \(data)")
    ///     }
    ///
    /// - parameters:
    ///     - path: Path to file on the disk.
    ///     - chunkSize: Maximum size for the file data chunks.
    ///     - onRead: Closure to be called sequentially for each file data chunk.
    /// - returns: `Void` when the file read is finished.
    // public func readFile(at path: String, chunkSize: Int = NonBlockingFileIO.defaultChunkSize, onRead: @escaping (ByteBuffer) async throws -> Void) async throws {
    //     // TODO
    //     // We should probably convert the internal private read function to async as well rather than wrapping it at this top level
    //     let promise = self.request.eventLoop.makePromise(of: Void.self)
    //     promise.completeWithTask {
    //         try await onRead
    //     }
    //     let closureFuture = promise.futureResult
    //     return try self.readFile(at: path, onRead: closureFuture).get()
    // }
```
I also found the useless code. Can it be removed?

0xTim:
@makhocheung that was left out waiting for either a bug to be fixed or better support in NIO for async file streaming. If you want to try a PR to bring it back, feel free!

0xTim (approved):
Thanks!

PR #2981, opened 2023-03-14
Improve error reporting for `EncodingError` and `DecodingError`

grahamburgsma:
As `EncodingError` and `DecodingError` both conform to `LocalizedError`, error reporting uses their `localizedDescription` which is just a generic error message.

So this changes the error reporting from:
> The operation could not be completed. The data isn’t in the correct format.

To a description much more useful (for example):
> invalidValue(inf, Swift.EncodingError.Context(codingPath: [...], debugDescription: "Unable to encode Double.inf directly in JSON.", underlyingError: nil))


0xTim (approved):
Thanks for this! LGTM

Issue #2994, opened 2023-04-04
Trivially enable async/await with Xcode 13.2.1 on Big Sur


mteep:
Currently Vapor cannot be built with async/await support using Xcode 13.2.1, which is the last version of Xcode supported on macOS Big Sur.

This is despite the fact that if Vapor is built with a later Xcode version, with async/await, it can still run on Big Sur. And also despite the fact that Xcode 13.2.1 contains everything actually needed to build Vapor with async/await.

A simple change to `Package.swift` in Vapor (and related packages like async-kit, fluent and fluent-kit) would fix this issue. An issue that seems to stem from uncertainty about what the swift-tools-version means.

Xcode 13.2.1 includes Swift 5.5.2, but the SwiftPM version is still 5.5.0, for reasons unknown to me. (There is a standalone build of Swift 5.5.3 for macOS, with no changes except bumping the versions of both Swift and SwiftPM to 5.5.3. The Swift part of this build can be used with Xcode 13.2.1, but that doesn't affect the SwiftPM support in Xcode.)

I'm not sure why Vapor wants Swift 5.5.2 over Swift 5.5, but if that distinction is important to Vapor (it isn't to Swift-NIO, as some people seem to claim), it can be expressed in the `Package.swift` file, without requiring SwiftPM beyond 5.5.

One way is to take advantage of the Swift version supported by the compiler compiling `Package.swift`. I have an example of that in a branch called [mergeable-big-sur](https://github.com/mteep/vapor/tree/mergeable-big-sur).

0xTim:
Just to clarify this as the PR is closed - 5.5.2 ensures that anyone building on Xcode 13.0 and 13.1 would correctly take an older version of Vapor that worked with these releases. 5.5.2 ensures that the concurrency runtime is bundled with the toolchain otherwise you'd get weird errors

mteep:
By "weird errors", do you mean that the final product needs to ensure that the rpath needs to be able to locate the concurrency library? That is correct, but simple to work around.

0xTim:
Yes if on macOS 13.0 there was no Concurrency runtime to link against so even manually setting an rpath wouldn't help without all kinds of hacks. This has been superseded by #2996 anyway as we need it for some `Sendable` stuff that's upcoming

PR #2995, opened 2023-04-05
Enable async/await with Xcode 13.2.1 on Big Sur

mteep:
Enables building with async/await using any Xcode 13.x or later, with corresponding target platform requirements.

This is achieved by adjusting `Package.swift` to only require SwiftPM 5.5.0, as shipped with all Xcode 13.x versions.

Xcode versions that shipped with Swift versions below 5.5.2 didn't fully support back-deployment of async/await. In this case the revised manifest requires Apple platforms that shipped with libswift_Concurrency.dylib, that is macOS 12, iOS/iPadOS/tvOS 15, and watchOS 8.

However, Xcode 13.2.x and later, that shipped with Swift 5.5.2 or later, does support back-deployment of async/await two major OS versions. Thus the revised manifest in this case only requires macOS 10.15, iOS/iPadOS/tvOS 13, and watchOS 6, as before.

Prior to this change, Xcode 13.3, which doesn't run on macOS 11 Big Sur, was required to build Vapor with async/await.

With this change, and corresponding changes in related packages such as async-kit, fluent, and fluent-kit, macOS 11 Big Sur becomes the first single platform on which Vapor with async/await can be built, back-deployed and debugged using Xcode. (#2994)


0xTim:
Unfortunately it's not this simple. If we were to enable this change we'd also need to wrap all of our concurrency code in availability checks and check we can import concurrency. Xcode 13.2.1 should work as it ships with Swift 5.5.2 and works on macOS 11 according to https://xcodereleases.com

Due to the increased maintenance burden I'm going to close this PR as 13.2.1 should work fine. Note that we are about to drop support for 5.5 altogether in the very near future so you might need older versions of Vapor anyway.

mteep:
I think you are confusing the Swift version with the SwiftPM version. They are not the same. No other changes are needed, except that the final product may have to ensure that the concurrency library is in the rpath.

The "not this simple" is exactly why this problem exists in the first place.

PR #2996, opened 2023-04-11
Update Minimum Swift Version to 5.6

0xTim:
Inline with the Swift version support and NIO's supported version, this sets the minimum supported Swift version to 5.6 now that Swift 5.8 is released.

This is also the first step in adopting `Sendable` properly across Vapor


gwynne (approved):
🎉❤️🎉❤️🎉❤️🎉

PR #2997, opened 2023-04-11
Support range-requests beyond EOF. Return up to EOF.

keniwhat:
Allow byte range requests to exceed EOF.

Support HTTP/1.1 range requests past end-of-file.

### DIFF
--- a/Sources/Vapor/HTTP/Headers/HTTPHeaders+ContentRange.swift
+++ b/Sources/Vapor/HTTP/Headers/HTTPHeaders+ContentRange.swift
@@ -262,11 +262,12 @@ extension HTTPHeaders.Range.Value {
             }
             return .withinWithLimit(start: (limit - end), end: limit - 1, limit: limit)
         case .within(let start, let end):
-            guard start >= 0, end >= 0, start < end, start <= limit, end <= limit else {
+            guard start >= 0, end >= 0, start < end, start <= limit else {
                 throw Abort(.badRequest)
             }
             
-            return .withinWithLimit(start: start, end: end, limit: limit)
+            let myEnd = min(end, limit)
+            return .withinWithLimit(start: start, end: myEnd, limit: limit)
         }
     }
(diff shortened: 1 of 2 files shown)


Sources/Vapor/HTTP/Headers/HTTPHeaders+ContentRange.swift
```diff
@@ -262,11 +262,12 @@ extension HTTPHeaders.Range.Value {
...
             return .withinWithLimit(start: (limit - end), end: limit - 1, limit: limit)
         case .within(let start, let end):
-            guard start >= 0, end >= 0, start < end, start <= limit, end <= limit else {
+            guard start >= 0, end >= 0, start < end, start <= limit else {
                 throw Abort(.badRequest)
             }
             
-            return .withinWithLimit(start: start, end: end, limit: limit)
+            let myEnd = min(end, limit)
+            return .withinWithLimit(start: start, end: myEnd, limit: limit)
```
0xTim, line 270: Clearer naming

```suggestion
            let endToRequest = min(end, limit)
            return .withinWithLimit(start: start, end: endToRequest, limit: limit)
```

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -288,7 +289,8 @@ extension HTTPHeaders.Range.Value {
                     logger.debug("Requested range was invalid: \(start)-\(end)")
                     throw Abort(.badRequest)
                 }
-                return (offset: numericCast(start), byteCount: byteCount)
+                let myByteCount = end <= size ? byteCount : byteCount - (end - size)
```
0xTim, line 292: Can you expand this into a proper if statement to make it clearer what's going on? And then name it to `byteCountToReqeust` or similar

0xTim (changes requested):
We'll need a test for this as well as the review comments - thanks!

PR #3000, opened 2023-04-17
Add Sendable Conformances to Vapor

0xTim:
Adds `Sendable` conformances where appropriate. This removes almost all of the warnings when building Vapor with strict concurrency checking. Fixes a large number of concurrency and thread-safety issues caused by the introduction of Swift concurrency.

Huge thanks to @dnadoba @Lukasa and everyone else who helped getting this over the line!

This also drops support for Swift 5.7 because of the use of `@preconcurrency`


Sources/Vapor/Bcrypt/Bcrypt.swift
```diff
@@ -210,7 +210,7 @@ public final class BCryptDigest {
     }
 }
 
-public enum BcryptError: Swift.Error, CustomStringConvertible, LocalizedError {
+public enum BcryptError: Swift.Error, CustomStringConvertible, LocalizedError, Sendable {
```
dnadoba, line 213: `Swift.Error` already requires `Sendable` conformance. No need to the conformance again. This is applies to all custom error types in this PR.

Sources/Vapor/Authentication/AuthenticationCache.swift
```diff
@@ -56,22 +58,33 @@ extension Request.Authentication {
         return self.get(A.self) != nil
     }
 
-    private final class Cache {
+    // Internal `storage` is behind a lock so this can be unchecked Sendable
+    private final class Cache: @unchecked Sendable {
         private var storage: [ObjectIdentifier: Any]
-
+        private let storageLock: NIOLock
```
dnadoba, line 64: Use `NIOLockedValueBox` instead of `storageLock` and separate `storage`. The compiler can then infere the `Sendable` conformance and you can drop `@unchecked`. This will also allow you to make this a struct instead of a class as `NIOLockedValueBox` is already a reference type. Document that the struct has reference semantics now.

Sources/Vapor/Bcrypt/Bcrypt.swift
```diff
@@ -180,7 +180,7 @@ public final class BCryptDigest {
     }
 
     /// Specific BCrypt algorithm.
-    private enum Algorithm: String, RawRepresentable {
+    private enum Algorithm: String, RawRepresentable, Sendable {
```
dnadoba, line 183: `Sendable` should be automatically inferred for internal and lower visibility types. If not, please fill an issue.

dnadoba (commented):
Looked at a small portion of it and left some nits on the way. Looks good overall though!

Sources/Vapor/Authentication/AuthenticationCache.swift
0xTim: I can't use `NIOLockedValueBox` though because of the use of `Any` right? I can't guarantee that anything being inserted into the storage is `Sendable` at which point `NIOLockedValueBox` is _not_ `Sendable`

stevapple:
> This also drops support for Swift 5.7 because of the use of `@preconcurrency`

Didn't expect to drop 5.7 before 5.6 LOL. A tiny (yet key) typo.

Sources/Vapor/View/Application+Views.swift
```diff
@@ -10,27 +10,28 @@ extension Application {
...
+            let run: @Sendable (Application) -> ()
 
-            public init(_ run: @escaping (Application) -> ()) {
+            public init(_ run: @Sendable @escaping (Application) -> ()) {
                 self.run = run
             }
         }
         
+        // This doesn't need a lock as it's only mutated during app configuration
         final class Storage {
```
ffried, line 29: Can become `Sendable` as well (like other `Storage`'s).

Tests/VaporTests/ApplicationTests.swift
```diff
@@ -22,7 +22,7 @@ final class ApplicationTests: XCTestCase {
     }
 
     func testLifecycleHandler() throws {
-        final class Foo: LifecycleHandler {
+        final class Foo: LifecycleHandler, @unchecked Sendable {
```
ffried, line 25: `@unchecked` is not needed, is it?

Sources/Vapor/Validation/Validators/Range.swift
```diff
@@ -53,9 +56,27 @@ extension Validator {
     }
 }
 
+// TODO: Remove when keypaths are `Sendable`
+/// ``UnsafeMutableTransferBox`` can be used to make non-`Sendable` values `Sendable` and mutable.
+/// It can be used to capture local mutable values in a `@Sendable` closure and mutate them from within the closure.
+/// As the name implies, the usage of this is unsafe because it disables the sendable checking of the compiler and does not add any synchronisation.
+@usableFromInline
+final class UnsafeMutableTransferBox<Wrapped> {
```
ffried, line 64: This is only used in this file as far as I've seen. Maybe make it `fileprivate` for now to prevent mis-use?

ffried (commented):
Came across this, waiting for it to drop :)

Left quite a few review comments (sorry for the noise!) - most of them are just small inconsistencies.
Maybe some of them aren't valid, e.g. if you always explicitly declare conformances (instead of relying on inheritance).

Sources/Vapor/Application.swift
```diff
@@ -26,54 +61,80 @@ public final class Application {
...
+            _lifecycle.withLockedValue { $0 = newValue }
+        }
+    }
 
-    public final class Locks {
+    // This does not have value semantics due to the use or `NIOLockedValueBox`
+    public struct Locks: Sendable {
         public let main: NIOLock
-        var storage: [ObjectIdentifier: NIOLock]
+        let storage: NIOLockedValueBox<[ObjectIdentifier: NIOLock]>
```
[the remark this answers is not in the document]
Joannis: It is necessary @ffried . Dictionary doesn't apply locking by itself, and access to Dictionary therefore needs to be Sendable as well, through an actor or locks. The inner lock serves a differet purpose

Sources/Vapor/Application.swift
```diff
@@ -26,54 +61,80 @@ public final class Application {
...
+            return _lifecycle.withLockedValue { $0 }
+        }
+        set {
+            _lifecycle.withLockedValue { $0 = newValue }
+        }
+    }
 
-    public final class Locks {
+    // This does not have value semantics due to the use or `NIOLockedValueBox`
+    public struct Locks: Sendable {
```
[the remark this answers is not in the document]
Joannis: This is very much a breaking change. You could do access `let locks = application.locks` and any access to `locks` would mismatch the state in `application.locks`

Sources/Vapor/Application.swift
```diff
@@ -26,54 +61,80 @@ public final class Application {
...
         }
     }
 
-    public var locks: Locks
+    public var locks: Locks {
+        get {
+            return _locks.withLockedValue { $0 }
+        }
+        set {
+            _locks.withLockedValue { $0 = newValue }
```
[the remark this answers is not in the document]
Joannis: It's not 'locking to access a lock'. It's 'locking to access a dictionary', independency of it containing a lock for a value

Sources/Vapor/Authentication/AuthenticationCache.swift
Joannis: I see what you mean @ffried. It's not a huge deal either way.

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
```diff
@@ -35,27 +38,29 @@ final class HTTPServerHandler: ChannelInboundHandler, RemovableChannelHandler {
...
     
     func serialize(_ response: Response, for request: Request, context: ChannelHandlerContext) {
+        let contextBox = NIOLoopBound(context, eventLoop: context.eventLoop)
         switch request.version.major {
         case 2:
             context.write(self.wrapOutboundOut(response), promise: nil)
         default:
-            let keepAlive = !self.isShuttingDown && request.isKeepAlive
-            if self.isShuttingDown {
+            let keepAlive = !self.isShuttingDown.withLockedValue({ $0 }) && request.isKeepAlive.withLockedValue { $0 }
```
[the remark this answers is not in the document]
Joannis: No, then you're holding the outer lock longer than necessary without advantages. If the holder of the `isKeepAlive` lock were to (indirectly) access `isShuttingDown`, you're deadlocking.

Sources/Vapor/Application.swift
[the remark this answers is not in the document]
ffried: @Joannis I don't think so. All properties of `Locks` have reference semantics (`NIOLock` is a class and `NIOLockedValueBox` has documented reference semantics). So `Locks` itself essentially still has reference semantics as well.

Sources/Vapor/Application.swift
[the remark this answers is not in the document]
ffried: @Joannis Which dictionary do you mean in this case? The one inside `Locks` is again protected by a separate lock. This access here only protects the `Locks` property on `Application` as far as I see. Which in term is only necessary because `locks` is settable. There are no `mutating` members on `Locks`, so I'd argue this setter is very seldomly used.

Sources/Vapor/Application.swift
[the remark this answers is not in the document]
ffried: @Joannis Sure, `Dictionary` itself isn't thread safe. But is there another access to it apart from the `lock(for:)` function below? Because said `lock(for:)` function below actually locks `main` first. So access to `storage` is already synchronized by main.
With the `NIOLockedValueBox` it currently has to hold **two** locks inside `lock(for:)`. That's what I was referring to.
If there are other access paths to `storage` that aren't protected by the `main` lock (like in `lock(for:)`), then it might make sense to have a `NIOLockedValueBox` around it (though holding the `main` lock could be removed in `lock(for:)` then).

chibombo:
Will this PR fix the Vapor `@preconcurrency` warnings?

`@preconcurrency 
/mydir/.build/checkouts/randomfile.swift:1:1: remark: add '@preconcurrency' to suppress 'Sendable'-related warnings from module 'Vapor'
import Vapor`

0xTim:
@chibombo the idea is yes but this PR won't be merged. We've been adding Sendable stuff in piece-by-piece as we learn more about it etc, including the recent #3082 that just got merged. 

I think we're just down to `Request` now which needs to be migrated, which is the hardest one 😅 

Once that's in, the warning should disappear

0xTim:
Closing this as we've been doing this in different parts

Issue #3003, opened 2023-04-21
Ability to not create 64 threads when booting vapor.


John-Connolly:
### Describe the bug

On iOS platforms the default threadpool size of 64 is a bit aggressive. Sometimes iOS will not want to create that many threads leading to a crash  in ``` ThreadOpsPosix.run (ThreadPosix.swift:109)```.  Even when swapping the ```NIOThreadPool``` to one with only one thread doesn't prevent the default one from creating 64 threads.

### To Reproduce
Boot a vapor app.

n/a this is the default behaviour.

### Expected behavior
```  app.threadPool = NIOThreadPool(numberOfThreads: 1)``` should replace the default threadpool before it's started. 

### Environment

* Vapor Framework version: [4.76.0](https://github.com/vapor/vapor/releases/tag/4.76.0)
* Vapor Toolbox version: N/A
* OS version: N/A

### Additional context
 I'm happy to provide a patch for this.

0xTim:
Yes this seems sensible and has come up in the past - happy to accept a patch for this change in behaviour

dfed:
Adding a +1 to this. When deploying Vapor on iOS we're creating _way_ too many threads. Would love to be able to cap this in the initializer.

gwynne:
Addressed by #3092 and the preexisting ability to replace the default thread pool.

PR #3008, opened 2023-05-01
Start thread pool when server is booted

John-Connolly:
Patch for https://github.com/vapor/vapor/issues/3003


John-Connolly:
@0xTim Let me know if this is an okay solution. To me it makes sense to start the thread pool in ```boot``` My only worry is that some users have depended on the thread pool being booted when the application is initialized.

PR #3009, opened 2023-05-05
Fixed drain handler call order in case of asynchronous buffer handling

salpieiev:
During request body drain each chunk of data may be handled asynchronously. This may lead to a call of handler with `.end` parameter before previous call with `.buffer(buffer)` finished it's execution.

For example:
1. Open new file descriptor
2. During `.collected` stage write ByteBuffer on disc using this file descriptor
3. During `.end` stage close file descriptor

This fix ensures that file descriptor won't be closed until disc write completes


0xTim (approved):
I have no idea why this was like this before! It was probably fine pre async/await so never got caught. Thanks!

PR #3010, opened 2023-05-06
Fix `Range: bytes=0-0` header not working properly

ptoffy:
This aims to fix the bug found in https://discord.com/channels/431917998102675485/519613337638797315/1104341522473812039, which returned a `Bad Response` rather then returning the first byte of the requested content when the `Range: bytes=0-0` request header was used


0xTim (changes requested):
Gonna need tests 😬

vzsg:
[These cases](https://devblogs.microsoft.com/dotnet/asp-net-web-api-and-http-byte-range-support/#trying-it-out) might be useful for writing tests.

PR #3015, opened 2023-05-14
Advanced ETag Comparison now supported

### LINKED ISSUE #2948: ETag should be a hash
### Describe the bug

Currently, ETag for file serving is generated from the modified date and file size. A hash is a stronger alternative, and can be precomputed if available. See #2943.

linus-hologram:
Vapor now supports strong (byte-by-byte) ETag validation and caches ETags for rapid responses. This provides a stronger alternative to the current weak comparison, which only guarantees semantic file equivalence. This new strong comparison is enabled by default and can be deactivated during `FileMiddleware` initialization if needed. This PR closes #2948.

- `streamFile` method was deprecated and replaced by an alternative returning an `EventLoopFuture`
- vapor's unit tests were updated to reflect the changes
- documentation was updated to reflect the changes


Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 import NIOCore
+import CryptoKit
```
vzsg, line 3:
```suggestion
import Crypto
```

linus-hologram: resolved with e2c3e4e

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -22,17 +23,28 @@ public final class FileMiddleware: Middleware {
         public static let publicDirectoryIsNotAFolder: Self = .init(description: "Cannot find any actual folder for the given Public Directory")
     }
 
+    public struct ETagHashes: StorageKey {
```
0xTim, line 26: This doesn't need to be public

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -215,10 +218,134 @@ public struct FileIO {
                 onCompleted(result)
             }
         }, count: byteCount, byteBufferAllocator: request.byteBufferAllocator)
-        
+
         return response
     }
 
+    /// Generates a chunked `Response` for the specified file. This method respects values in
```
0xTim, line 225: We should add here that the hash is returned in the ETag header

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -265,11 +392,37 @@ public struct FileIO {
...
         }
     }
+
+    /// Generates a fresh ETag for a file or returns its currently cached one.
+    /// - Parameters:
+    ///   - path: The file's path.
+    ///   - lastModified: When the file was last modified.
+    /// - Returns: An `EventLoopFuture<String>` which holds the ETag.
+    private func generateETagHash(path: String, lastModified: Date) -> EventLoopFuture<String> {
+        return NIOLock().withLock {
```
0xTim, line 408: We should have a central lock on `FileIO` rather than recreating one every time

0xTim: On second thoughts we don't need a lock here at all - once the `Sendable` stuff is landed, the application storage will be thread-safe so we can just use that

Tests/VaporTests/RequestTests.swift
```diff
@@ -261,13 +261,13 @@ final class RequestTests: XCTestCase {
         app.http.client.configuration.redirectConfiguration = .disallow
 
         app.get("redirect_normal") {
-            $0.redirect(to: "foo", type: .normal)
+            $0.redirect(to: "foo", redirectType: .normal)
```
0xTim, line 264: These are deliberately testing the old API to ensure it still works

Tests/VaporTests/FileTests.swift
```diff
@@ -374,8 +374,8 @@ final class FileTests: XCTestCase {
         let app = Application(.testing)
         defer { app.shutdown() }
 
-        app.get("file-stream") { req in
-            return req.fileio.streamFile(at: #file)
+        app.get("file-stream") { req -> EventLoopFuture<Response> in
+            return req.fileio.streamFile(at: #file, advancedETagComparison: true)
```
0xTim, line 378: We're setting this to true everywhere but not actually testing it. We should have assertions on the etag value returned for both when this is false and when this is true

0xTim (changes requested):
Thanks for this! Here are my initial review comments

Sources/Vapor/Middleware/FileMiddleware.swift
linus-hologram: resolved with f192a76

Sources/Vapor/Utilities/FileIO.swift
linus-hologram: resolved with f192a76

Sources/Vapor/Utilities/FileIO.swift
linus-hologram: > On second thoughts we don't need a lock here at all - once the `Sendable` stuff is landed, the application storage will be thread-safe so we can just use that

Alright sweet, I removed the lock with f192a76 👍

Tests/VaporTests/RequestTests.swift
linus-hologram: resolved with f192a76

Tests/VaporTests/FileTests.swift
linus-hologram: @0xTim alright, can you please briefly elaborate what the `#file` is in the code/what it refers to? I've never seen this before... I assume it's some sort of reference to a sample file.

If I can find that sample file, I'll write checks on the respective ETag headers.

linus-hologram: resolved with 37f16d1

0xTim (approved):
Ok this looks good once #3000 lands. I'll do a final review before merging

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -265,11 +394,35 @@ public struct FileIO {
...
+
+        if let hash = hashingDictionary[path], hash.lastModified == lastModified {
+            return request.eventLoop.makeSucceededFuture(hash.digest.hex)
+        } else {
+            return collectFile(at: path).map { buffer in
+                let digest = SHA256.hash(data: buffer.readableBytesView)
+
+                // update hash in dictionary
+                hashingDictionary[path] = FileMiddleware.ETagHashes.FileHash(lastModified: lastModified, digest: digest)
+                request.application.storage[FileMiddleware.ETagHashes.self] = hashingDictionary
```
0xTim, line 420: Is there a risk here we'll overwrite the cache with multiple concurrent requests? Would it not be better to just access it directly?

Tests/VaporTests/FileTests.swift
```diff
@@ -3,14 +3,15 @@ import XCTest
...
+import Crypto
 
 final class FileTests: XCTestCase {
     func testStreamFile() throws {
         let app = Application(.testing)
         defer { app.shutdown() }
 
-        app.get("file-stream") { req in
-            return req.fileio.streamFile(at: #file) { result in
+        app.get("file-stream") { req -> EventLoopFuture<Response> in
```
0xTim, line 13: Can you add a deprecated test that tests the old code just to ensure we don't break it?

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -215,10 +218,136 @@ public struct FileIO {
...
+
+        // Respond with lastModified header
+        headers.lastModified = HTTPHeaders.LastModified(value: modifiedAt)
+
+        var eTagFuture: EventLoopFuture<String>
+
+        if advancedETagComparison {
+            eTagFuture = generateETagHash(path: path, lastModified: modifiedAt)
+        } else {
+            // Generate ETag value, "HEX value of last modified date" + "-" + "file size"
```
0xTim, line 285: This comment seems not to match with the code, shouldn't it be epoch time instead of hex value?

0xTim (changes requested):
Sorry for the delay in reviewing this! I have a few questions now I've looked at this with a fresh set of eyes, but I think we're almost there

linus-hologram:
I will look at this as soon as I can. Pretty covered up in work now, but I'll get there!

Sources/Vapor/Utilities/FileIO.swift
linus-hologram: yes, there would be. thanks for pointing that out!

Sources/Vapor/Utilities/FileIO.swift
linus-hologram: correct, overlooked that one

Tests/VaporTests/FileTests.swift
linus-hologram: added a deprecated stream File test

linus-hologram:
@0xTim applied your feedback - please have a look.

I've noticed that the now deprecated `streamFile` function received the `@preprocessor` annotation. I haven't been following vapor's recent progress, should I add that annotation to the new `streamFile` function as well?

Cheer for the review 🎉

Tests/VaporTests/FileTests.swift
```diff
@@ -3,17 +3,39 @@ import XCTest
...
+        }
+
+        try app.testable(method: .running).test(.GET, "/file-stream") { res in
+            let test = "the quick brown fox"
+            XCTAssertNotNil(res.headers.first(name: .eTag))
+            XCTAssertContains(res.body.string, test)
+        }
+    }
+
+    func testLegacyStreamFile() throws {
```
0xTim, line 30: Does this produce a warning when built? If so you can add a deprecated annotation to the test to silence it

0xTim (commented):
One query then good to go I think

Tests/VaporTests/FileTests.swift
linus-hologram: yes, it does. Added deprecated annotation.

0xTim (approved):
LGTM - @gwynne any last thoughts?

mkll:
Guys, did I not understand something? We use `streamFile()` when we don’t want, for one reason or another (the main one being file size), to read the entire file into memory.

The new version of the `streamFile()` method, which appeared in release 4.93.0,, does exactly this (when the `advancedETagComparison` flag is turned on) — it reads the **entire file** into memory. Moreover, this is the default behavior of this method. I say "default" because the `advancedETagComparison` parameter has a default value of `true`.

So our `streamFile()` method magically becomes a `collectFile()` method, at least in terms of memory consumption.

Of course, I understand that in order to take a hash from the contents of a file, it needs to be read into memory, but I am afraid that this innovation “slightly” changes the **default** behavior of the `streamFile()` method and makes it unexpected.

I also understand that Vapor **must** be able to generate `ETag` in full accordance with the specification. But maybe it’s not worth, firstly, forcing it on everyone, and secondly, it was worth at least warning that if user need an `ETag`, then he will have a "magical" transformation of a `streamFile()` into a `collectFile()`?

It seems to me that it would be worth doing two things:
1. Do not assign a default value to the `advancedETagComparison` flag.
2. (maybe) Do not deprecate the previous variant of the `streamFile()` method.

mkll:
In addition, one more point needs to be mentioned. The new `streamFile()` method deprives me of the opportunity to create an `HTTPHeaders` myself. Here is an example of a working code snippet in which I generate the headers I need:
```
.flatMap { (zipPath, zipFileName) in
	var headers: HTTPHeaders = .init()
	headers.add(name: .contentType, value: "application/zip")
	headers.add(name: .contentDisposition, value: "attachment; filename=\"\(zipFileName)\"")
	
	guard FileManager.default.fileExists(atPath: zipPath) else {
		return req.eventLoop.makeFailedFuture(Abort(.internalServerError, reason: "ZIP archive not found (was not created?)"))
	}
	
	return req.fileio.streamFile(at: zipPath, onCompleted: { result in
		try? FileManager.default.removeItem(atPath: zipPath)
	}).encodeResponse(status: .ok, headers: headers, for: req)
}
```
It is quite obvious that with the new `streamFile()` method I will not be able to do this. Of course, I can add my own to the generated headers; there is an “addOrReplace” method (or something similar) for this, but I wouldn’t want to lose the opportunity to completely generate my own headers.

Therefore, I would not say that deprecation of the old method is generally possible. The new method is not an equivalent improved replacement, it has its own area of ​​use, it cannot replace the old one.

mkll:
Well, we can probably also mention that the new method imposes on me, as a developer, the use of a specific method of generating ETag, namely `SHA256`. Meanwhile, the specification does not regulate the method of generation; it only stipulates that the values of the ETag tag must be generated in a way that reflects the uniqueness of the file contents.

In my opinion, using the strong cryptographic function SHA256 for this purpose is redundant. I would use some non-cryptographic function (after all, we don’t need cryptographic strength here), for example, `xx_hash`.

I am not going to argue that it is necessary to use `xx_hash` by default, I am only demonstrating that there is more than one option, and the developer has the right to use the one that suits him best.

To summarize the above, I would say this: if I personally ever need to generate an ETag for the content being streamed, I will not use the new method for this because it does not suit me. I will write my own implementation of ETag generation, my own storage location, choose a convenient moment when the ETag will be generated and stored somewhere, create my own HTTP headers.

And for this I, among other things, will need an old method that is prematurely marked as deprecated.

linus-hologram:
> Guys, did I not understand something? We use `streamFile()` when we don’t want, for one reason or another (the main one being file size), to read the entire file into memory.
> 
> The new version of the `streamFile()` method, which appeared in release 4.93.0,, does exactly this (when the `advancedETagComparison` flag is turned on) — it reads the **entire file** into memory. Moreover, this is the default behavior of this method. I say "default" because the `advancedETagComparison` parameter has a default value of `true`.
> 
> So our `streamFile()` method magically becomes a `collectFile()` method, at least in terms of memory consumption.
> 
> Of course, I understand that in order to take a hash from the contents of a file, it needs to be read into memory, but I am afraid that this innovation “slightly” changes the **default** behavior of the `streamFile()` method and makes it unexpected.
> 
> I also understand that Vapor **must** be able to generate `ETag` in full accordance with the specification. But maybe it’s not worth, firstly, forcing it on everyone, and secondly, it was worth at least warning that if user need an `ETag`, then he will have a "magical" transformation of a `streamFile()` into a `collectFile()`?
> 
> It seems to me that it would be worth doing two things:
> 
> 1. Do not assign a default value to the `advancedETagComparison` flag.
> 2. (maybe) Do not deprecate the previous variant of the `streamFile()` method.

You are right in that the file is loaded into memory - however, just once. After the hash is generated, it is cached and future requests for that file will not load the entire file into memory. I agree with you though that this is different to the original `streamFile` implementation.

Additionally, could you please elaborate on the issue with the response headers? I believe you can still use `.encodeResponse(status: .ok, headers: yourHeaders, for: req)` even with the new `streamFile` method, no?

That being said, @0xTim what's your opinion on this?

I'm happy to go over it again :)

mkll:
> You are right in that the file is loaded into memory - however, just once.

Well, my server with 2GB of memory will not benefit from the fact that a 10GB file will be loaded into memory **only once**. 

To be honest, I am somewhat shocked by the “_you can’t do this, but if you only do it once, then you can_” approach. Nope. I disagree. If we can't, then we can't.

I would like to draw your attention to the most important and obvious point among everything: we use the `streamFile()` method precisely so as **not to load the entire file** into memory. Period.

We use it when we have reasons to do so. We intentionally use the `streamFile()` method rather than the `collectFile()` method.

Your edit negates the difference between the two.

If you need a more detailed explanation of my thoughts, here it is: 

There may be cases where we can neglect the fact that the entire file is loaded into memory. When can we neglect it? In cases where the file is small. But in these cases, most likely, we will not use the `streamFile()` method at all.

But when the `streamFile()` method is vital to us for one reason or another (the most obvious reason, the file is very large) and we use it, we are surprised to discover that the new “streaming” method ceases to be a "streaming", it's now a "collect in memory" method. Without warning, I'll note. And there is simply no point in using it.

And when it comes to really large files that we would not like to give to the client twice without any need, then in this case we would take some special actions to generate the ETag in advance, in the background or by a separate service values and place it in some store.

I assure you that even if my server has 128GB of RAM, I will never want to load a 10GB file into memory like this, on the fly, at the time of processing a request. This is completely unacceptable.

To be honest, I don't understand in what cases the new `streamFile()` method can be used. In mild cases it is not needed, and in severe cases its implementation simply does not meet the requirements for processing this type of content, since this requires an individual approach to each specific implementation.

The new version of the `streamFile()` method is suitable, it seems to me, for only one thing: if we are asked whether Vapor supports generating hashes in `ETag`, we can answer “**yes**” (but it’s better not to go into details :)).

mkll:
> Additionally, could you please elaborate on the issue with the response headers? I believe you can still use .encodeResponse(status: .ok, headers: yourHeaders, for: req) even with the new streamFile method, no?

Most likely not, but I'll check it a bit later, re-installing Xcode right now. :)

linus-hologram:
Discussed this with @gwynne and @0xTim and i'll be revising my code

mkll:
> Additionally, could you please elaborate on the issue with the response headers? I believe you can still use .encodeResponse(status: .ok, headers: yourHeaders, for: req) even with the new streamFile method, no?

@linus-hologram You are right, I checked, we can "stack" multiple `.encodeResponse()` methods one after the other.

mkll:
@linus-hologram I think that release 4.93.2 is optimal, providing a default solution, but at the same time leaving the developer the freedom to choose his own implementation option. Thanks!

PR #3016, opened 2023-05-15
Improved error logging

gwynne:
Some kinds of errors provide additional "debug" information, which can give much more detail than the "plain" description of the error. In many cases this debug info can contain sensitive data, such as specifics about a database schema, so Vapor only uses the plain description when sending errors to clients (and in release environments, _all_ details are suppressed).

To date, the plain description has also been used for logging errors. This can make it very difficult for developers to figure out what's going wrong with their code if the error in question only provides meaningful information in its debug data - for example, the PostgreSQL database driver implementation does this rather than relying on a higher-level layer like Vapor to obfuscate potentially sensitive information. This PR changes the logging of errors to include the debug information (and _only_ the logging; the responses sent to clients are unchanged).

### DIFF
--- a/.github/workflows/projectboard.yml
+++ b/.github/workflows/projectboard.yml
@@ -5,27 +5,7 @@ on:
     types: [reopened, closed, labeled, unlabeled, assigned, unassigned]
 
 jobs:
-  setup_matrix_input:
-    runs-on: ubuntu-latest
-
-    steps:
-      - id: set-matrix 
-        run: |
-          output=$(curl ${{ github.event.issue.url }}/labels | jq '.[] | .name') || output=""
-
-          echo '======================'
-          echo 'Process incoming data'
-          echo '======================'
-          json=$(echo $output | sed 's/"\s"/","/g')
-          echo $json
-          echo "::set-output name=matrix::$(echo $json)"
-    outputs:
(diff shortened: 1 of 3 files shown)


madsodgaard:
@gwynne Couldn't sensitive information contained stored in log files still be a concern in production? Not sure what "sensitive" information we are talking about here. Is it database connection details or something less concerning such as schema information?

0xTim (approved):

0xTim:
@madsodgaard this is only really converting the error to a better string - if the error contains sensitive information then yeah it will be logged, but in the same way that logging the whole request will leak sensitive info. We can't really police that as a framework and instead should provide best practices for logging and errors (i.e. errors should be fairly generic, log messages should be specific but either scrubbed from PII or using a custom backend combined with log metadata to ensure nothing is output

gwynne:
In this case I did the update pretty much specifically for the sake of `PSQLError` (as a companion to https://github.com/vapor/postgres-kit/pull/244 and/or https://github.com/vapor/postgres-nio/pull/360) - "potentially sensitive" data in that case means things like details of the database schema and entire queries, which can easily reveal exploitable info. The expectation is that sending the details to the request's logger will put the details somewhere that they can be seen by the server admin (which is potentially lifesaving in a production environment), while still keeping them from clients even in debug builds.

PR #3019, opened 2023-05-18
Avoid deadlocking websocket tests

0xTim:
Wait for the text rather than connection to avoid deadlocking

I'm half tempted to just remove this test anyway as it doesn't really do a lot


gwynne:
I concur with removing the test entirely.

0xTim:
@gwynne removed, want to review?

gwynne (approved):

PR #3023, opened 2023-05-25
add `ContentContainer.decode(_:as:)`

MahdiBM:
This came up in a conversation were both Tim and I were expecting such a function, but realized there it isn't there. 
The fact that we expected it is probably because there is a `ContentContainer.encode(_:as:)`, but no `decode(_:as:)`. 
This should be useful to users that are trying to decode content of a request with a bad/unexpected content-type header.

### DIFF
--- a/Tests/VaporTests/ContentTests.swift
+++ b/Tests/VaporTests/ContentTests.swift
@@ -84,7 +84,7 @@ final class ContentTests: XCTestCase {
         }
     }
 
-    func testContentContainer() throws {
+    func testContentContainerEncode() throws {
         struct FooContent: Content {
             var message: String = "hi"
         }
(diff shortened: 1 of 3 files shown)


MahdiBM:
(I guess my branch name isn't accurate. Should have been something like `mahdibm-add-as-contenttype`)

Tests/VaporTests/RequestTests.swift
```diff
@@ -261,13 +261,13 @@ final class RequestTests: XCTestCase {
         app.http.client.configuration.redirectConfiguration = .disallow
 
         app.get("redirect_normal") {
-            $0.redirect(to: "foo", type: .normal)
+            $0.redirect(to: "foo", redirectType: .normal)
```
0xTim, line 264: These need to be reverted. As the test name implies, we have this to ensure the old redirects work

0xTim (changes requested):

Tests/VaporTests/RequestTests.swift
MahdiBM: O smart 😅

Tests/VaporTests/ContentTests.swift
```diff
@@ -108,6 +108,31 @@ final class ContentTests: XCTestCase {
...
+        struct FooDecodable: Decodable, Equatable {
+            var message: String = "hi"
+        }
+
+        let app = Application(.testing)
+        defer { app.shutdown() }
+
+        app.routes.get("decode") { req async throws -> String in
+            XCTAssertEqual(try req.content.decode(FooContent.self), FooContent())
+            XCTAssertEqual(try req.content.decode(FooDecodable.self, as: .json), FooDecodable())
```
0xTim, line 125: I would like to see another test route (either new test or the same test) where you send a request body that's JSON but some random content-type and then ensure we can decode the JSON body. This matches the behaviour we want and problems people have hit

0xTim (approved):
Thanks!

0xTim:
Original comment:

> This came up in a conversation were both Tim and I were expecting such a function, but realized there it isn't there.
The fact that we expected it is probably because there is a ContentContainer.encode(_:as:), but no decode(_:as:).
This should be useful to users that are trying to decode content of a request with a bad/unexpected content-type header.

PR #3040, opened 2023-07-21
Remove unnecessary efficiency bottleneck from server channel configuration

gwynne:
We have been setting `ChannelOption.maxMessagesPerRead` to `1` throughout Vapor 4's lifetime, for no apparent reason. Let's stop doing that.

Thanks to @FranzBusch for pointing this out!


MahdiBM (approved):

gwynne:
Based on additional discussion and some extremely helpful clarification from @weissi (thanks, Johannes!!), for Vapor's use case it seems that `1` is a reasonable default. We should look into exposing this value through Vapor's server configuration.

t089:
Would it be possible to share some of the clarification? I always wondered how to best reason about and how to pick this value.

weissi:
> Would it be possible to share some of the clarification? I always wondered how to best reason about and how to pick this value.

Very roughly, smaller numbers (1 is the smallest allowable) allow better latency as fewer `read` syscalls in a row are attempted. Higher values allow theoretically better throughput because NIO can call `read` a bunch of times in a loop (pumping more data through the pipeline). But don't go nuts, if you were to set something suuper high, one channel that has a lot of input data might starve all others because we'd just keep `read`ing from it.

My advice would be: Stick to `1 ... 16`. Given that Vapor mostly gets small requests, `1` seems entirely reasonable.

PR #3043, opened 2023-07-26
Allow specifying a timeout for client requests

ffried:
Adds a `timeout` property to `ClientRequest`, which is forwarded to `HTTPClient` by `EventLoopHTTPClient`.
To use it, the `beforeSend` closure can be used:

```swift
request.client.get("http://example.com") {
    $0.timeout = .seconds(5)
}
```

### DIFF
--- a/Sources/Vapor/Client/ClientRequest.swift
+++ b/Sources/Vapor/Client/ClientRequest.swift
@@ -7,6 +7,7 @@ public struct ClientRequest {
     public var url: URI
     public var headers: HTTPHeaders
     public var body: ByteBuffer?
+    public var timeout: TimeAmount?
     private let byteBufferAllocator: ByteBufferAllocator
 
     public init(
@@ -14,12 +15,14 @@ public struct ClientRequest {
         url: URI = "/",
         headers: HTTPHeaders = [:],
         body: ByteBuffer? = nil,
+        timeout: TimeAmount? = nil,
         byteBufferAllocator: ByteBufferAllocator = ByteBufferAllocator()
     ) {
         self.method = method
         self.url = url
         self.headers = headers
         self.body = body
+        self.timeout = timeout
         self.byteBufferAllocator = byteBufferAllocator
     }
 }
--- a/Sources/Vapor/Client/ClientRequest.swift
+++ b/Sources/Vapor/Client/ClientRequest.swift
@@ -15,7 +15,7 @@ public struct ClientRequest {
         url: URI = "/",
         headers: HTTPHeaders = [:],
         body: ByteBuffer? = nil,
-        timeout: TimeAmount? = nil,
+        timeout: TimeAmount?,
         byteBufferAllocator: ByteBufferAllocator = ByteBufferAllocator()
     ) {
         self.method = method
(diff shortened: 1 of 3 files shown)


MahdiBM (approved):
Thanks, this looks good.

gwynne (approved):
Looks excellent to me, a very clean and tidy improvement!

Sources/Vapor/Client/ClientRequest.swift
```diff
@@ -7,19 +7,22 @@ public struct ClientRequest {
...
     private let byteBufferAllocator: ByteBufferAllocator
 
     public init(
         method: HTTPMethod = .GET,
         url: URI = "/",
         headers: HTTPHeaders = [:],
         body: ByteBuffer? = nil,
+        timeout: TimeAmount? = nil,
         byteBufferAllocator: ByteBufferAllocator = ByteBufferAllocator()
     ) {
```
gwynne, line 20: For 100% pedantic source compatibility, this should be written this way:
```suggestion
    public init(
        method: HTTPMethod = .GET,
        url: URI = "/",
        headers: HTTPHeaders = [:],
        body: ByteBuffer? = nil,
        byteBufferAllocator: ByteBufferAllocator = ByteBufferAllocator()
    ) {
        self.init(
            method: method,
            url: url,
            headers: headers,
            timeout: nil,
            body: body,
            byteBufferAllocator: byteBufferAllocator
        )
    }
    
    public init(
        method: HTTPMethod = .GET,
        url: URI = "/",
        headers: HTTPHeaders = [:],
        body: ByteBuffer? = nil,
        timeout: TimeAmount?,
        byteBufferAllocator: ByteBufferAllocator = ByteBufferAllocator()
    ) {
```

gwynne: (I'm not a fan of it personally, but unfortunately it does actually affect compatibility in unlikely but existing case >.<)

ffried: I'm aware of this. But I decided against it given that other semver-minor changes (like the one that introduced the `byteBufferAllocator`) also used default arguments.

But I can also change it to this form if this is now preferred.

gwynne: I've historically been extremely lax about it myself, but then I actually got bitten by the "referring to the changed method as a callable rather than actually calling it" edge case and I've started (**very** reluctantly) being more stern about it (hence https://github.com/vapor/postgres-kit/blob/main/Sources/PostgresKit/PostgresDatabase%2BSQL.swift#L6-L17, for example)

ffried: Yea, this is the only use-case I know where this actually matters.
It's fine, I'll change it real quick.
<hr/>

_Can someone please submit a Swift evolution proposal to allow referencing callables by omitting default argument values? 😝_

ffried: Done in [adc52e0](https://github.com/vapor/vapor/pull/3043/commits/adc52e0bea4aefa318a6d6d071320a41835247d0)

gwynne (approved):

PR #3054, opened 2023-08-10
Add Sendable Conformances to undelying types

0xTim:
This adds Sendable annotations to many of Vapor's types and APIs, including the request handlers.

APIs are marked `@proconcurrency` where it makes sense to reduce warnings for anyone using the release. To fix your code and see the warnings, you can enable strict concurrency checking.

**Note:** this PR does not touch `Request`, `Response` or `Application` as they require fundamental changes that are larger in scope and will be done in a future release.


Sources/Vapor/Error/StackTrace.swift
```diff
@@ -3,15 +3,28 @@ import Foundation
...
 
 extension Optional where Wrapped == StackTrace {
     public static func capture(skip: Int = 0) -> Self {
         StackTrace.capture(skip: 1 + skip)
     }
 }
 
-public struct StackTrace {
-    public static var isCaptureEnabled: Bool = true
+public struct StackTrace: Sendable {
```
gwynne, line 14: I hate that we even have this code 😕

Sources/Vapor/Exports.swift
```diff
@@ -1,17 +1,46 @@
...
+@_documentation(visibility: internal) @_exported import class NIO.EmbeddedChannel
+@_documentation(visibility: internal) @_exported import class NIO.EmbeddedEventLoop
+@_documentation(visibility: internal) @_exported import protocol NIO.EventLoop
+@_documentation(visibility: internal) @_exported import class NIO.EventLoopFuture
+@_documentation(visibility: internal) @_exported import protocol NIO.EventLoopGroup
+@_documentation(visibility: internal) @_exported import struct NIO.EventLoopPromise
+@_documentation(visibility: internal) @_exported import class NIO.MultiThreadedEventLoopGroup
+@_documentation(visibility: internal) @_exported import struct NIO.NonBlockingFileIO
+@_documentation(visibility: internal) @_exported import class NIO.NIOThreadPool
+@_documentation(visibility: internal) @_exported import enum NIO.System
```
gwynne, line 23:
```suggestion
@_documentation(visibility: internal) @_exported import struct NIOCore.ByteBuffer
@_documentation(visibility: internal) @_exported import struct NIOCore.ByteBufferAllocator
@_documentation(visibility: internal) @_exported import protocol NIOCore.Channel
@_documentation(visibility: internal) @_exported import class NIOCore.EmbeddedChannel
@_documentation(visibility: internal) @_exported import class NIOCore.EmbeddedEventLoop
@_documentation(visibility: internal) @_exported import protocol NIOCore.EventLoop
@_documentation(visibility: internal) @_exported import class NIOCore.EventLoopFuture
@_documentation(visibility: internal) @_exported import protocol NIOCore.EventLoopGroup
@_documentation(visibility: internal) @_exported import struct NIOCore.EventLoopPromise
@_documentation(visibility: internal) @_exported import class NIOCore.MultiThreadedEventLoopGroup
@_documentation(visibility: internal) @_exported import struct NIOCore.NonBlockingFileIO
@_documentation(visibility: internal) @_exported import class NIOCore.NIOThreadPool
@_documentation(visibility: internal) @_exported import enum NIOCore.System
```
Plus editing the imports in the `#else` clause of course.

Sources/Vapor/Utilities/DotEnv.swift
```diff
@@ -172,10 +172,11 @@ public struct DotEnvFile {
         on eventLoop: EventLoop
     ) -> EventLoopFuture<DotEnvFile> {
         return fileio.openFile(path: path, eventLoop: eventLoop).flatMap { arg -> EventLoopFuture<ByteBuffer> in
+            let fileHandleWrapper = NIOLoopBound(arg.0, eventLoop: eventLoop)
```
gwynne, line 175: Okay, I've seen this pattern with file handles enough times now to be seriously bothered. This doesn't feel at all correct - `FileHandle` isn't `Sendable` because it implies sharing file descriptors across threads, which is a recipe for truly epic amounts of disaster. I don't have a better suggestion, but I'd love it if someone like Fabian or Cory could weigh in 😕

Sources/Vapor/Utilities/DotEnv.swift
```diff
@@ -172,10 +172,11 @@ public struct DotEnvFile {
         on eventLoop: EventLoop
     ) -> EventLoopFuture<DotEnvFile> {
         return fileio.openFile(path: path, eventLoop: eventLoop).flatMap { arg -> EventLoopFuture<ByteBuffer> in
```
gwynne, line 174:
```suggestion
        return fileio.openFile(path: path, eventLoop: eventLoop).flatMapWithEventLoop { arg, eventLoop -> EventLoopFuture<ByteBuffer> in
```

Sources/Vapor/Utilities/OptionalType.swift
```diff
@@ -26,7 +26,7 @@ extension EventLoopFuture where Value: OptionalType {
 ///     let futureInt = maybeFutureInt ?? 0
 ///     print(futureInt) // Future<Int>
 ///
-public func ??<T>(lhs: EventLoopFuture<T?>, rhs: T) -> EventLoopFuture<T> {
+public func ??<T: Sendable>(lhs: EventLoopFuture<T?>, rhs: T) -> EventLoopFuture<T> {
```
gwynne, line 29: It's extremely unfortunate that this is `public`...

Sources/Vapor/Utilities/RFC1123.swift
```diff
@@ -106,13 +108,13 @@ internal final class RFC1123DateCache {
...
             dateComponents = cached.components
         } else {
             var tc = tm.init()
             gmtime_r(&date, &tc)
             dateComponents = tc
-            self.cachedTimeComponents = (key: key, components: tc)
+            self.cachedTimeComponents.withLockedValue { $0 = (key: key, components: tc) }
         }
         
         // parse components
```
gwynne, line 120: No. This is wrong - you leave a nice wide gap where the components are out of sync with the timestamp. They need to both be under the same lock.

gwynne (changes requested):

Sources/Vapor/Utilities/DotEnv.swift
0xTim: This was the suggestion - file handle isn't Sendable but wrapping it with this at least makes it safe to use (and gets rid of the warnings)

gwynne: Gets rid of the warnings, certainly. I'm not convinced it does a darn thing in terms of safety in either direction.

Sources/Vapor/Utilities/OptionalType.swift
0xTim: On the plus side this should be pretty out of use with async/await

Sources/Vapor/Utilities/DotEnv.swift
0xTim: What benefit does this give us over the current implementation given we're sticking to the same event loop?

Sources/Vapor/Error/StackTrace.swift
0xTim: We can probably wrap this in a 5.9 flag when that's out as that provides proper runtime backtrace support in the language

Sources/Vapor/Utilities/DotEnv.swift
0xTim: @fabianfett @FranzBusch any thoughts?

Sources/Vapor/Content/ContentConfiguration.swift
```diff
@@ -16,7 +17,17 @@ import MultipartKit
...
-    public static var global: ContentConfiguration = .default()
+    public static var global: ContentConfiguration {
+        get {
+            _global.withLockedValue { $0 }
+        }
+        set {
+            _global.withLockedValue {
+                $0 = newValue
+            }
+        }
```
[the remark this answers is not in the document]
0xTim: The compiler is having issues with this
```
'&' may only be used to pass an argument to inout parameter
Cannot find 'yield' in scope
```

Sources/Development/routes.swift
```diff
@@ -268,6 +272,7 @@ public func routes(_ app: Application) throws {
         return [cred1]
     }
     
+    @Sendable
```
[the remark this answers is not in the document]
0xTim: Essentially because it's required at the low level by NIO in `completeWithTask`, which means that the route builders are required to be `Sendable` so routes should be `Sendable` hence the annotation. The `@preconcurrency` annotation removes it for most users unless they have strict concurrency checking on

Package.swift
```diff
@@ -131,3 +132,11 @@ let package = Package(
...
     ]
 )
+
+if ProcessInfo.processInfo.environment["STRICT_CONCURRENCY"] == "true" {
+    for target in package.targets {
+        if !target.isTest {
+            target.swiftSettings = [.unsafeFlags(["-Xfrontend", "-strict-concurrency=complete"])]
+        }
+    }
+}
```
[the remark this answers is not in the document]
0xTim: We can only do this if we set the tools version to 5.9 for the whole package

Sources/Vapor/Utilities/DotEnv.swift
gwynne: Fewer captures is always a good thing - especially when you remember that `EventLoop` is actually `any EventLoop` (existentials suck).

Sources/Vapor/Content/ContentConfiguration.swift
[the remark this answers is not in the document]
gwynne: Ah, I'm guessing the special `_modify` syntax only works within the same scope... Try this:
```swift
_modify {
    self._globalLock.lock()
    defer { self._globalLock.unlock() }
    yield &self._global
}
```
(Obviously you'll have to use a plain `NIOLock` rather than a box for this)

Package.swift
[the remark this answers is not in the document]
0xTim: Ok separate manifests it is

Sources/Vapor/Utilities/DotEnv.swift
Lukasa: `NIOLoopBound` makes its `Sendable` contract work by runtime checks: if you create a `NIOLoopBound` from off the `eventLoop` you've bound it to, or use it from off that `eventLoop`, you'll get a runtime crash.

gwynne: Nothing I didn't already know, but good enough just the same 🙂

Package@swift-5.7.swift
gwynne, line 1: @0xTim The pattern I'm using in `MySQLNIO` is to keep `Package.swift` at tools-version 5.7 and have `Package@swift-5.8.swift` adding the use of `.enableUpcomingFeature()`. (I also have an additional `Package@swift-5.9.swift` to further reference the couple of macros I've tinkered with, but that'll probably go away before I PR the lot.) I'm not actually clear on whether it's more idiomatic to put the highest or the lowest version in the base manifest, though. Definitely wouldn't mind getting @neonichu's thoughts on that one.

gwynne (approved):
Tentatively approved, but FWIW I can't really claim to be happy with it yet.

Package@swift-5.9.swift
gwynne, line 1: @0xTim The pattern I'm using in `MySQLNIO` is to keep `Package.swift` at tools-version 5.7 and have `Package@swift-5.8.swift` adding the use of `.enableUpcomingFeature()`. (I also have an additional `Package@swift-5.9.swift` to further reference the couple of macros I've tinkered with, but that'll probably go away before I PR the lot.) Basically, what you've done here. I'm not actually clear on whether it's more idiomatic to put the highest or the lowest version in the base manifest, though. Definitely wouldn't mind getting @neonichu's thoughts on that one.

Sources/Development/routes.swift
```diff
@@ -244,6 +246,7 @@ public func routes(_ app: Application) throws {
         return String(buffer: body)
     }
 
+    @Sendable
```
FranzBusch, line 249: Nothing here should ever use `@Sendable` on a method. This is highly unlikely to be the right thing ever. `@Sendable` on a method just disables all sendable checking at all. Rather the type itself needs to become `Sendable` here.

Sources/Vapor/Content/ContentConfiguration.swift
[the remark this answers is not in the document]
FranzBusch: Do we actually want people to be able to modify the global configuration here? Otherwise we could just trap on the `set` path and disallow this.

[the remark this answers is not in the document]
gwynne: Unfortunately, yes, Vapor is stuck with this very broken API pattern where modifying the global configuration is the "correct" behavior. Yet another fun consequence of Tanner's designs and semver rules.

Sources/Development/routes.swift
gwynne: Thank you, I knew this felt wrong 😰

Package.swift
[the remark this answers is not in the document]
gwynne: Since you've done that, please remove the changes to the original manifest and the test workflow

Package@swift-5.9.swift
0xTim: Yeah I switched to that. Having the highest version in the base manifest seemed to break various CI tests, which seems like a SwiftPM bug but I didn't have a chance to investigate

Sources/Development/routes.swift
0xTim: @FranzBusch hmm even making `Request` `@unchecked Sendable` just to try it out results in
```swift
Converting non-sendable function value to '@Sendable (Request) async throws -> String' may introduce data races
```
any idea?

FranzBusch: You have to keep in mind that `self` is also a parameter for this method so you gotta make sure that `Self` is also `Sendable` in this case probably `Creds`

0xTim: `Self` here is top level code, it's a function not in any type. Is that going to cause issues?

0xTim: Just tried putting the routes in a type that's `Sendable` and no changes

Sources/Vapor/Content/ContentConfiguration.swift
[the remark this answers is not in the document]
gwynne: The compiler doesn't synthesize the efficient version, considering the lock, but in retrospect that's a silly consideration for `ContentConfiguration` anyway, I withdraw this particular suggestion 😅

Sources/Development/routes.swift
FranzBusch: Ah wait sorry I was mis reading the surrounding context in the diff. Yeah this is a known compiler deficiency. I recommend changing these local methods to be local closures. You can then write them like this

```
let asyncRouteTester: @Sendable (Request) async throws -> String = {
}
```

This will make the compiler properly check the closure for `@Sendable` and also solve the original warning that you got.

0xTim: @FranzBusch Is there a bug on GH for this or plans to solve this? To give you some context this file is just for trying out Vapor's APIs so rewriting the functions to closures here is fine but then we'd be telling everyone that every single route handler would need to be changed from a function to a closure

0xTim: PS it's not a major issue at the moment as the APIs are marked with `@preconcurrency` so you only get the warnings if you manually turn on strict concurrency checking

FranzBusch: It is not about people getting warnings when using `@Sendable` on a method but rather that the compiler is not checking your code if it really is `Sendable`. There is a pitch and some discussion around this topic here: https://forums.swift.org/t/pitch-inferring-sendable-for-methods/66565

.github/workflows/test.yml
```diff
@@ -42,8 +42,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    container: swiftlang/swift:nightly-5.9-jammy@sha256:a3c3ec5e4436c14e44759e38416bb0d46813dc6dd050e787ef3a80b2c3051a36
+    steps: 
+      - name: Check out Vapor
+        uses: actions/checkout@v3
+      - name: Run tests
+        env:
+            SWIFT_DETERMINISTIC_HASHING: 1
+        run: swift test
+
```
gwynne, line 55: This is useless; we already run against the latest 5.9 nightly in the reusable workflow.

gwynne (changes requested):
One last nit

.github/workflows/test.yml
0xTim: That's a good point actually. The check doesn't really do much yet anyway since we don't have warnings as errors turned on (and I'm doubtful if we ever will be able to with strict concurrency checking in Vapor 4)

gwynne (approved):

PR #3055, opened 2023-08-10
Make Response Sendable

0xTim:
Makes `Response` `Sendable` as well as the types that it uses.

> **Note:** `Storage` and therefore anything put into `Storage` is now annotated as `Storage`. For most things, this should be fine but for anything that can't be `Sendable` like Fluent's `Model` you may need to either wrap these in a `NIOLoopBound` or find another solution

### DIFF
--- a/Sources/Vapor/HTTP/BodyStream.swift
+++ b/Sources/Vapor/HTTP/BodyStream.swift
@@ -1,6 +1,6 @@
 import NIOCore
 
-public enum BodyStreamResult {
+public enum BodyStreamResult: Sendable {
     /// A normal data chunk.
     /// There will be 0 or more of these.
     case buffer(ByteBuffer)
@@ -39,7 +39,7 @@ extension BodyStreamResult: CustomDebugStringConvertible {
     }
 }
 
-public protocol BodyStreamWriter {
+public protocol BodyStreamWriter: Sendable {
     var eventLoop: EventLoop { get }
     func write(_ result: BodyStreamResult, promise: EventLoopPromise<Void>?)
 }
--- a/Sources/Vapor/HTTP/BodyStream.swift
+++ b/Sources/Vapor/HTTP/BodyStream.swift
@@ -39,7 +39,7 @@ extension BodyStreamResult: CustomDebugStringConvertible {
     }
 }
 
-public protocol BodyStreamWriter: Sendable {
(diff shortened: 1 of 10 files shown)


0xTim:
This should be merged after #3054

Sources/Vapor/HTTP/BodyStream.swift
```diff
@@ -46,8 +46,18 @@ public protocol BodyStreamWriter {
...
+            self.write(result, promise: promise)
+            return promise.futureResult
+        }
+        if self.eventLoop.inEventLoop {
+            return write0()
+        } else {
+            return self.eventLoop.flatSubmit {
+                return write0()
+            }
+        }
```
gwynne, line 61: Ditto on "Trash all this; I have a fully rewritten state-machine-based properly Sendable version locally that I'll PR shortly" 😅

Sources/Vapor/Response/Response.swift
```diff
@@ -1,26 +1,48 @@
...
     /// when the `body` property is mutated.
-    public var headers: HTTPHeaders
+    public var headers: HTTPHeaders {
+        get {
+            return _headers.withLockedValue { $0 }
+        }
+        set {
+            _headers.withLockedValue { $0 = newValue }
+        }
+    }
```
gwynne, line 45: It's not _technically_ necessary, but for cleanliness' sake, can we put these three under a single lock, given how closely related to each other they are?

Sources/Vapor/Response/Response.swift
```diff
@@ -29,17 +51,37 @@ public final class Response: CustomStringConvertible {
...
     /// represents the `Body`.
     public var body: Body {
-        didSet { self.headers.updateContentLength(self.body.count) }
+        get {
+            return _body.withLockedValue { $0 }
+        }
+        set {
+            _body.withLockedValue { $0 = newValue }
+            self.headers.updateContentLength(newValue.count)
+        }
```
gwynne, line 60: Not correct unless you take the headers lock as well during both get and set (that's "take the headers lock _AND_ the body lock" not "take them in sequence")

gwynne (changes requested):
There's probably more, this is just off the top of my head

0xTim:
Implemented in #3082

PR #3056, opened 2023-08-19
Make Storage Sendable

0xTim:
Make Vapor's `Storage` and `Application` `Sendable`. This is a key piece in allowing `Request` and `Response` to be `Sendable`.

Note that types that are stored in `Storage` should now be `Sendable`. If you get any warnings, you may need to update your code for this new change.

### DIFF
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -174,15 +174,14 @@ public final class Application: Sendable {
     }
 
     public func boot() throws {
-        self.isBooted.withLockedValue { booted in
+        try self.isBooted.withLockedValue { booted in
             guard !booted else {
                 return
             }
             booted = true
+            try self.lifecycle.handlers.forEach { try $0.willBoot(self) }
+            try self.lifecycle.handlers.forEach { try $0.didBoot(self) }
         }
-        
-        try self.lifecycle.handlers.forEach { try $0.willBoot(self) }
-        try self.lifecycle.handlers.forEach { try $0.didBoot(self) }
     }
     
     public func shutdown() {
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -6,16 +6,41 @@ import ConsoleKit
 import NIOPosix
 
 /// Core type representing a Vapor application.
-public final class Application {
-    public var environment: Environment
-    public let eventLoopGroupProvider: EventLoopGroupProvider
-    public let eventLoopGroup: EventLoopGroup
-    public var storage: Storage
-    public private(set) var didShutdown: Bool
-    public var logger: Logger
-    var isBooted: Bool
+public final class Application: Sendable {
+    public var environment: Environment {
+        get {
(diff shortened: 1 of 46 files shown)


0xTim:
Need to be careful here on how the Sendable stuff changes things like Authentication in Fluent

0xTim:
@gwynne @franzbusch @fabianfett this is ready for review if anyone would like to take a look

Sources/Vapor/Application.swift
```diff
@@ -119,10 +174,13 @@ public final class Application {
     }
 
     public func boot() throws {
-        guard !self.isBooted else {
-            return
+        self.isBooted.withLockedValue { booted in
+            guard !booted else {
+                return
+            }
```
gwynne, line 180: This will not work as intended. I assume I don't have to explain why 😅

Sources/Vapor/HTTP/EndpointCache.swift
```diff
@@ -134,8 +142,10 @@ public final class EndpointCache<T> where T: Decodable {
                     return eventLoop.makeFailedFuture(EndpointCacheError.contentDecodeFailure(error))
                 }
 
-                if self.cacheUntil != nil {
-                    self.cached = data
+                self.cacheUntil.withLockedValue { cachedDate in
+                    if cachedDate != nil {
+                        self.cached.withLockedValue { $0 = data }
+                    }
```
gwynne, line 148: If you have to nest locks like this, that's a sign that both items should be guarded by the same lock. In general I think you're making your locks too fine-grained because `NIOLockedValueBox` makes it so easy. An actor's properties are all guarded by the same task, for example.

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -354,7 +357,9 @@ private final class HTTPServerConnection {
             }
             
             // Set the handlers that are applied to the accepted Channels
-            .childChannelInitializer { [weak application] channel in
+            .childChannelInitializer { channel in
+//            .childChannelInitializer { [weak application] channel in
+#warning("We were storing a weak reference to the applicaiton at this point - why")
```
gwynne, line 362: Because otherwise you'd have a retain cycle between the application and the channel handlers.

gwynne (changes requested):

Sources/Vapor/Application.swift
0xTim: 🤦

Sources/Vapor/HTTP/EndpointCache.swift
0xTim: Tuples for the win!

Sources/Vapor/HTTP/Server/HTTPServer.swift
0xTim: I'm not overtly happy with having a crash point here but I don't think there's any other way to recover

gwynne (approved):

PR #3057, opened 2023-08-30
More Sendable Conformances

0xTim:
This adds more Sendable conformances to Vapor and resolves any Sendable warnings that aren't related to `Request` or `Response`


0xTim:
@fabianfett @franzbusch if you'd like to have a look that would be great! This has more changes than simply wrapping properties in `NIOLockedValueBox` as it touches parts of NIO that can't be `Sendable`

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
```diff
@@ -81,15 +82,15 @@ final class HTTPServerResponseEncoder: ChannelOutboundHandler, RemovableChannelH
...
-    let context: ChannelHandlerContext
-    let handler: HTTPServerResponseEncoder
+    let contextBox: NIOLoopBound<ChannelHandlerContext>
+    let handlerBox: NIOLoopBound<HTTPServerResponseEncoder>
     let promise: EventLoopPromise<Void>?
     let count: Int?
-    var currentCount: Int
-    var isComplete: Bool
+    let currentCount: NIOLockedValueBox<Int>
+    let isComplete: NIOLockedValueBox<Bool>
```
fabianfett, line 90: why not use ManagedAtomics here?

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
```diff
@@ -81,15 +82,15 @@ final class HTTPServerResponseEncoder: ChannelOutboundHandler, RemovableChannelH
...
 
     var eventLoop: EventLoop {
-        return self.context.eventLoop
+        return self.contextBox.value.eventLoop
     }
```
fabianfett, line 94: trick17: pass `the eventLoop` in the init and then store it as a let

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
```diff
@@ -103,41 +104,45 @@ private final class ChannelResponseBodyStream: BodyStreamWriter {
...
         case .end:
-            self.isComplete = true
-            if let count = self.count, self.currentCount != count {
-                self.promise?.fail(Error.notEnoughBytes)
-                promise?.fail(Error.notEnoughBytes)
+            self.isComplete.withLockedValue { $0 = true }
+            self.currentCount.withLockedValue {
+                if let count = self.count, $0 != count {
+                    self.promise?.fail(Error.notEnoughBytes)
+                    promise?.fail(Error.notEnoughBytes)
```
fabianfett, line 131: same as above.

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
fabianfett: are those only changed on EL? If that's the case, why do they need a Lock?

Sources/Vapor/Response/Response+Body.swift
```diff
@@ -174,24 +175,27 @@ extension Response {
 }
 
 private final class ResponseBodyCollector: BodyStreamWriter {
-    var buffer: ByteBuffer
-    var eventLoop: EventLoop
-    var promise: EventLoopPromise<ByteBuffer>
+    let buffer: NIOLockedValueBox<ByteBuffer>
+    let eventLoopBox: NIOLockedValueBox<EventLoop>
+    let promise: NIOLockedValueBox<EventLoopPromise<ByteBuffer>>
```
fabianfett, line 180: Why do you need locks here? Those are `let`s...

0xTim: They weren't before the were wrapped in `NIOLockedValueBox` - we mutate them inside the class, hence the need for a lock

Sources/Vapor/Response/Response+Body.swift
```diff
@@ -174,24 +175,27 @@ extension Response {
...
+    var eventLoop: EventLoop {
+        self.eventLoopBox.withLockedValue { $0 }
+    }
```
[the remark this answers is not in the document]
0xTim: Yeah I don't know why this was mutable to start with in hindsight, it doesn't look like we mutate it anywhere

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
0xTim: Inside the stream, they should only be changed on the event loop. We could use `LoopBound` 👍

0xTim: I hadn't actually considered ManagedAtomics

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
```diff
@@ -103,41 +104,45 @@ private final class ChannelResponseBodyStream: BodyStreamWriter {
...
-            self.currentCount += buffer.readableBytes
-            if let count = self.count, self.currentCount > count {
-                self.promise?.fail(Error.tooManyBytes)
-                promise?.fail(Error.notEnoughBytes)
+            self.contextBox.value.writeAndFlush(self.handlerBox.value.wrapOutboundOut(.body(.byteBuffer(buffer))), promise: promise)
+            self.currentCount.withLockedValue {
+                $0 += buffer.readableBytes
+                if let count = self.count, $0 > count {
+                    self.promise?.fail(Error.tooManyBytes)
+                    promise?.fail(Error.notEnoughBytes)
```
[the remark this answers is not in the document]
0xTim: No longer an issue if we moved to managed atomics

Tests/VaporTests/ApplicationTests.swift
```diff
@@ -23,26 +24,26 @@ final class ApplicationTests: XCTestCase {
 
     func testLifecycleHandler() throws {
         final class Foo: LifecycleHandler {
-            var willBootFlag: Bool
-            var didBootFlag: Bool
-            var shutdownFlag: Bool
+            let willBootFlag: NIOLockedValueBox<Bool>
+            let didBootFlag: NIOLockedValueBox<Bool>
+            let shutdownFlag: NIOLockedValueBox<Bool>
```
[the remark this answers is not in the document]
0xTim: Nice idea 👍

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
```diff
@@ -16,9 +16,10 @@ final class HTTPServerHandler: ChannelInboundHandler, RemovableChannelHandler {
     }
     
     func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        let box = NIOLoopBound((context, self), eventLoop: context.eventLoop)
         let request = self.unwrapInboundIn(data)
         self.responder.respond(to: request).whenComplete { response in
-            self.serialize(response, for: request, context: context)
+            box.value.1.serialize(response, for: request, context: box.value.0)
```
fabianfett, line 22: this is likely still two checks? Why not just get the value once and then do the split after that?

Tests/VaporTests/ApplicationTests.swift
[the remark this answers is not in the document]
fabianfett: will you fix it?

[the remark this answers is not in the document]
0xTim: Ah missed this. No I decided against this because we check the state for the 3 different properties independently (e.g. when `didBoot` is called, `willBoot` should also have been called. The state machine for lifecycle stuff doesn't have a way to distinguish this without an enum with lots of different values. Since it's just tests I stuck it with 3 locks

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
```diff
@@ -103,41 +102,42 @@ private final class ChannelResponseBodyStream: BodyStreamWriter {
...
+            self.contextBox.value.writeAndFlush(self.handlerBox.value.wrapOutboundOut(.body(.byteBuffer(buffer))), promise: promise)
+            self.currentCount.wrappingIncrement(by: buffer.readableBytes, ordering: .sequentiallyConsistent)
+            if let count = self.count, self.currentCount.load(ordering: .sequentiallyConsistent) > count {
                 self.promise?.fail(Error.tooManyBytes)
                 promise?.fail(Error.notEnoughBytes)
             }
         case .end:
-            self.isComplete = true
-            if let count = self.count, self.currentCount != count {
+            self.isComplete.store(true, ordering: .relaxed)
```
gwynne, line 124: Why relaxed ordering on this but not elsewhere?

Sources/Vapor/Request/Request+BodyStream.swift
```diff
@@ -6,27 +6,35 @@ extension Request {
...
+        struct HandlerBufferContainer: @unchecked Sendable {
+            var handler: ((BodyStreamResult, EventLoopPromise<Void>?) -> ())?
+            var buffer: [(BodyStreamResult, EventLoopPromise<Void>?)]
         }
 
-        private(set) var isClosed: Bool
-        private var handler: ((BodyStreamResult, EventLoopPromise<Void>?) -> ())?
-        private var buffer: [(BodyStreamResult, EventLoopPromise<Void>?)]
+        private let isClosed: NIOLockedValueBox<Bool>
+        private let handlerBuffer: NIOLoopBoundBox<HandlerBufferContainer>
```
gwynne, line 20: The properties are either loop-bound or they're locked, why would they be both?

Sources/Vapor/Response/Response+Body.swift
```diff
@@ -174,24 +175,24 @@ extension Response {
 }
 
 private final class ResponseBodyCollector: BodyStreamWriter {
-    var buffer: ByteBuffer
-    var eventLoop: EventLoop
-    var promise: EventLoopPromise<ByteBuffer>
+    let buffer: NIOLockedValueBox<ByteBuffer>
```
gwynne, line 178: This can and should be loop-bound instead of locked; see also below...

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
```diff
@@ -103,41 +102,42 @@ private final class ChannelResponseBodyStream: BodyStreamWriter {
...
     
     func write(_ result: BodyStreamResult, promise: EventLoopPromise<Void>?) {
         switch result {
         case .buffer(let buffer):
-            self.context.writeAndFlush(self.handler.wrapOutboundOut(.body(.byteBuffer(buffer))), promise: promise)
-            self.currentCount += buffer.readableBytes
-            if let count = self.count, self.currentCount > count {
+            self.contextBox.value.writeAndFlush(self.handlerBox.value.wrapOutboundOut(.body(.byteBuffer(buffer))), promise: promise)
+            self.currentCount.wrappingIncrement(by: buffer.readableBytes, ordering: .sequentiallyConsistent)
+            if let count = self.count, self.currentCount.load(ordering: .sequentiallyConsistent) > count {
```
gwynne, line 119: Why not `self.currentCount.wrappingIncrementThenLoad(by: buffer.readableBytes, ordering: .sequentiallyConsistent)`?

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
```diff
@@ -103,41 +102,42 @@ private final class ChannelResponseBodyStream: BodyStreamWriter {
...
                 self.promise?.fail(Error.tooManyBytes)
                 promise?.fail(Error.notEnoughBytes)
             }
         case .end:
-            self.isComplete = true
-            if let count = self.count, self.currentCount != count {
+            self.isComplete.store(true, ordering: .relaxed)
+            if let count = self.count, self.currentCount.load(ordering: .sequentiallyConsistent) != count {
```
gwynne, line 125:
```suggestion
            if let count = self.count, self.currentCount.load(ordering: .sequentiallyConsistent) < count {
```
Preexisting semantic bug - "too many bytes" is not the same as "not enough bytes".

Sources/Vapor/Response/Response+Body.swift
```diff
@@ -174,24 +175,24 @@ extension Response {
...
-            self.buffer.writeBuffer(&buffer)
+            _ = self.buffer.withLockedValue { $0.writeBuffer(&buffer) }
         case .error(let error):
             self.promise.fail(error)
         case .end:
-            self.promise.succeed(self.buffer)
+            self.promise.succeed(self.buffer.withLockedValue { $0 })
         }
         promise?.succeed(())
     }
```
gwynne, line 198: Borrowed slightly from my much more heavily revised version:
```swift
    func write(_ result: BodyStreamResult, promise: EventLoopPromise<Void>?) {
        let future = self.eventLoop.submit {
            switch result {
            case .buffer(var buffer):
                self.buffer.writeBuffer(&buffer)
            case .error(let error):
                self.promise.fail(error)
                throw error
            case .end:
                self.promise.succeed(self.buffer)
            }
        }
        // N.B.: Preexisting semantic bug: Errors in the stream should fail the individual write promise.
        if let promise { future.cascade(to: promise) }
    }
```

gwynne (changes requested):
Definitely shaping up, but several additional nits.

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
0xTim: Because I wasn't reading it elsewhere, I'll switch to sequential

Sources/Vapor/Request/Request+BodyStream.swift
0xTim: Sendable types are protected by locks, anything that isn't sendable needs to go in a `LoopBoundBox`. I haven't got a clear answer on which is the most performant

Sources/Vapor/Response/Response+Body.swift
0xTim: It can't be unfortunately. There are instances where the `BodyStreamWriter` is created off an event loop. If we tried to use `LoopBound` at this point, we'd fail the preconditions

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
0xTim: I didn't know the API existed 😅

Sources/Vapor/HTTP/Server/HTTPServerResponseEncoder.swift
0xTim: Ok I need to dig into this, because we should probably have a test for it

Sources/Vapor/Response/Response+Body.swift
0xTim: Do you have a corresponding test I could pull in for this?

gwynne: I wish; I never got that far, unfortunately 😕.

Sources/Vapor/Response/Response+Body.swift
gwynne: You can work around that with `NIOLoopBoundBox.makeEmptyBox()`

Sources/Vapor/Request/Request+BodyStream.swift
gwynne: - `NIOLockedValueBox` is as performant as the allocation plus a standard `pthread_mutex_t`. 
- `NIOLoopBound[Box]` is as performant as the cost of `EventLoop.inEventLoop`, which given a `MTELG` with Glibc is `pthread_equal()` + `pthread_self()`, where `pthread_equal()` is a single `==` and `pthread_self()` is a simple register load on both amd64 and arm64), plus the cost of allocation if using the box.

Given that a `pthread_mutex_t` can cost anywhere from a single atomic-CAS operation to an arbitrarily long wait time, I'd say it's a safe bet the loop-bound is less expensive.

Sources/Vapor/Response/Response+Body.swift
0xTim: Will fix this is a future PR

Sources/Vapor/Response/Response+Body.swift
```diff
@@ -186,14 +186,18 @@ private final class ResponseBodyCollector: BodyStreamWriter {
...
-        switch result {
-        case .buffer(var buffer):
-            _ = self.buffer.withLockedValue { $0.writeBuffer(&buffer) }
-        case .error(let error):
-            self.promise.fail(error)
-        case .end:
-            self.promise.succeed(self.buffer.withLockedValue { $0 })
-        }
-        promise?.succeed(())
+        let future = self.eventLoop.submit {
```
gwynne, line 189: Since the execution of this logic is now event loop-bound, does the buffer still need to be locked?

0xTim: Technically no, we can make this unchecked

Package.swift
```diff
@@ -59,6 +59,9 @@ let package = Package(
         
         // MultipartKit, Multipart encoding and decoding
         .package(url: "https://github.com/vapor/multipart-kit.git", from: "4.2.1"),
+        
+        // Low-level atomic operations
+        .package(url: "https://github.com/apple/swift-atomics.git", from: "1.1.0"),
```
gwynne, line 64: Add to `Package@swift-5.9.swift` also.

Package.swift
```diff
@@ -90,6 +93,7 @@ let package = Package(
             .product(name: "RoutingKit", package: "routing-kit"),
             .product(name: "WebSocketKit", package: "websocket-kit"),
             .product(name: "MultipartKit", package: "multipart-kit"),
+            .product(name: "Atomics", package: "swift-atomics"),
```
gwynne, line 96: Add to `Package@swift-5.9.swift`

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -88,7 +89,7 @@ public struct FileIO {
     public func readFile(
         at path: String,
         chunkSize: Int = NonBlockingFileIO.defaultChunkSize,
-        onRead: @escaping (ByteBuffer) -> EventLoopFuture<Void>
+        onRead: @Sendable @escaping (ByteBuffer) -> EventLoopFuture<Void>
```
gwynne, line 92: Am I completely off my rocker, or wouldn't it be valid to just _NOT_ make this `@Sendable`? Per SE-0302, `@Sendable` means "this function type is safe to pass across concurrency domains" - except this is the ELF-based API, which means there are no concurrency domains to cross here. And if the closure isn't sendable, we no longer have to waste a bunch of time acquiring and releasing a lock over and over for the buffer in `collectFile()`.

(And, obviously, if my understanding here is correct then the same "let's only use the attribute when it's accurate to use it" logic should be applied to the entire PR. Too bad we can't remove it from anywhere that already has it in released code, though...)

Tests/VaporTests/Utilities/CapturingMetricsSystem.swift
```diff
@@ -103,7 +103,7 @@ internal class TestCounter: CounterHandler, Equatable {
     }
 }
 
-internal class TestRecorder: RecorderHandler, Equatable {
+internal class TestRecorder: RecorderHandler, Equatable, @unchecked Sendable {
```
gwynne, line 106: If we're gonna made `TestTimer` `final` (per below) then do it for this too.

Tests/VaporTests/Utilities/CapturingMetricsSystem.swift
```diff
@@ -72,7 +72,7 @@ internal final class CapturingMetricsSystem: MetricsFactory {
     }
 }
 
-internal class TestCounter: CounterHandler, Equatable {
+internal class TestCounter: CounterHandler, Equatable, @unchecked Sendable {
```
gwynne, line 75: If we're gonna made `TestTimer` `final` (per below) then do it for this too.

gwynne (changes requested):
A couple bits and pieces, and one "hey if this is correct (and you should make sure I'm right before doing anything about it) it blows the whole PR out of the water" concern 😅

Sources/Vapor/Utilities/FileIO.swift
0xTim: > which means there are no concurrency domains to cross here

I think this isn't entirely correct. The caller of the function could create the function in one domain (such as an async route handler) that's then passed to another concurrency domain when we jump on to the event loop to actually call it so I think it is required.

(The eventual callsite of this where we pass the function is also marked as Sendable - https://github.com/apple/swift-nio/blob/main/Sources/NIOPosix/NonBlockingFileIO.swift#L82 - which also required the closure passed in to be Sendable I believe)

gwynne (approved):
I can't claim I find this PR to be a masterpiece of elegance and precision, but preconcurrency is preconcurrency, and as far as I can tell there aren't any actual _issues_ left.

PR #3064, opened 2023-09-06
The swift-backtrace library is no longer needed in 5.9

gwynne:
Per https://github.com/swift-server/swift-backtrace/pull/68.

### DIFF
--- a/Sources/Vapor/Error/StackTrace.swift
+++ b/Sources/Vapor/Error/StackTrace.swift
@@ -1,123 +1,37 @@
 import Foundation
-#if (swift(<5.9) && os(Linux)) || os(Windows)
-import Backtrace
-import CBacktrace
-#endif
 import NIOConcurrencyHelpers
 
+@available(*, deprecated, message: "Captured stack traces are no longer supported by Vapor")
 extension Optional where Wrapped == StackTrace {
-    public static func capture(skip: Int = 0) -> Self {
-        StackTrace.capture(skip: 1 + skip)
-    }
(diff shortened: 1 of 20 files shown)


Joannis:
Isn't this API breaking for users moving to 5.9? If they rely on Backtrace themselves, through this package, this could lead to confusing situations.

gwynne:
We (mercifully) don't have an `@_exported` line for this one, by some great virtue, so it's not part of Vapor's API. Relying on it as a transitive dependency if it's being used directly is incorrect usage.

Sources/Vapor/Error/StackTrace.swift
```diff
@@ -34,12 +34,12 @@ public struct StackTrace: Sendable {
         return .init(rawFrames: .init(frames))
     }
 
-    #if os(Linux)
+    #if swift(<5.9) || os(Windows)
     private static let state = backtrace_create_state(CommandLine.arguments[0], /* supportThreading: */ 1, nil, nil)
     #endif
 
     static func captureRaw() -> [RawFrame] {
```
0xTim, line 41: Do we need anything from this function _at all_ any more? I know we need to maintain the API, but `capture(skip:)` could just return `nil` in 5.9 which is non-breaking

0xTim (approved):

gwynne:
> Much nicer! But let's remove all the commented out code

Done 😉

0xTim (approved):

gwynne:
Bypassing the usual incorrectness of the API breakage checker.

PR #3068, opened 2023-09-19
`Logger.report()` should handle `LocalizedError`, `DecodingError`, and `EncodingError` identically to other errors

MahdiBM:
`LocalizedError.localizedDesceiption` is useless. We shouldn't use it.
The special handling of `DecodingError` and `EncodingError` also doesn't do much.

See https://github.com/apple/swift/issues/68102 and a bunch more issues about `LocalizedError`.


Sources/Vapor/Logging/Logger+Report.swift
```diff
@@ -37,7 +37,7 @@ extension Logger {
             source = nil
             level = .warning
         case let localized as LocalizedError:
-            reason = localized.localizedDescription
+            reason = localized.errorDescription ?? "\(localized)"
```
gwynne, line 40: I would prefer that we just remove the `as LocalizedError` case altogether and let it fall through to the default. Arguably we should do the same for `EncodingError` and `DecodingError` as well, since we don't do anything special for those errors (not anymore, anyway; we did a long time ago, before they had built-in descriptions of their own). This also leaves us no longer adding "special" handling for any error types except our own custom ones, which is in keeping with (if not strictly a case of obedience to) the general rule to only extend other frameworks' types with one's own protocols).

Honestly, the entire thing is a bit ridiculous in modern Swift (though it made sense in the 4.2 days). No one, in my experience, actually uses `DebuggableError` (certainly Vapor itself doesn't), nor `AbortError` for that matter (aside from the obvious sole purview of `Abort` itself).

gwynne (changes requested):

Sources/Vapor/Logging/Logger+Report.swift
MahdiBM: I agree honestly, that was one other way I could have gone with the PR.

MahdiBM: The `String(reflecting:)` is still useful though, specially with PostgresNIO and the upcoming other SQL driver changes.

Sources/Vapor/Logging/Logger+Report.swift
```diff
@@ -28,24 +28,12 @@ extension Logger {
...
-            reason = localized.localizedDescription
-            source = nil
-            level = .warning
         default:
             reason = String(reflecting: error)
             source = nil
             level = .warning
         }
-        
+
```
gwynne, line 36: Believe it or not, I actually approve of this whitespace change on cleanliness grounds 😆

Sources/Vapor/Logging/Logger+Report.swift
gwynne: Precisely so!

Sources/Vapor/Logging/Logger+Report.swift
MahdiBM: That's just my workspace settings doing that 😅
Xcode does have a settings for enabling that although this was vscode doing this.

MahdiBM:
Updated description of the PR since Penny uses that in the Release.

gwynne (approved):

Gerzer:
This change is a regression that should be reverted. See [my comment](https://github.com/apple/swift/issues/68102#issuecomment-1728355333) on the `apple/swift` issue.

PR #3069, opened 2023-09-20
Revert "`Logger.report()` should handle `LocalizedError`, `DecodingError`, and `EncodingError` identically to other errors"

Gerzer:
Reverts vapor/vapor#3068

The best practice is to use `localizedDescription` for reporting errors in UI and logs. The reason why you might see a default string is that the author of whichever error type is being reported didn’t both make the type conform to `LocalizedError` and implement `errorDescription` as an instance property of type `String?`. Whatever string is returned in an `errorDescription` implementation in a `LocalizedError`-conforming type is automatically used as the value of `localizedDescription`, even in contexts in which the type is statically bound to implementations on `any Error`.

There’s no need to handle `LocalizedError` specially to take advantage of this behavior. I would suggest that a separate pull request remove the undesirable special handling without the collateral damage of ignoring `localizedDescription` entirely.

See [my comment](https://github.com/apple/swift/issues/68102#issuecomment-1728355333) on the relevant issue in `apple/swift`.

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -11,4 +11,4 @@ jobs:
      with:
        package_name: vapor
        modules: Vapor,XCTVapor
-       pathsToInvalidate: /vapor/* /xctvapor/*
+       pathsToInvalidate: /vapor /xctvapor
--- a/.github/workflows/projectboard.yml
+++ b/.github/workflows/projectboard.yml
@@ -1,0 +1,11 @@
+name: issue-to-project-board-workflow
+on:
+  # Trigger when an issue gets labeled or deleted
+  issues:
+    types: [reopened, closed, labeled, unlabeled, assigned, unassigned]
+
+jobs:
+  update_project_boards:
+    name: Update project boards
+    uses: vapor/ci/.github/workflows/update-project-boards-for-issue.yml@reusable-workflows
+    secrets: inherit
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,14 +23,14 @@ jobs:
           - provider: vapor/apns
             ref: 3.0.0
     runs-on: ubuntu-latest
-    container: swift:5.9-jammy
+    container: swift:5.8-jammy
     steps: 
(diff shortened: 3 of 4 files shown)


vzsg:
> The **best practice** is to use localizedDescription for reporting errors in UI and logs.

Where does that come from? Especially in the context of logs, when even most of Apple's own Error types don't have this conformance, and printing localizedDescriptions would definitely lead to nothing else but non-actionable noise.

MahdiBM:
(For reference, answered your comment in the Swift repo)

gwynne:
Agreed. The behavior of the entire `NSError` subsystem (of which `LocalizedError` is a part) on Linux is inconsistent, misleading, and frankly borderline malicious. Trivial behaviors in `swift-corelibs-foundation` malfunction. For example:

- Take the underscored `_ErrorCodeProtocol` whose `~=` operator is intended to allow using "nice" error codes (such as `CocoaError.fileReadNoSuchFile`) in `case` and `catch` patterns. This works beautifully on macOS - but on Linux, even though the operator itself (and even it's implementation) is clearly visible in the open-source repo, it yields syntax errors in practice (at least in Swift 5.8).
- Or take `LocalizedError` itself, whose conformance to `CustomStringConvertible` looks in the error's `userInfo` for `NSLocalizedDescriptionKey` on macOS, but on Linux uses the value of `NSLocalizedFailureReasonErrorKey`.
- Then there's the `CustomNSError` protocol, which is public but to all appearances is patently useless on Linux, where bridging between `Swift.Error` and `NSError` _doesn't work_ (you can forcibly cast with `as!`, but the toll-free `as` cast that works on macOS won't even compile elsewhere).
- And certainly you can forget about the fancy magic that's expected to make an `NSError` appear as different Swift types depending on its domain.

Like the rest of the story when it comes to both the Swift and Foundation visions of error handling, the ideal has not panned out in practical use, and pretending that leveraging `LocalizedError`'s API yields anything other than confusing and useless obscurity flies directly in the face of the usability that's supposed to be one of Vapor's goals.

Issue #2508, opened 2020-10-14
Use logger request-id from Heroku's X-Request-ID header


grennis:
### Steps to reproduce

Vapor is using a generated UUID for each request-id for the logger (see `Request` initializer). It would be nice for Vapor and Heroku to use the same ID, for tracing requests. Heroku passes the request ID in the X-Request-ID header. What is the best way for Vapor to use this request ID instead?

I tried to create middleware to do this, but the Vapor default Responder logs the request before any middleware is executed. Therefore, I had to write my own responder which just set the logger request ID from the header, and then forwards to the default responder. Ideally this wouldn't be necessary.

### Expected behavior

A way to get the request ID from a header, or a way to customize the logger request ID via callback.

### Actual behavior

Creating a forwarding responder seems the only way to do this currently.

### Environment

* Vapor Framework version: 4.31.0

gwynne:
Implemented in #3072

PR #3072, opened 2023-09-21
Pass X-Request-Id if present

paunik:
Pass the value from the X-Request-Id to the logger context

- pass the X-Request-Id header value if present, pass generated UUID if the header not present

Mentioned:
- https://github.com/vapor/vapor/issues/2508

Why:
 - We needed a way to trace router -> web -> worker on the logger on the heroku and we found this https://devcenter.heroku.com/articles/http-request-id

Inspiration:
- https://api.rubyonrails.org/classes/ActionDispatch/RequestId.html
- https://http.dev/x-request-id

### DIFF
--- a/Sources/Vapor/Request/Request.swift
+++ b/Sources/Vapor/Request/Request.swift
@@ -250,7 +250,7 @@ public final class Request: CustomStringConvertible {
         self.storage = .init()
         self.isKeepAlive = true
         self.logger = logger
-        if let requestId = self.headers["X_REQUEST_ID"].first {
+        if let requestId = self.headers[.xRequestId].first {
             self.logger[metadataKey: "request-id"] = .string(requestId)
         } else {
             self.logger[metadataKey: "request-id"] = .string(UUID().uuidString)
--- a/Sources/Vapor/Request/Request.swift
+++ b/Sources/Vapor/Request/Request.swift
@@ -250,16 +250,11 @@ public final class Request: CustomStringConvertible {
         self.storage = .init()
         self.isKeepAlive = true
         self.logger = logger
-<<<<<<< Updated upstream
-        self.logger[metadataKey: "request-id"] = .string(id)
-=======
         if let requestId = self.headers["X_REQUEST_ID"].first {
             self.logger[metadataKey: "request-id"] = .string(requestId)
         } else {
             self.logger[metadataKey: "request-id"] = .string(UUID().uuidString)
         }
-
->>>>>>> Stashed changes
         self.byteBufferAllocator = byteBufferAllocator
     }
 }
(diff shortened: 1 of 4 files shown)


Sources/Vapor/Request/Request.swift
```diff
@@ -250,7 +250,16 @@ public final class Request: CustomStringConvertible {
...
         self.logger[metadataKey: "request-id"] = .string(id)
+=======
+        if let requestId = self.headers["X_REQUEST_ID"].first {
+            self.logger[metadataKey: "request-id"] = .string(requestId)
+        } else {
+            self.logger[metadataKey: "request-id"] = .string(UUID().uuidString)
+        }
+
+>>>>>>> Stashed changes
         self.byteBufferAllocator = byteBufferAllocator
```
gwynne, line 263: Please fixup this leftover merge conflict.

gwynne (changes requested):
This looks fine, aside from the conflict markers left unresolved.

Sources/Vapor/Request/Request.swift
paunik: on it

paunik: Should be done, tnx for the quick check, I should have seen this myself.

Tests/VaporTests/HTTPHeaderTests.swift
```diff
@@ -201,6 +201,13 @@ final class HTTPHeaderTests: XCTestCase {
         )
     }
 
+    func testXRequestId() throws {
+        var headers = HTTPHeaders()
+        let xRequestId = UUID().uuidString
+        headers.replaceOrAdd(name: .xRequestId, value: xRequestId)
+        XCTAssertEqual(headers.first(name: "X-Request-Id"), xRequestId)
+    }
+
```
gwynne, line 210: Can you also add an end-to-end test which checks that the header is used when present in a request?

gwynne (approved):

dannflor (approved):
Beautamous

MahdiBM (commented):
Honestly considering this header doesn't make an appearance in MDN, I wouldn't hard-code it like this 🤔

I would instead add a configuration option so other users can configure it "based on headers", and there they could specify the "X-Request-ID" header.

I know this complicates the solution, but we have "tracing" and "task-locals" for logging now which enable you to this yourself to some extent, and caring about this particular header seems out of scope for Vapor to me.

gwynne:
@MahdiBM I disagree - `X-Request-ID` is not an official standard, true, but it is a de facto one. (Enough so, even for people to have gone to the trouble of noting that [RFC 6648](https://www.rfc-editor.org/rfc/rfc6648) deprecated the `X-` prefix on headers and remarked on the need to find a new name.) Unless you can cite links to any alternate names commonly used for the same purpose, configurability would just bloat the API surface and force users to do additional work without adding any real value.

PR #3074, opened 2023-09-27
Add numerous missing `@preconcurrency` attributes

gwynne:
It has become standard practice to add `@Sendable` to `@escaping` closures passed as method parameters to improve Concurrency correctness. However, when this is done for pre-existing `public` methods that are _not_ `async`, the result is source incompatibility for some users, as mutable values captured by such closures will cause unexpected build errors. The correct way to suppress this behavior is to mark such methods with the `@preconcurrency` attribute, signaling to the compiler that users may not yet expect the additional restrictions of `@Sendable` to apply without sacrificing correctness for Concurrency-ready code. Unfortunately, Vapor recently added `@Sendable` annotations to many of its APIs without also adding the `@preconcurrency` annotation; this update addresses that oversight.

### DIFF
--- a/Sources/Development/routes.swift
+++ b/Sources/Development/routes.swift
@@ -1,5 +1,5 @@
 import class Foundation.Bundle
-import Vapor
+@preconcurrency import Vapor
 import NIOCore
 import NIOHTTP1
 import NIOConcurrencyHelpers
--- a/Sources/Development/routes.swift
+++ b/Sources/Development/routes.swift
@@ -1,5 +1,5 @@
 import class Foundation.Bundle
-@preconcurrency import Vapor
+import Vapor
 import NIOCore
 import NIOHTTP1
 import NIOConcurrencyHelpers
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
         <img src="https://github.com/vapor/vapor/actions/workflows/test.yml/badge.svg?branch=main" alt="Continuous Integration">
     </a>
     <a href="https://swift.org">
-        <img src="https://img.shields.io/badge/swift-5.6-brightgreen.svg" alt="Swift 5.6">
+        <img src="https://img.shields.io/badge/swift-5.7-brightgreen.svg" alt="Swift 5.7">
     </a>
     <a href="https://twitter.com/codevapor">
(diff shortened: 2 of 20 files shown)


ptoffy (approved):

MahdiBM (approved):
These are all old functions which slept through a sendable PR without the @preconcurrency from what I can tell.

~~What I want to say is that even they were new function I would have still liked them to have that @preconcurrency considering a lot of Vapor stuff already have that and the inconsistency is kind of unnecessary to not let those who are not caring about sendable warnings just continue to not care for now.~~

Sources/Development/routes.swift
```diff
@@ -1,5 +1,5 @@
 import class Foundation.Bundle
-import Vapor
+@preconcurrency import Vapor
```
0xTim, line 2: I think we should keep this so we know when the Sendable warning is gone. This will also mask anything we've missed in the future etc

0xTim (approved):
One comment on the development target but LGTM

0xTim:
> What I want to say is that even they were new function I would have still liked them to have that @preconcurrency considering a lot of Vapor stuff already have that and the inconsistency is kind of unnecessary to not let those who are not caring about sendable warnings just continue to not care for now.

We want people to care though because so they migrate incrementally as more and more checking gets turned on rather than trying to build with Swift 6 and getting a ton of errors. New functions should definitely be Sendable by default and we shouldn't be giving people a get out clause if they want to use them

MahdiBM:
@0xTim you're right. I take what I said back 😅

I was thinking about massive code bases which don't comply with Senabilities and don't want to bother yet. Though thinking more, it's probably not that hard for them to make something Sendable. If they really can't make it Sendable, then they should just mark it as @unchecked, not Vapor adding @preconcurrency to every thing just in that case.

0xTim:
> it's probably not that hard for them to make something Sendable

This I may take some exception to 😆

MahdiBM:
I meant with some dirty use of @unchecked assuming they really don't want to care about Sendable 😅 otherwise it could be a pretty hard task

gwynne:
> I meant with some dirty use of @unchecked assuming they really don't want to care about Sendable 😅 otherwise it could be a pretty hard task

Or, yanno, just `.enableExperimentalFeature("StrictConcurrency=minimal")`... 🙃

Issue #3076, opened 2023-10-01
File Middleware strips query parameters with redirect


Captain-Kirkie:
### Describe the bug
When using the the DirectoryAction .redirect on FileMiddleWare we lose any query parameters sent with the request on redirect. 

Example:
GET /Utilities?test=test -> 301 Moved Permanently /Utilities/ (Query parameters are stripped.)

### To Reproduce

To reproduce:
1) Use Vapor FileMiddleware with directory action .redirect. 
2) Serve files through file middleware. 
3) Make getRequest without trailing "/" and add query parameters GET /Utilities?test=test.

1. Add package with configuration '...'
2. Send request with options '...'
3. See error

### Expected behavior
GET /Utilities?test=test -> 301 Moved Permanently /Utilities/?test=test (Query parameters persist)

### Environment

* Vapor Framework version: framework: 4.78.2
* Vapor Toolbox version: toolbox: 18.7.1
* OS version: MacOS Ventura Version 13.6

Captain-Kirkie:
Pull request [Here](https://github.com/vapor/vapor/pull/3077)

0xTim:
Fixed in #3077

PR #3077, opened 2023-10-01
Handle query parameters in FileMiddleware redirects

Captain-Kirkie:
Correctly handle query parameters when using the redirect functionality in `FileMiddleware`.

### DIFF
--- a/Sources/Vapor/Middleware/FileMiddleware.swift
+++ b/Sources/Vapor/Middleware/FileMiddleware.swift
@@ -61,9 +61,11 @@ public final class FileMiddleware: Middleware {
         if isDir.boolValue {
             guard absPath.hasSuffix("/") else {
                 switch directoryAction.kind {
-                case .redirect:
+                case .redirect:                    
+                    var redirectUrl = request.url
+                    redirectUrl.path += "/"
                     return request.eventLoop.future(
-                        request.redirect(to: request.url.path + "/", redirectType: .permanent)
+                        request.redirect(to: redirectUrl.string, redirectType: .permanent)
                     )
                 case .none:
                     return next.respond(to: request)
--- a/Tests/VaporTests/FileTests.swift
+++ b/Tests/VaporTests/FileTests.swift
@@ -373,6 +373,31 @@ final class FileTests: XCTestCase {
         }
     }
     
+    func testRedirectWithQueryParams() throws {
+        let app = Application(.testing)
+        defer { app.shutdown() }
+
+        let path = #file.split(separator: "/").dropLast().joined(separator: "/")
+        app.middleware.use(
+            FileMiddleware(
+                publicDirectory: "/" + path,
+                defaultFile: "index.html",
+                directoryAction: .redirect
+            )
+        )
+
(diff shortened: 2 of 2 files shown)


Captain-Kirkie:
Issue [Here](https://github.com/vapor/vapor/issues/3076)

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -164,3 +165,16 @@ fileprivate extension String {
         return newPath
     }
 }
+
+
+extension Request {
+    public func getQueryString() -> String {
```
0xTim, line 171: You can just use `req.url.query` instead

0xTim (changes requested):
Thanks for the PR! One change then we can merge

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -62,8 +62,9 @@ public final class FileMiddleware: Middleware {
             guard absPath.hasSuffix("/") else {
                 switch directoryAction.kind {
                 case .redirect:
+                    let queryStr = request.getQueryString()
                     return request.eventLoop.future(
-                        request.redirect(to: request.url.path + "/", redirectType: .permanent)
+                        request.redirect(to: request.url.path + "/" + queryStr, redirectType: .permanent)
```
gwynne, line 67: Why not redirect the entire URL (with the modified path) so as to also preserve authority parameters (username, password) and other components such as the fragment (text following a `#`)?

0xTim: That's a better solution

Not that fragment would be passed back to the server but the auth stuff would be

Captain-Kirkie: That is more elegant. I have updated the PR to use the entire URL.

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -62,8 +62,9 @@ public final class FileMiddleware: Middleware {
             guard absPath.hasSuffix("/") else {
                 switch directoryAction.kind {
                 case .redirect:
+                    let redirectLocation = request.url.string.replacingOccurrences(of: request.url.path, with: "\(request.url.path)/")
                     return request.eventLoop.future(
-                        request.redirect(to: request.url.path + "/", redirectType: .permanent)
+                        request.redirect(to: redirectLocation, redirectType: .permanent)
```
gwynne, line 67: This is unsafe if the path occurs multiple times in the URL (which is perfectly possible). Try something like this, maybe:
```suggestion
                    var redirectUrl = request.url
                    redirectUrl.path += "/"
                    return request.eventLoop.future(request.redirect(to: redirectUrl.string, redirectType: .permanent))
```

gwynne (changes requested):

Sources/Vapor/Middleware/FileMiddleware.swift
Captain-Kirkie: whoops, good catch. Updated.

0xTim (approved):
Thanks!

PR #3081, opened 2023-10-05
Fix NIOLoopBound issues

0xTim:
Fixes a number of issues where `NIOLoopBound` and `NIOLoopBoundBox` were used without ensuring we were on the correct event loop before accessing them. This could lead to precondition crashes

### DIFF
--- a/Tests/VaporTests/PipelineTests.swift
+++ b/Tests/VaporTests/PipelineTests.swift
@@ -5,10 +5,21 @@ import NIOEmbedded
 import NIOCore
 
 final class PipelineTests: XCTestCase {
+    var app: Application!
+    var eventLoopGroup: EventLoopGroup!
+    
+    override func setUp() async throws {
+        eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 8)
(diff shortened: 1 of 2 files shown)


Tests/VaporTests/PipelineTests.swift
```diff
@@ -5,10 +5,21 @@ import NIOEmbedded
 import NIOCore
 
 final class PipelineTests: XCTestCase {
+    var app: Application!
+    var eventLoopGroup: EventLoopGroup!
+    
+    override func setUp() async throws {
+        eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 8)
```
gwynne, line 12: Why not just use the singleton?

0xTim: I'll switch the this, this was my attempts to make it crash and ensure we definitely had more than one event loop

0xTim:
I've audited the other usages of `NIOLoopBound` and `NIOLoopBoundBox` and I can't see a way for the others to get off the event loop - there's always a check, or submit so I think we're good

gwynne (approved):

PR #3082, opened 2023-10-06
Sendable Response

0xTim:
Make `Response` `Sendable`

### DIFF
--- a/Sources/Vapor/HTTP/Server/HTTPServerUpgradeHandler.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServerUpgradeHandler.swift
@@ -128,7 +128,7 @@ private final class UpgradeBufferHandler: ChannelInboundHandler, RemovableChanne
 }
 
 /// Conformance for any struct that performs an HTTP Upgrade
-public protocol Upgrader {
+public protocol Upgrader: Sendable {
     func applyUpgrade(req: Request, res: Response) -> HTTPServerProtocolUpgrader
 }
 
--- a/Sources/Vapor/Response/Response.swift
+++ b/Sources/Vapor/Response/Response.swift
@@ -1,26 +1,48 @@
 import NIOCore
 import NIOHTTP1
 import NIOFoundationCompat
+import NIOConcurrencyHelpers
 
 /// An HTTP response from a server back to the client.
 ///
 ///     let res = Response(status: .ok)
 ///
 /// See `HTTPClient` and `HTTPServer`.
-public final class Response: CustomStringConvertible {
+public final class Response: CustomStringConvertible, Sendable {
     /// Maximum streaming body size to use for `debugPrint(_:)`.
     private let maxDebugStreamingBodySize: Int = 1_000_000
 
     /// The HTTP version that corresponds to this response.
-    public var version: HTTPVersion
+    public var version: HTTPVersion {
+        get {
+            self._version.withLockedValue { $0 }
+        } 
+        set {
+            self._version.withLockedValue { $0 = newValue }
+        }
+    }
     
     /// The HTTP response status.
-    public var status: HTTPResponseStatus
+    public var status: HTTPResponseStatus {
+        get {
+            self._status.withLockedValue { $0 }
+        }
+        set {
+            self._status.withLockedValue { $0 = newValue }
+        }
+    }
     
     /// The header fields for this HTTP response.
     /// The `"Content-Length"` and `"Transfer-Encoding"` headers will be set automatically
     /// when the `body` property is mutated.
-    public var headers: HTTPHeaders
+    public var headers: HTTPHeaders {
+        get {
+            self._headers.withLockedValue { $0 }
+        }
+        set {
+            self._headers.withLockedValue { $0 = newValue }
+        }
+    }
     
     /// The `Body`. Updating this property will also update the associated transport headers.
     ///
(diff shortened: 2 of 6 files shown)


Sources/Vapor/Response/Response.swift
```diff
@@ -29,17 +51,37 @@ public final class Response: CustomStringConvertible {
     /// Also be sure to set this message's `contentType` property to a `MediaType` that correctly
     /// represents the `Body`.
     public var body: Body {
-        didSet { self.headers.updateContentLength(self.body.count) }
+        get {
+            _body.withLockedValue { $0 }
+        }
+        set {
+            _body.withLockedValue { $0 = newValue }
+            self.headers.updateContentLength(self.body.count)
```
MahdiBM, line 59: perhaps we could just do `self.headers.updateContentLength(newValue)` and skip a `_body.withLockedValue { $0 }`?

MahdiBM (approved):
one comment otherwise looks good

MahdiBM:
There are some warnings in the CI which don't seem blocking for this PR:
* Some `Request` warnings, I think you're planning for adding `Sendable` conformance to that in the future.
* Some `Commands` warnings which there is already a PR up for.

API breakage seems to be inaccurate per usual as well.

Sources/Vapor/Response/Response.swift
0xTim: Good shout

Sources/Vapor/Response/Response.swift
```diff
@@ -110,6 +152,13 @@ public final class Response: CustomStringConvertible {
         }
     }
     
+    private let _version: NIOLockedValueBox<HTTPVersion>
+    private let _status: NIOLockedValueBox<HTTPResponseStatus>
+    private let _headers: NIOLockedValueBox<HTTPHeaders>
+    private let _body: NIOLockedValueBox<Body>
+    private let _upgrader: NIOLockedValueBox<Upgrader?>
+    private let _storage: NIOLockedValueBox<Storage>
```
gwynne, line 160: Fine-grained locking is all well and good for efficiency, but this is an awful lot of additional allocations on a per-response basis. You can easily combine, at the very least, `version`, `status`, and `headers` under a single lock - it's exceedingly rare (in fact, it might never happen at all) to need to access any of these values simultaneously with one or both of the others from another thread, so contention ought not to be an issue. It might even be reasonable to include `body` in the same lock as well, though I'm less certain on that one.

(Truthfully, I might wonder if it didn't make more sense to put _everything_ under a single lock, maybe with the exception of `storage`.)

gwynne: I mean, for example, look at the implementation of `description` and think for a second about just how painful that's going to actually be at runtime...

MahdiBM: @0xTim I can't see a new commit on this branch 🤔

0xTim: I mean it would really help if I actually pushed my changes

Sources/Vapor/HTTP/Server/HTTPServerUpgradeHandler.swift
```diff
@@ -128,7 +131,8 @@ private final class UpgradeBufferHandler: ChannelInboundHandler, RemovableChanne
 }
 
 /// Conformance for any struct that performs an HTTP Upgrade
-public protocol Upgrader {
+@preconcurrency
+public protocol Upgrader: Sendable {
     func applyUpgrade(req: Request, res: Response) -> HTTPServerProtocolUpgrader
```
MahdiBM, line 136: why `@preconcurrency`?

Sources/Vapor/Response/Response.swift
```diff
@@ -71,33 +117,41 @@ public final class Response: CustomStringConvertible {
...
-            guard let body = self.response.body.buffer else {
-                throw Abort(.unprocessableEntity)
+            try self.response.responseBox.withLockedValue { box in
+                guard let body = box.body.buffer else {
+                    throw Abort(.unprocessableEntity)
+                }
+                var decoded = try decoder.decode(C.self, from: body, headers: box.headers)
+                try decoded.afterDecode()
+                return decoded
             }
```
MahdiBM, line 154: The `try decoded.afterDecode()` should be outside?
```suggestion
            var decoded = try self.response.responseBox.withLockedValue { box in
                guard let body = box.body.buffer else {
                    throw Abort(.unprocessableEntity)
                }
                return try decoder.decode(C.self, from: body, headers: box.headers)
            }
            try decoded.afterDecode()
            return decoded
```

Sources/Vapor/Response/Response.swift
```diff
@@ -110,6 +164,24 @@ public final class Response: CustomStringConvertible {
...
+                self.headers.updateContentLength(body.count)
+            }
+        }
+        var upgrader: Upgrader?
+        // If `true`, don't serialize the body.
+        var forHeadRequest: Bool
+
+    }
+    
+    let responseBox: NIOLockedValueBox<ResponseBox>
```
MahdiBM, line 182: I _feel_ like this should be private. No logical explanations though.

MahdiBM (changes requested):
Looks good. 1 issue, 1 suggestion, 1 question 🙂.

Sources/Vapor/HTTP/Server/HTTPServerUpgradeHandler.swift
MahdiBM: Nevermind i see it now. The protocol is conforms to `Sendable` now, so it needs the `@preconcurrency`.

Sources/Vapor/Response/Response.swift
0xTim: We use it internally elsewhere to bypass multiple lock accesses

Sources/Vapor/Response/Response.swift
0xTim: Functionally it doesn't make a difference, the code has the same result, it just releases the lock at a different time. For a sequential block like this it shouldn't matter

MahdiBM: Right the lock release was what I was slightly worrying about. It won't matter much but I think it's still better to release it when you can.

MahdiBM (approved):

gwynne (approved):

PR #3091, opened 2023-10-24
Lazily start thread pool on access

dfed:
Addresses #3003 by lazily starting the thread pool on first access, and deferring first access to the thread pool outside of `Application`'s `init`.

The only downside of this approach is that we're now deferring loading the environment until we call `Application`'s `boot()` method. I'm hoping this is an acceptable compromise in return for allowing consumers to reduce the number of threads consumed by Vapor.

This approach is similar to #3008 but attempts to increase the likelihood of acceptance by reducing the potential side-effects of the change.

### DIFF
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -147,7 +147,6 @@ public final class Application: Sendable {
         self.clients.use(.http)
         self.commands.use(self.servers.command, as: "serve", isDefault: true)
         self.commands.use(RoutesCommand(), as: "routes")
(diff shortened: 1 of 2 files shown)


gwynne:
No; I don't think this is viable. What would be preferable would be to reduce (_significantly_) the number of threads Vapor spins up, the pool is currently configured with an absurdly high number (64) by default. I've been meaning to reduce it (say, to something more like `System.numberOfCores`, and even only that much as a concession to the needs of SQLiteNIO).

dfed:
Got it. Reducing the number of threads spun up from 64 to the current number of cores would be a huge improvement for us. I'll put up a PR for that approach, even though that approach doesn't directly solve #3003.

PR #3092, opened 2023-10-24
Create a thread pool of System.coreCount rather than 64 when initializing an Application

dfed:
Following [this recommendation](https://github.com/vapor/vapor/pull/3091#issuecomment-1778078831), this PR simply changes the default number of cores when initializing an `Application` from 64 to the current number of cores.

This means that initializing an `Application` with default arguments now spins up 2x the number of threads as cores on device. We'll create `System.coreCount` threads when instantiating the `threadPool`, and another `System.coreCount` threads when we instantiate the `eventLoopGroup`.

2x `System.coreCount` is still _way_ better than `System.coreCount + 64` on older devices, so while this PR does not directly address #3003, it does reduce total thread usage on `init`.


gwynne:
Arguably, on iOS users ought to be spinning up an NIOTSEventLoopGroup instead (since that can be configured, by replacing the ELG early in the lifecycle). But that's another issue 😅

dfed:
> Arguably, on iOS users ought to be spinning up an NIOTSEventLoopGroup instead (since that can be configured, by replacing the ELG early in the lifecycle). But that's another issue 😅

Absolutely! We'll still have the threads from this thread pool that we can't limit, but between this change and utilizing a NIOTSEventLoopGroup we can dramatically lower our overhead 🙂

gwynne:
Can you elaborate on "isn't a viable loop"?

Also, I should mention that the thread pool absolutely is configurable, as long as it happens before `Application.boot()`:

```swift
func configure(_ app: Application) async throws {
    app.threadPool = NIOThreadPool(numberOfThreads: /* any value you like here */)
    // do _NOT_ call `start()` on the new thread pool (or `shutdown()` on the old one); Application does it for you
}
```

dfed:
> Can you elaborate on "isn't a viable loop"?

Happily! When I used a `NIOTSEventLoopGroup` I hit [this assert](https://github.com/apple/swift-nio/blob/935dbdf114e8008ae190af78442db50a7c4eaec9/Sources/NIOPosix/Bootstrap.swift#L101-L102) when starting a `HTTPServer`:

image

> Also, I should mention that the thread pool absolutely is configurable, as long as it happens before Application.boot()

True! While the larger thread pool will have already started within `init`, tearing it down and providing a smaller one is certainly helpful.

gwynne:
Whoops... guess we never switched to a universal bootstrap 🤦‍♀️ cc @0xTim

gwynne (approved):

0xTim:
> Whoops... guess we never switched to a universal bootstrap 🤦‍♀️ cc @0xTim

@gwynne yeah it's never come up as a major priority, but should be done at some point

John-Connolly:
@dfed Thanks for pushing this forward!

PR #3093, opened 2023-11-01
Make Request Sendable

0xTim:
Final stage of Vapor's `Sendable` journey as `Request` is now `Sendable`. 

There should be no more `Sendable` warnings in Vapor, even with complete concurrency checking turned on.


0xTim:
This was surprisingly easy in the end. Tried to reduce the locks where possible, I've added individual locks to where it's likely to be accessed concurrently

Sources/Vapor/Routing/Route.swift
```diff
@@ -1,33 +1,94 @@
...
+    }
+    
+    public var userInfo: [AnySendableHashable: Sendable] {
+        get {
+            self.sendableBox.withLockedValue { $0.userInfo }
+        }
+        set {
+            self.sendableBox.withLockedValue { $0.userInfo = newValue }
+        }
+    }
```
MahdiBM, line 67: This breaks API?
`[AnyHashable: Any]` -> `[AnySendableHashable: Sendable]`

MahdiBM (changes requested):
Looks good but some requested changes.

I also see one probably-resolvable warning in "Development" on 5.9:

> /__w/vapor/vapor/Sources/Development/routes.swift:274:36: warning: converting non-sendable function value to '@Sendable (Request) async throws -> some AsyncResponseEncodable' may introduce data races
    asyncRoutes.get("opaque", use: opaqueRouteTester)

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,44 @@
...
+        // Configure properties specific to VS Code.
+        "vscode": {
+            // Set *default* container specific settings.json values on container create.
+            "settings": {
+                "lldb.library": "/usr/lib/liblldb.so"
+            },
+            // Add the IDs of extensions you want installed when the container is created.
+            "extensions": [
+                "sswg.swift-lang"
+            ]
```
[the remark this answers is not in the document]
gwynne: I tend to agree; this is an excessive addition unless we want to commit to keeping it maintained at least as much so as the rest of the support files we provide.

.gitignore
```diff
@@ -7,6 +7,8 @@ Package.resolved
 .swiftpm
 Tests/LinuxMain.swift
 .vscode
+.bash_history
+.cache/
```
gwynne, line 11: Why would these be in the Vapor repo? Does VSCode end up using the package root as a user's homedir in dev containers? 🤨

Sources/Vapor/Commands/ServeCommand.swift
```diff
@@ -1,5 +1,4 @@
-import Foundation
-@preconcurrency import Dispatch
+@preconcurrency import Foundation
```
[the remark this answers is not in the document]
gwynne: Good catch, Mahdi 👍

Sources/Vapor/Commands/ServeCommand.swift
```diff
@@ -39,15 +38,17 @@ public final class ServeCommand: Command, Sendable {
...
     }
 
-    private var signalSources: [DispatchSourceSignal]
-    private var didShutdown: Bool
-    private var server: Server?
-    private var running: Application.Running?
+    private let didShutdown: NIOLockedValueBox<Bool>
+    private let running: NIOLockedValueBox<Application.Running?>
+    private let signalSources: NIOLockedValueBox<[DispatchSourceSignal]>
+    private let server: NIOLockedValueBox<Server?>
```
gwynne, line 44: Too-granular locking again, Tim.

Sources/Vapor/HTTP/Headers/HTTPCookies.swift
```diff
@@ -266,13 +266,13 @@ public struct HTTPCookies: ExpressibleByDictionaryLiteral {
...
             return "\(name)=\(value.string)"
         }.joined(separator: "; ")
-
+        
         return cookie
     }
-
+    
     var setCookieHeaders: [String] {
         return self.cookies.map { $0.value.serialize(name: $0.key) }
```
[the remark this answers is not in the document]
gwynne: I would much rather add it where it's missing than remove it where it exists. Lack of that setting was the very first thing I noticed as a problem with `swift-format`, even before its newlines problem

Sources/Vapor/HTTP/Headers/HTTPCookies.swift
```diff
@@ -287,7 +287,7 @@ public struct HTTPCookies: ExpressibleByDictionaryLiteral {
     
     /// Access `HTTPCookies` by name
     public subscript(name: String) -> Value? {
-        get { return cookies[name] }
+        get { return self.cookies[name] }
```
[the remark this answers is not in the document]
gwynne: Fix the statements around it.

Sources/Vapor/Routing/Route.swift
gwynne: Correct, it does.

Sources/Vapor/Routing/Route.swift
```diff
@@ -1,33 +1,94 @@
 import NIOHTTP1
-import RoutingKit
+@preconcurrency import RoutingKit
+import NIOConcurrencyHelpers
```
gwynne, line 4: I would hasten to point out that there's nothing held by `Route` that inherently makes it non-`Sendable` in theory- or for that matter, that justifies it being a reference type rather than a value type in the first place (or for _that_ matter, that justifies it being mutable at _all_!). I know that can't be changed without breaking API, but it does suggest that `@unchecked Sendable` ("I know what I'm doing, punish me if I hork it up") might be a better choice than a bunch of locking no one should be hitting to begin with.

Sources/Vapor/Utilities/AnySendableHashable.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+/// A `Sendable` version of the standard library's `AnyHashable` type.
+public struct AnySendableHashable: @unchecked Sendable, Hashable, ExpressibleByStringLiteral {
+    // Note: @unchecked Sendable since there's no way to express that `wrappedValue` is Sendable, even though we ensure that it is in the init.
+    let wrappedValue: AnyHashable
+    
+    public init(_ wrappedValue: any Hashable & Sendable) {
```
gwynne, line 8:
```suggestion
    public init(_ wrappedValue: some Hashable & Sendable) {
```
This brings it in line with `AnyHashable`'s initializer:
```swift
@frozen public struct AnyHashable {
    public init<H>(_ base: H) where H : Hashable
```
(which is functionally identical to `public init(_ base: some Hashable)`)

Sources/Vapor/Utilities/AnySendableHashable.swift
```diff
@@ -0,0 +1,15 @@
...
+        self.wrappedValue = AnyHashable(wrappedValue)
+    }
+    
+    public init(stringLiteral value: String) {
+        self.init(value)
+    }
+}
```
gwynne, line 15: `AnyHashable` isn't `ExpressibleByStringLiteral`, where did that come from?

Also, to be equivalent to `AnyHashable`, you need to add:
```swift
extension AnySendableHashable: CustomStringConvertible, CustomDebugStringConvertible, CustomReflectable {
    public var description: String { self.wrappedValue.description }
    public var debugDescription: String { self.wrappedValue.debugDescription }
    public var customMirror: Mirror { self.wrappedValue.customMirror }
}

gwynne (changes requested):
Per comments above

Sources/Vapor/Routing/Route.swift
MahdiBM: What do we want to do then? Live with it and just make it as least noticeable as possible?

.gitignore
0xTim: This was from me running a quick interactive Docker container to test on Linux - I didn't think it was an issue to add them

Sources/Vapor/Utilities/AnySendableHashable.swift
0xTim: This was a direct lift from [ConsoleKit](https://github.com/vapor/console-kit/blob/21b72c579a0d04a530e2b2f456c551c5b8940cf8/Sources/ConsoleKit/Terminal/Terminal.swift#L12) - we should probably do the same there

gwynne: Then why in the world are you copying it into Vapor? You've already got ConsoleKit's version available, it's public.

Sources/Vapor/Routing/Route.swift
0xTim: It _should_ only introduce warnings for people (unless they're relying on specific signatures). We did the same in ConsoleKit so the precedent has been set there

Sources/Vapor/Utilities/AnySendableHashable.swift
0xTim: Err that I don't have a good answer for!

Sources/Vapor/Commands/ServeCommand.swift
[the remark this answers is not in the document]
0xTim: `DispatchSourceSignal` isn't `Sendable` on either platform so it's required here for now

Sources/Vapor/Routing/Route.swift
0xTim: The only thing that requires this here is `PathComponent` which we can make Sendable and I'll do in another PR when I tackle RoutingKit. That one I want to be very careful with though for obvious reasons

0xTim: Wrt to the whole thing of `Route` and unchecked `Sendable` I'm planning on doing a better audit of Vapor things, deprecating thing and then introducing a new `AsyncRoute` in it's place

Sources/Vapor/Commands/ServeCommand.swift
0xTim: I did this because it's in the top level command so shouldn't affect things (since it's only used at app start up) but will change

0xTim:
@MahdiBM that's an annoying Swift bug with complete concurrency checking and nothing we can do at the moment until it's fixed in Swift

MahdiBM:
O it's that bug... you can just open up the closure and the bug will go away. Though I'm fine with it either way since it's a bug.

MahdiBM (approved):

gwynne (approved):

PR #3094, opened 2023-11-02
Add snake-key decoding strategy for URLQuery

MahdiBM:
Adds snake-key decoding strategy for URLQuery.
Matches Foundation's JSONDecoder implementation although i found like 3 different key-handling bugs in that implementation while writing this.


MahdiBM:
I thought my misclick is merging the PR to main without a review 😵‍💫

0xTim:
It can't until it's had an approval

MahdiBM:
hmm that didn't turn out like i was thinking it would 🤔 just caused a bunch of whitespace changes for the most part

PR #3095, opened 2023-11-02
Remove Sendable requirements on Authenticatable

0xTim:
Removes the requirement for `Authenticatable` types to be `Sendable` which was causing issues with Fluent models (and any reference types) and wasn't solvable in a non-breaking way.

This uses an unsafe box to wrap the `Authenticatable` types which removes compiler checking on usage of the box but should not be an issue due to the way Vapor's auth is implemented

### DIFF
--- a/Sources/Vapor/Authentication/AuthenticationCache.swift
+++ b/Sources/Vapor/Authentication/AuthenticationCache.swift
@@ -101,7 +101,7 @@ extension Request.Authentication {
 // types (e.g. Fluent 4 models can never be Sendable because they're reference types with mutable values
 // required by protocols and property wrappers). This allows us to store the Authenticatable type in a
 // safe-most-of-the-time way. This does introduce an edge case where type could be stored and mutated in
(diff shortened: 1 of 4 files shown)


Sources/Vapor/Authentication/AuthenticationCache.swift
```diff
@@ -96,3 +96,22 @@ extension Request.Authentication {
         }
     }
 }
+
+// This is to get around the fact that for legacy reasons we can't enforce Sendability on Authenticatable
+// types (e.g. Fluent 4 models can never be Sendable because they're reference types with mutable values
+// required by protocols and property wrappers). This allows us to store the Authenticatable type in a
+// safe-most-of-the-time way. This does introduce an edge case where type could be stored and mutated in
+// multiple places. But given how Vapor and its users use Authentication this should almost always never
```
gwynne, line 104:
```suggestion
// multiple places. But given how Vapor and its users use Authentication this should almost never
```

0xTim:
Of course 😛

PR #3096, opened 2023-11-02
Make Async Request Body actually work

### LINKED ISSUE #3088: Streaming request crash
### Describe the bug

Streaming request body causes a crash.

### To Reproduce

Include the following code in a request handler:

```swift
for try await buffer in req.body {
    print(buffer)
}
````

### Environment

* Vapor Framework version: 4.84.6 (also happens with 4.84.3)
* Swift version: Swift 5.9

### Additional context
Briefly discussed on [Discord](https://discord.com/channels/431917998102675485/472522745067077632/1164305973700218920).

### PR DESCRIPTION
**Pull request by 0xTim**
The existing implementation of adding an `AsyncSequence` to `Request.Body` had two issues:

* it didn't ensure code was being called from the correct event loop which broke `Sendable` guarantees and was unsafe
* it would hit a precondition failure in the implementation if backpressure was triggered because the initial state was not accounted for

This fixes that

Resolves #3088

### CODE REVIEW

**Review by MahdiBM — APPROVED** [review]
Looks as good as it could be to me

File: Tests/AsyncTests/AsyncRequestTests.swift [hunk 1]
```diff
@@ -33,13 +43,80 @@ final class AsyncRequestTests: XCTestCase {
...
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@usableFromInline
+struct AsyncLazySequence<Base: Sequence>: AsyncSequence {
+    @usableFromInline typealias Element = Base.Element
+    @usableFromInline struct AsyncIterator: AsyncIteratorProtocol {
```
MahdiBM, line 89: We could just use `AsyncAlgorithms`'s `var async`, but not sure if it's fine to add a dependency to that.
That also takes Task cancellation into account which we don't want, but this is in tests and that's not going to happen anyway.
https://github.com/apple/swift-async-algorithms/blob/5bbdcc1d37990cca524852c5a3924c02fa2da983/Sources/AsyncAlgorithms/AsyncSyncSequence.swift#L46C5-L46C5

gwynne: Honestly I think adding an entire "this uses code from ..." mess etc. etc. and all this format boilerplate is a hell of a lot of overkill for a test to begin with. The logic here is just not involved enough to merit the effort. I mean for pity's sake, look at what happens when you yank all the noise, it's literally nothing but another layer of boilerplate:
```swift
struct AsyncLazySequence<Base: Sequence & Sendable>: AsyncSequence, Sendable {
    typealias Element = Base.Element
    struct AsyncIterator: AsyncIteratorProtocol, Sendable {
        var iter: Base.Iterator
        mutating func next() async throws -> Base.Element? { self.iter.next() }
    }
    var base: Base
    func makeAsyncIterator() -> AsyncIterator { .init(iter: self.base.makeIterator()) }
}
extension Sequence where Self: Sendable { var async: AsyncLazySequence<Self> { .init(base: self) } }
```
(Also can I just point out that the `@available` annotations are completely meaningless in this context? Like, triply so. And the `@inlinable` annotations are equally worthless in a test, since it's not optimizing compilation.)

Tests/AsyncTests/AsyncRequestTests.swift
```diff
@@ -33,13 +43,80 @@ final class AsyncRequestTests: XCTestCase {
...
-            XCTAssertEqual(testValue, returnedString)
+        app.environment.arguments = ["serve"]
+        XCTAssertNoThrow(try app.start())
+        
+        XCTAssertNotNil(app.http.server.shared.localAddress)
+        guard let localAddress = app.http.server.shared.localAddress,
+              let ip = localAddress.ipAddress,
+              let port = localAddress.port else {
+            XCTFail("couldn't get ip/port from \(app.http.server.shared.localAddress.debugDescription)")
+            return
```
gwynne, line 54:
```suggestion
            return XCTFail("couldn't get ip/port from \(app.http.server.shared.localAddress.debugDescription)")
```

gwynne (approved):

Tests/AsyncTests/AsyncRequestTests.swift
0xTim: Yeah I cut it down significantly. There are a few other options and the best way to do this would be Async Algorithm but it's a test and this was the quickest easiest thing!

PR #3100, opened 2023-11-06
Add SendableRoute type

0xTim:
Adds a new `SendableRoute` to replace `Route`. This is designed to be a value type from the ground up which removes the need to access properties via locks in a large number of places.

Most users should be switched over automatically.


0xTim:
Closing as this will become part of Vapor 5

PR #3102, opened 2023-11-07
Fix issue when client disconnects midway through a stream

0xTim:
Fixes an issue when a client disconnects mid way through streaming a request in a Swift concurrency context. In certain cases this would trigger a de-init off the event loop, leading to a crash.

This fixes the issue by using a lock instead of a loop bound wrapper


Sources/Vapor/Request/Request+BodyStream.swift
```diff
@@ -16,13 +16,13 @@ extension Request {
             var buffer: [(BodyStreamResult, EventLoopPromise<Void>?)]
         }
 
-        private let isClosed: NIOLoopBoundBox<Bool>
+        private let isClosed: NIOLockedValueBox<Bool>
```
MahdiBM, line 19: Shouldn't an atomic be better?

gwynne: No.

gwynne:
@0xTim I'm not delighted with this solution; I'd much rather understand the underlying reason that an early disconnect causes a wrong-thread reinit.

Sources/Vapor/Request/Request+BodyStream.swift
MahdiBM: Why? 🤔

gwynne (approved):
Reluctantly.

PR #3104, opened 2023-11-15
Fix issue with AnySendableHashable in Route

0xTim:
Fixes a regression when using the `userInfo` dictionary which required the use of `AnySendableHashable`. Passing in `String`s (as opposed to string literals) would fail to compile as we don't have access to all the fancy stuff that `AnyHashable` does.

This fixes it to require `Strings` which is technically a source-breaking change but all known examples use `String` as the key and it was a potential DoS attack vector if this wasn't safe across concurrent environments.


0xTim:
Closing in favour of https://github.com/vapor/console-kit/pull/193

PR #3107, opened 2023-11-21
General warnings and tests cleanup

gwynne:
These changes fix almost all of the extant warnings in Vapor, and clean up some issues in the tests, which can now run fully parallelized. There are no functional changes.

### DIFF
--- a/Sources/Vapor/Concurrency/RequestBody+Concurrency.swift
+++ b/Sources/Vapor/Concurrency/RequestBody+Concurrency.swift
@@ -130,6 +130,7 @@ extension Request.Body: AsyncSequence {
             failureType: Error.self,
             backPressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies
                 .HighLowWatermark(lowWatermark: 5, highWatermark: 20),
+            finishOnDeinit: true,
             delegate: delegate
         )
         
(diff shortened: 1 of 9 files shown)


MahdiBM (approved):
There are still some warnings on linux, see the gql build log in CI.

Most of them are Date/Data warnings which I wouldn't bother fixing as Franz's changes should "take effect" sooner or later, but there are some NIOLockedValueBox warnings that can be resolved.

Sources/Vapor/Concurrency/RequestBody+Concurrency.swift
```diff
@@ -130,6 +130,7 @@ extension Request.Body: AsyncSequence {
             failureType: Error.self,
             backPressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies
                 .HighLowWatermark(lowWatermark: 5, highWatermark: 20),
+            finishOnDeinit: true,
```
0xTim, line 133: I wonder if this fixes our crash on connection close issue

0xTim (approved):

0xTim:
@MahdiBM those should all be fixed in 5.9 - the only one that's left is the `@Sendable` function which requires the compiler change that was pitched to be implemented

PR #3108, opened 2023-11-30
Make Middleware Async

0xTim:
Makes all of Vapor's internal middleware async so we can have a full async chain all the way down in anticipation of distributed tracing (and adopting NIO's new AsyncChannel stuff)

### DIFF
--- a/Sources/Vapor/Middleware/ErrorMiddleware.swift
+++ b/Sources/Vapor/Middleware/ErrorMiddleware.swift
@@ -3,7 +3,7 @@ import NIOCore
 import NIOHTTP1
 
 /// Captures all errors and transforms them into an internal server error HTTP response.
-public final class ErrorMiddleware: Middleware {
+public final class ErrorMiddleware: AsyncMiddleware {
     /// Structure of `ErrorMiddleware` default response.
     internal struct ErrorResponse: Codable {
         /// Always `true` to indicate this is a non-typical JSON response.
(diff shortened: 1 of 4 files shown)


Sources/Vapor/Middleware/ErrorMiddleware.swift
```diff
@@ -3,7 +3,7 @@ import NIOCore
 import NIOHTTP1
 
 /// Captures all errors and transforms them into an internal server error HTTP response.
-public final class ErrorMiddleware: Middleware {
+public final class ErrorMiddleware: AsyncMiddleware {
     /// Structure of `ErrorMiddleware` default response.
     internal struct ErrorResponse: Codable {
```
Patrick-Kladek, line 8: While you are at it, is it possible to make `ErrorResponse` public? 

In my project I use https://github.com/mattpolzin/VaporOpenAPI to generate the openApi.yml file from the server code and with this in turn the client facing structs - the server is the single source of truth for the models.

Right now I need to duplicate this struct and make it conform to `Sampleable & OpenAPIExampleProvider` for this to work, which is not ideal.

I also tried writing my own ErrorMiddleware with a public ErrorResponse struct but this failed as I couldn't use `response.responseBox` as its marked internal and can only be used from the vapor module.

What would be the recommended approach here?

0xTim:
Going to close this - this will get consumed by Vapor 5

PR #3109, opened 2023-12-01
Bring back AsyncCommands

marius-se:
Brings async commands back by adding a new property `asyncCommands` to `Application`.

### DIFF
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -1,11 +1,12 @@
-#if swift(<5.9)
-import Backtrace
-#endif
+import ConsoleKit
+import Logging
 import NIOConcurrencyHelpers
 import NIOCore
-import Logging
-import ConsoleKit
 import NIOPosix
+
+#if swift(<5.9)
+    import Backtrace
+#endif
 
 /// Core type representing a Vapor application.
 public final class Application: Sendable {
@@ -17,7 +18,7 @@ public final class Application: Sendable {
             self._environment.withLockedValue { $0 = newValue }
         }
     }
-    
+
     public var storage: Storage {
         get {
             self._storage.withLockedValue { $0 }
@@ -26,13 +27,11 @@ public final class Application: Sendable {
             self._storage.withLockedValue { $0 = newValue }
         }
     }
-    
+
     public var didShutdown: Bool {
-        get {
-            self._didShutdown.withLockedValue { $0 }
-        }
+        self._didShutdown.withLockedValue { $0 }
     }
-    
+
     public var logger: Logger {
         get {
             self._logger.withLockedValue { $0 }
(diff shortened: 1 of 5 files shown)


marius-se:
I just saw my formatter went for a little rampage in `Application.swift`, I hope that's not a problem :D

Sources/Vapor/Application.swift
```diff
@@ -163,18 +161,31 @@ public final class Application: Sendable {
...
+        let combinedCommands = asyncCommands.merge(
+            with: commands,
+            defaultCommand: commands.defaultCommand,
+            help: commands.help
+        )
+
         var context = CommandContext(console: self.console, input: self.environment.commandInput)
         context.application = self
-        try self.console.run(command, with: context)
+        try self.eventLoopGroup.next()
```
0xTim, line 184:
```suggestion
        try self.eventLoopGroup.any()
```

Sources/Vapor/Core/Core.swift
```diff
@@ -77,6 +83,7 @@ extension Application {
                 var commands = Commands()
                 commands.use(BootCommand(), as: "boot")
                 self.commands = .init(commands)
+                self.asyncCommands = .init(AsyncCommands())
```
0xTim, line 86: Tempted to say we should switch all the internals over but I think that can wait

0xTim (changes requested):
One nit and we need to update both package manifests to point to 4.13.0

Sources/Vapor/Core/Core.swift
marius-se: Yeah I was also thinking about that, but I'd say that belongs to a separate PR. 👍🏼

Tests/AsyncTests/AsyncCommandsTests.swift
```diff
@@ -0,0 +1,28 @@
...
+            let help = "Does the foo."
+
+            func run(using context: CommandContext, signature: Signature) throws {
+                context.console.output("Hello, \(signature.name)!")
+            }
+        }
+
+        app.asyncCommands.use(FooCommand(), as: "foo")
+
+        app.environment.arguments = ["vapor", "foo", "bar"]
```
0xTim, line 24: Shouldn't this be
```suggestion
        app.environment.arguments = ["vapor", "foo", "--name", "bar"]
```

Tests/AsyncTests/AsyncCommandsTests.swift
```diff
@@ -0,0 +1,28 @@
...
+
+        XCTAssertNoThrow(try app.start())
```
0xTim, line 26: Is there any way we can verify it runs, like incrementing a global counter? A manual run will be good enough if you haven't done that. (Do we have any tests for normal commands)

Tests/AsyncTests/AsyncCommandsTests.swift
marius-se: No it's an `@Argument` not `@Option`

Tests/AsyncTests/AsyncCommandsTests.swift
marius-se: Good idea, I just pushed a commit. Unfortunately we don't have any tests for commands (or I couldn't find one)

0xTim (approved):
🥳

PR #3111, opened 2023-12-03
Allow to define handlers without response

xeraph-dev:
When writing RESTful APIs, it's useful to have response-less handlers that just have the No Content status code.

### DIFF
--- a/Sources/Vapor/Concurrency/RoutesBuilder+Concurrency.swift
+++ b/Sources/Vapor/Concurrency/RoutesBuilder+Concurrency.swift
@@ -27,6 +27,24 @@ extension RoutesBuilder {
     
     @discardableResult
     @preconcurrency
+    public func get(
+        _ path: PathComponent...,
+        use closure: @Sendable @escaping (Request) async throws -> Void
+    ) -> Route {
+        return self.on(.GET, path, use: closure)
+    }
+    
+    @discardableResult
+    @preconcurrency
+    public func get(
+        _ path: [PathComponent],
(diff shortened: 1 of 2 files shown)


MahdiBM:
You can return a `HTTPStatus` from your route if you want the response to only contain an status code and no body.
Try something like `return HTTPStatus.noContent` in your route.

gwynne:
Among other things, because `204 No Content` is not an appropriate response to any arbitrary request (see [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110) § [9.3.4](https://datatracker.ietf.org/doc/html/rfc9110#name-put), [9.3.5](https://datatracker.ietf.org/doc/html/rfc9110#name-delete), [15.3.1](https://datatracker.ietf.org/doc/html/rfc9110#name-200-ok), and [15.3.5](https://datatracker.ietf.org/doc/html/rfc9110#name-204-no-content), among others).

xeraph-dev:
> Among other things, because `204 No Content` is not an appropriate response to any arbitrary request (see [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110) § [9.3.4](https://datatracker.ietf.org/doc/html/rfc9110#name-put), [9.3.5](https://datatracker.ietf.org/doc/html/rfc9110#name-delete), [15.3.1](https://datatracker.ietf.org/doc/html/rfc9110#name-200-ok), and [15.3.5](https://datatracker.ietf.org/doc/html/rfc9110#name-204-no-content), among others).

That's a really good resource. Thanks

PR #3113, opened 2023-12-05
Fix setting public folder for `FileMiddleware` when using bundles

grantjbutler:
This PR fixes an issue where, if you provided a subfolder within a bundle's resources, the wrong path would be provided to the `FileMiddleware`, causing the resources to not be loaded.

For example, given a bundle with the following structure:

```
App.app/
└── Contents/
    ├── MacOS/
    │   └── App
    └── Resources/
        └── web-app/
            └── Public
                └── index.html
```

If you tried to create an instance of `FileMiddleware` that tried to use `web-app/Public/` as the folder to serve files from, `FileMiddleware` would incorrectly use the resource path of the bundle (`App.app/Resources/`) instead of the full path to the specified folder (`App.app/Resources/web-app/Public/`).

### DIFF
--- a/Sources/Vapor/Middleware/FileMiddleware.swift
+++ b/Sources/Vapor/Middleware/FileMiddleware.swift
@@ -118,7 +118,7 @@ public final class FileMiddleware: Middleware {
             throw BundleSetupError.publicDirectoryIsNotAFolder
         }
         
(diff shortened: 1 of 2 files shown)


dannflor (approved):

gwynne (approved):
Nice catch!

PR #3118, opened 2023-12-11
Allow parsing of non-ASCII URIs

baarde:
While the HTTP parser rejects URLs containing unescaped non-ASCII characters, `URI` values may be constructed from other sources and the current implementation allows any String to be used as an `URI`. We should therefore be ready to safely parse an `URI` containing non-ASCII characters.

`vapor_urlparser_parse` operates on bytes while `URI` treats string as a collection of `Character`. This causes parsed components to be shifted toward the end.

For example, `http://test/😀?👍#12345678` currently parses as:
* scheme: `http`
* host: `test`
* path: `/😀?👍`
* query: `2345`
* fragment: `78`

 This PR computes the buffer length and component ranges using the string's UTF8 view.

### DIFF
--- a/Sources/Vapor/Utilities/URI.swift
+++ b/Sources/Vapor/Utilities/URI.swift
@@ -200,7 +200,8 @@ public struct URI: Sendable, ExpressibleByStringInterpolation, CustomStringConve
 
     private func parse(_ component: Component) -> String? {
         var url = vapor_urlparser_url()
-        vapor_urlparser_parse(self.string, self.string.count, 0, &url)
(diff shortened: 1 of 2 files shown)


0xTim:
@baarde I think this is now fixed by the migration to the Swift parser - happy to accept the tests in a PR to confirm though if you want?

baarde:
Thanks for the update. Invalid URLs are now converted to an empty `URI`, so handling non-ASCII strings is not an issue anymore. I don't think we need to add specific tests. I'll close this PR as it's no longer relevant.

PR #3119, opened 2023-12-11
Preserve empty path when constructing a URI

baarde:
### Motivation

When constructing a URI from its component, an empty path is replaced by `"/"`.

This means that any of `URI`'s setters may update the path:

```swift
var uri = URI(string: "http://foo")
uri.query = "bar"
// uri is now "http://foo/?bar"
```

This also means that while an `URI`'s path may be empty, trying to set a `URI`'s path to the empty string actually sets the path to `"/"`.

```swift
var uri = URI(string: "http://foo")
uri.path = ""
// uri.path is now "/"
```

This PR adds an extra check to preserve the empty path.


baarde:
This is no longer relevant since 6db3d917b5ce5024a84eb265ef65691383305d70.

Closing the PR.

PR #3126, opened 2024-01-03
HTTP2 Response Compression/Request Decompression

### LINKED ISSUE #3125: Response compression handler not added for HTTP2 requests
### Describe the bug

I noticed the response compression didn't work for HTTP2 requests, because the `addVaporHTTP2Handlers()` method doesn't seem to add the handlers like the HTTP1 handler does.

### To Reproduce

1. Configure an app with TLS and the following lines:
```swift
    app.http.server.configuration.responseCompression = .enabled
    app.http.server.configuration.supportVersions = [.one, .two]
```

2. Access the route via Safari
3. Notice requests aren't compressed. Removing the `supportedVersions` line (and thus HTTP2) enables compression as expected.

HTTP1.1:
image

HTTP2:
image

### Expected behavior

Responses should be compressed when accessed over typical browsers.

### Environment

* Vapor Framework version: 4.87.1
* Vapor Toolbox version: 18.7.4
* OS version: macOS 14.1.2

dimitribouniol:
Fixed an issue where HTTP2 didn't support response compression and request decompression.

It seems like it may have been omitted when adding explicit support for HTTP2. Not sure what to do about testing as I couldn't find any tests for the HTTP1.1 pathway, but I did verify it works in my pet project 😅 

Fixes #3125

### DIFF
--- a/Tests/VaporTests/ServerTests.swift
+++ b/Tests/VaporTests/ServerTests.swift
@@ -264,7 +264,7 @@ final class ServerTests: XCTestCase {
             struct Nothing: Codable {}
             XCTAssertNoThrow(try JSONDecoder().decode(Nothing.self, from: body))
         } else {
-            XCTFail()
+            XCTFail("Missing response.body")
         }
     }
     
@@ -277,9 +277,9 @@ final class ServerTests: XCTestCase {
         let smallBody = ByteBuffer(base64String: "H4sIAAAAAAAAE/NIzcnJ11Eozy/KSVEEAObG5usNAAAA")! // "Hello, world!"
         let bigBody = ByteBuffer(base64String: "H4sIAAAAAAAAE/NIzcnJ11HILU3OgBBJmenpqUUK5flFOSkKJRmJeQpJqWn5RamKAICcGhUqAAAA")! // "Hello, much much bigger world than before!"
         
-        // Max out at the smaller payload (.size is of compressed data)
+        // Max out at the smaller payload (.size is of uncompressed data)
         app.http.server.configuration.requestDecompression = .enabled(
-            limit: .size(smallBody.readableBytes)
+            limit: .size(smallOrigString.utf8.count)
         )
         app.post("gzip") { $0.body.string ?? "" }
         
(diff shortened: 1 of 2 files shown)


gwynne:
I couldn't tell you why for the life of me, but IIRC this omission was deliberate.

@0xTim @Lukasa Can you provide any insights here?

0xTim:
Probably just missed when we added compression support. I can't think of anything against it

dimitribouniol:
Linking https://github.com/vapor/vapor/pull/2214 which seems to have been when it was modernized, though HTTP2 didn't have any support then either.

dimitribouniol:
Rebased the merge conflict — this has been working well for me over the past few weeks of use, and could really help reduce the amount of data transmitted when large JSON objects are sent out over HTTP/2.

dimitribouniol:
Since it's been a month, figured I'd check in to see if there are still unknowns to figure out here.

Tests/VaporTests/ServerTests.swift
```diff
@@ -314,6 +314,364 @@ final class ServerTests: XCTestCase {
...
+            request.body = ByteBuffer(data: compressedPayload)
+        }.wait()
+        
+        if let body = unsupportedCompressedResponse.body {
+            let decodedResponse = try JSONDecoder().decode(TestResponse.self, from: body)
+            XCTAssertEqual(decodedResponse.contentString, String(decoding: compressedPayload, as: UTF8.self))
+            XCTAssertEqual(decodedResponse.contentLength, compressedPayload.count)
+        } else {
+            XCTFail()
+        }
```
dimitribouniol, line 382: Probably better as a follow-up PR, but a question regarding this currently working behavior: Should this actually be rejected, because the server explicitly doesn't support request decompression? I can file an issue if there is agreement here.

gwynne: I would have thought this would be rejected. @0xTim Any thoughts?

0xTim: Yeah I think the server should reject the request in this case

0xTim: We should probably return a 415 in this case

Tests/VaporTests/ServerTests.swift
```diff
@@ -314,6 +314,364 @@ final class ServerTests: XCTestCase {
...
+        
+        if let body = unsupportedNoncompressedResponse.body {
+            let decodedResponse = try JSONDecoder().decode(TestResponse.self, from: body)
+            XCTAssertEqual(decodedResponse.content, compressedPayload)
+            XCTAssertEqual(decodedResponse.contentLength, compressedPayload.readableBytes)
+        } else {
+            XCTFail("Missing unsupportedNoncompressedResponse.body")
+        }
+        
+        // TODO: The server should probably reject this?
```
0xTim, line 370: Yeah I agree with this. Though it's not the end of the world. I do think we should probably default to enabled for request decompression, what do you think @gwynne ?

0xTim (approved):
Looks good to me! I agree we should reject the request if it's compressed and it's been disabled but I also think we should enable request decompression by default. Happy to discuss this in a new PR though

Tests/VaporTests/ServerTests.swift
gwynne: Agreed; as far as I know, there's no reason not to allow it by default, notwithstanding such bugs as https://github.com/swift-server/async-http-client/issues/739

dimitribouniol: Funnily enough, I wasted probably close to an hour due to that default ratio of 10 wondering why the tests were failing with my test strings. The .size constraint is relative to the _decompressed_ size, so it might make more sense to have that be the default constraint for decompression on both the server and client.

PR #3129, opened 2024-01-05
fix: problem of '%2B' being decoded into ' ' in URL

Raymond-HuangL:

### DIFF
--- a/Sources/Vapor/URLEncodedForm/URLEncodedFormDecoder.swift
+++ b/Sources/Vapor/URLEncodedForm/URLEncodedFormDecoder.swift
@@ -111,7 +111,7 @@ public struct URLEncodedFormDecoder: ContentDecoder, URLQueryDecoder {
     ///   - url: ``URI`` to read the query string from
     ///   - userInfo: Overrides the default coder user info
     public func decode<D>(_ decodable: D.Type, from url: URI, userInfo: [CodingUserInfoKey: Any]) throws -> D where D : Decodable {
-        try self.decode(D.self, from: url.query ?? "", userInfo: userInfo)
+        try self.decode(D.self, from: url.percentEncodedQuery ?? "", userInfo: userInfo)
     }
 
     /// Decodes an instance of the supplied ``Decodable`` type from a ``String``.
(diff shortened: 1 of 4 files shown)


gwynne:
The conversion of `+` to `%20` is correct for the modern specification of URIs (per RFC 3986); `+` has always been a legacy encoding. Converting it to `%2B` is re-encoding, not decoding. Can you be more detailed about the problem you're trying to solve?

dimitribouniol:
> I have a url parameter like 'number=+123'. When I read using request.query[String.self, at: "number"], I get ' 123', but I need '+123'. I tried 'number=%2B123', same result. Changing this can meet my needs

Would it be better to add a configurable decoder for these, then the default can remain, but this behavior can be turned off if needed?

gwynne:
Your input is incorrectly encoded by spec; your parameter should be written as `number=%2B123`

gwynne:
(to be 100% precise, by spec Vapor's decoding should not be translating `+` to ` ` at all - or anything else, for that matter - but this behavior is necessary to maintain legacy compatibility; we can't drop it until Vapor 5)

Raymond-HuangL:
I understand why I can't modify this, and I noticed that the fundamental problem is that `%2B` has been decoded into ` `. The input parameter `query` of this method `func parse(_ query: String)`, should be an encoded query string. So, I reverted the commit and modified the place where the query is passed in. After that, everything is satisfactory.

'number=+123' will be decoded to ' 123'.
'number=%2B123' will be decoded to '+123'.

0xTim:
@gwynne was this resolved in #3140

gwynne:
> @gwynne was this resolved in #3140

This PR is unrelated; the OP is asking for an incompatible (and spec-violating) change to preexisting semantics which were not affected by the URI issues.

gwynne:
@0xTim Ignore my former comments. Yes, this was superseded by #3140; I wasn't tracking it as a URI-related issue as the PR was originally trying to make an incompatible change (and I hadn't seen the subsequent update).

gwynne:
@Raymond-HuangL Can you confirm that https://github.com/vapor/vapor/releases/tag/4.92.0 resolved this problem for you?

Raymond-HuangL:
Yes, I see my problem is solved in this Vapor 4.92.0, thank you all

PR #3132, opened 2024-01-07
Allow `HTTPServer`'s configuration to be dynamically updatable

### LINKED ISSUE #3130: Updating HTTP server configuration on the fly (ie. TLS settings)
I would like to dynamically update some of the HTTP server's configuration on the fly, specifically TLS settings like with an updated certificate, but without bringing down the server and dropping all clients. It seems like this would be straightforward to implement since the `HTTPServerConnection` configures child handlers as they come in, so this configuration could be dynamic, but I had some questions and wanted to make sure the idea jived with the project before pushing a PR 😅 

### Details

Unfortunately, `Application.http.server.configuration` becomes read-only once a server starts, despite some configuration options still being configurable for new connections or other non-server-start tasks:
- `responseCompression`
- `requestDecompression`
- `supportPipelining`
- `supportVersions`
- `tlsConfiguration`
- `serverName`
- `reportMetrics`
- `logger`
- `shutdownTimeout`

Additionally, this configuration is cached by `HTTPServerConnection`'s various closures, which would need to be updated as well.

I propose adding the following public API to `HTTPServer` to allow updating safe configuration options dynamically:
```swift
public func updateConfiguration(
    responseCompression: CompressionConfiguration? = nil,
    requestDecompression: DecompressionConfiguration? = nil,
    supportPipelining: Bool? = nil,
    supportVersions: Set<HTTPVersionMajor>?? = nil,
    tlsConfiguration: TLSConfiguration?? = nil,
    serverName: String?? = nil,
    reportMetrics: Bool? = nil,
    logger: Logger?? = nil,
    shutdownTimeout: TimeAmount?
) {
    var updatedConfiguration = _configuration.withLockedValue { $0 }
    
    if let responseCompression { updatedConfiguration.responseCompression = responseCompression }
    if let requestDecompression { updatedConfiguration.requestDecompression = requestDecompression }
    if let supportPipelining { updatedConfiguration.supportPipelining = supportPipelining }
    if let supportVersions { updatedConfiguration.supportVersions = supportVersions }
    if let tlsConfiguration { updatedConfiguration.tlsConfiguration = tlsConfiguration }
    if let serverName { updatedConfiguration.serverName = serverName }
    if let reportMetrics { updatedConfiguration.reportMetrics = reportMetrics }
    if let logger { updatedConfiguration.logger = logger }
    if let shutdownTimeout { updatedConfiguration.shutdownTimeout = shutdownTimeout }

    self.application.storage[Application.HTTP.Server.ConfigurationKey.self] = newValue
    _configuration.withLockedValue { $0 = newValue }
}
```
This would allow any of the safe options to be re-specified individually.

Then, `HTTPServerConnection` can fetch the configuration on-demand in its `.childChannelInitializer {}` closure:
```swift
    static func start(
        application: Application,
        server: HTTPServer,
        responder: Responder,
        configuration: HTTPServer.Configuration,
        on eventLoopGroup: EventLoopGroup
    ) -> EventLoopFuture<HTTPServerConnection> {
        let quiesce = ServerQuiescingHelper(group: eventLoopGroup)
        let bootstrap = ServerBootstrap(group: eventLoopGroup)
            // Specify backlog and enable SO_REUSEADDR for the server itself
            .serverChannelOption(ChannelOptions.backlog, value: Int32(configuration.backlog))
            .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: configuration.reuseAddress ? SocketOptionValue(1) : SocketOptionValue(0))
            
            // Set handlers that are applied to the Server's channel
            .serverChannelInitializer { channel in
                channel.pipeline.addHandler(quiesce.makeServerChannelHandler(channel: channel))
            }
            
            // Set the handlers that are applied to the accepted Channels
            .childChannelInitializer { [unowned application, unowned server] channel in
                let configuration = server.configuration
                // add TLS handlers if configured
                if var tlsConfiguration = configuration.tlsConfiguration {
```

From that point onwards, the in-scope configuration will be used for that child connection until it is closed, but new connections would grab a fresh configuration struct as needed.

Unfortunately, this introduces a few double optionals, which isn't great. An alternative would be to offer a `dynamicConfiguration` property on `HTTPServer` which would be a safe view into the underlying configuration, but that would instead require another type to be maintained along-side the main configuration type.

A final alternative could be to only warn in the `configuration` setter if an unsafe property were changed:
```swift
    public var configuration: Configuration {
        get { _configuration.withLockedValue { $0 } }
        set {
            let oldValue = _configuration.withLockedValue { $0 }
            var canBeUpdatedDynamically = true
            
            if oldValue.address != newValue.address
                || oldValue.backlog != newValue.backlog
                || oldValue.reuseAddress != newValue.reuseAddress
                || oldValue.tcpNoDelay != newValue.tcpNoDelay {
                canBeUpdatedDynamically = false
            }
            
            guard canBeUpdatedDynamically || !didStart.withLockedValue({ $0 }) else {
                _configuration.withLockedValue({ $0 }).logger.warning("Cannot modify server configuration after server has been started.")
                return
            }
            self.application.storage[Application.HTTP.Server.ConfigurationKey.self] = newValue
            _configuration.withLockedValue { $0 = newValue }
        }
    }
```
… but this version would be less permissive to forgetting to add a new configuration to the unsafe list, though the only downside would be not seeing the warning and not getting the desired effect, so maybe that's not too bad?

Anywho, curious to hear which we would prefer. In the meantime I'll fork and work on a POC.

dimitribouniol:
This allows many aspects of the HTTP server configuration to be changed after the server starts without needing to stop and restart it, or drop existing connections in the process.

Some things that can now be re-configured include request/response configuration options, HTTP version support, HTTP pipelining, TLS configuration (ie. enabling/disabling, rotating certificates, etc…), server name, metrics reporting, the logger, and the shutdown timer.

Fixes #3130.

### DIFF
--- a/Sources/Vapor/HTTP/Server/HTTPServer.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServer.swift
@@ -297,15 +297,18 @@ public final class HTTPServer: Server, Sendable {
         var configuration = self.configuration
         
         switch address {
-        case .none: // use the configuration as is
+        case .none: 
+            /// Use the configuration as is.
             break
-        case .hostname(let hostname, let port): // override the hostname, port, neither, or both
+        case .hostname(let hostname, let port): 
+            /// Override the hostname, port, neither, or both.
             configuration.address = .hostname(hostname ?? configuration.hostname, port: port ?? configuration.port)
-        case .unixDomainSocket: // override the socket path
+        case .unixDomainSocket: 
+            /// Override the socket path.
             configuration.address = address!
         }
         
-        // print starting message
+        /// Print starting message.
         let scheme = configuration.tlsConfiguration == nil ? "http" : "https"
         let addressDescription: String
         switch configuration.address {
@@ -317,7 +320,7 @@ public final class HTTPServer: Server, Sendable {
         
         self.configuration.logger.notice("Server starting on \(addressDescription)")
 
-        // start the actual HTTPServer
+        /// Start the actual `HTTPServer`.
         try self.connection.withLockedValue {
             $0 = try HTTPServerConnection.start(
                 application: self.application,
@@ -370,23 +373,23 @@ private final class HTTPServerConnection: Sendable {
     ) -> EventLoopFuture<HTTPServerConnection> {
         let quiesce = ServerQuiescingHelper(group: eventLoopGroup)
         let bootstrap = ServerBootstrap(group: eventLoopGroup)
-            // Specify backlog and enable SO_REUSEADDR for the server itself
+            /// Specify backlog and enable `SO_REUSEADDR` for the server itself.
             .serverChannelOption(ChannelOptions.backlog, value: Int32(configuration.backlog))
             .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: configuration.reuseAddress ? SocketOptionValue(1) : SocketOptionValue(0))
             
-            // Set handlers that are applied to the Server's channel
+            /// Set handlers that are applied to the Server's channel.
             .serverChannelInitializer { channel in
                 channel.pipeline.addHandler(quiesce.makeServerChannelHandler(channel: channel))
             }
             
-            // Set the handlers that are applied to the accepted Channels
+            /// Set the handlers that are applied to the accepted Channels.
             .childChannelInitializer { [unowned application, unowned server] channel in
                 /// Copy the most up-to-date configuration.
                 let configuration = server.configuration
 
-                // add TLS handlers if configured
+                /// Add TLS handlers if configured.
                 if var tlsConfiguration = configuration.tlsConfiguration {
-                    // prioritize http/2
+                    /// Prioritize http/2 if supported.
                     if configuration.supportVersions.contains(.two) {
                         tlsConfiguration.applicationProtocols.append("h2")
                     }
@@ -434,7 +437,7 @@ private final class HTTPServerConnection: Sendable {
                 }
             }
             
-            // Enable TCP_NODELAY and SO_REUSEADDR for the accepted Channels
+            /// Enable `TCP_NODELAY` and `SO_REUSEADDR` for the accepted Channels.
             .childChannelOption(ChannelOptions.socket(IPPROTO_TCP, TCP_NODELAY), value: configuration.tcpNoDelay ? SocketOptionValue(1) : SocketOptionValue(0))
             .childChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: configuration.reuseAddress ? SocketOptionValue(1) : SocketOptionValue(0))
             .childChannelOption(ChannelOptions.maxMessagesPerRead, value: 1)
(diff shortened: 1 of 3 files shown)


dimitribouniol:
Gentle poke to see if there's anything I can help with to get this over the finish line 😊

dimitribouniol:
Just wanted to check in again to see if there's anything holding this from being merged in — if it's a migration to async channels, I can dedicate some time to making that happen as I'm highly motivated to no longer rely on a private fork of vapor for my production deploy 😅

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -357,7 +380,10 @@ private final class HTTPServerConnection: Sendable {
             }
             
             // Set the handlers that are applied to the accepted Channels
-            .childChannelInitializer { [unowned application] channel in
+            .childChannelInitializer { [unowned application, unowned server] channel in
+                /// Copy the most up-to-date configuration.
```
0xTim, line 384: Nit - this doesn't need to be a doc comment
```suggestion
                // Copy the most up-to-date configuration.
```

Tests/VaporTests/ServerTests.swift
```diff
@@ -967,6 +966,90 @@ final class ServerTests: XCTestCase {
...
+            request: try HTTPClient.Request(
+                url: "https://\(ip):\(port)/hello",
+                method: .GET
+            )
+        ).wait()
+        XCTAssertEqual(b.headers[.server], ["New"])
+        XCTAssertEqual(b.body, ByteBuffer(string: "world"))
+        
+        /// Non-TLS request should now fail
+        let c = try? app.http.client.shared.execute(
```
0xTim, line 1044: Using an `XCTAssertThrows` may be better and checking the actual error in case we're missing the server actually being down etc

0xTim (changes requested):
@dimitribouniol sorry for the delay! I have no objections to allowing the suggested options to be reloaded at runtime. 

@gwynne any thoughts?

Tests/VaporTests/ServerTests.swift
dimitribouniol: Good call, fixed!

Sources/Vapor/HTTP/Server/HTTPServer.swift
gwynne: I disagree with Tim here - I also use doc comments most of the time (though more for the inline formatting than for a font per se xD), I'd be more inclined to change the existing comments rather than the new one 🙂

dimitribouniol: Happy to revert them, let me know 🤣

dimitribouniol: Updated! Let me know if you'd rather them in a separate PR for commit history reasons, otherwise good to merge if you are happy with it.

0xTim (approved):
Thanks!

0xTim:
The failing test is a JWT issue that will be resolved tomorrow

PR #3146, opened 2024-01-30
Delete socket file if it exists. Fixes #3085.

### LINKED ISSUE #3085: If Unix domain socket file exists then Vapor crashes when starting instead of taking over the socket file
### Describe the bug

In the app's `configure.swift`, `func configure()`, set to use Unix domain socket (UDS):

```
app.http.server.configuration.address = .unixDomainSocket(path: unixDomainSocket)
```

If the UDS file exists then the app crashes upon start.

### To Reproduce

- Create an app to use UDS.
- Start the app.
- Verify that the app created the UDS in the specified path.
- Kill the app so that it doesn't shut down cleanly.
- Verify that the app's UDS still exists.
- Start the app and see that it fails: ` bind(descriptor:ptr:bytes:): Address already in use (errno: 98) (Vapor/Application.swift:162)`

### Expected behavior

Vapor removes the old UDS file and recreates it. As per @gwynne on Discord, the bug is at https://github.com/vapor/vapor/blob/main/Sources/Vapor/HTTP/Server/HTTPServer.swift#L421:

`channel = bootstrap.bind(unixDomainSocketPath: socketPath)`

And the fix:

`channel = bootstrap.bind(unixDomainSocketPath: socketPath, cleanupExistingSocketFile: true)`

Tested working on my machine.

### Environment

OS = Ubuntu 22.04 aarch64
Swift v5.9
```
    {
      "identity" : "vapor",
      "kind" : "remoteSourceControl",
      "location" : "https://github.com/vapor/vapor.git",
      "state" : {
        "revision" : "036d67e4da46126810f56c6a6ce813bcf259a745",
        "version" : "4.84.1"
      }
    },
```

talmeme:
#3085: If Unix domain socket file exists then Vapor crashes when starting instead of taking over the socket file.

This PR fixes by deleting the socket file if it exists.

### DIFF
--- a/Sources/Vapor/HTTP/Server/HTTPServer.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServer.swift
@@ -418,7 +418,7 @@ private final class HTTPServerConnection: Sendable {
         case .hostname:
             channel = bootstrap.bind(host: configuration.hostname, port: configuration.port)
         case .unixDomainSocket(let socketPath):
-            channel = bootstrap.bind(unixDomainSocketPath: socketPath)
+            channel = bootstrap.bind(unixDomainSocketPath: socketPath, cleanupExistingSocketFile: true)
         }
         
         return channel.map { channel in


0xTim (approved):
Thanks! Sorry for the delay in this - do you think it would be possible to write a test for this?

Naveen-C-Ramachandrappa (approved):

Issue #2749, opened 2021-12-17
Reverse Pull #2492


trametheka:
### Describe the bug

Edit - Whopps, 2492 is the issue, Pull is:
https://github.com/vapor/vapor/pull/2614

Pull #2492 should, in my opinion at least, be reversed. In general, HEAD requests are expected to represent a GET request without a body:
https://datatracker.ietf.org/doc/html/rfc2616/#section-9.4

I can see why the push was made, however it breaks HTTP specifications and what I would consider expected behaviour for 99.9% of Vapor users. In the users case who made the pull, they should be handling the .HEAD request themself if they need the HEAD to be different to a GET's headers. There is nothing in the specifications that a HEAD request cannot have request parameters in the URL.

This change is also creating inconsistent behaviour between routes when a HEAD request is made. For example a static route of "/hello" returns .OK regardless of the GET route, where as a dynamic route of "/hello/:name" is (correctly) forwarded to GET request.

This change also means that routes are bound to an order. Defining a route.on(.HEAD) before a route.on(.GET) will cause the .HEAD route to be overwritten.

### To Reproduce

Take for example a route of https://localhost/download?file=thefile.txt

A HEAD request to this route would expect a content-length to be returned so the requester, for example a browser, would know how much is to be downloaded. As by default this GET request has a .OK HEAD responder generated for it, that content-length wouldn't be included.

### Expected behavior

- A HEAD request returns the same headers as GET request in the absence of a HEAD route
- HEAD routes work regardless of order they are defined with other routes.

trametheka:
I assume this would fix https://github.com/vapor/vapor/issues/2680 as well?

baarde:
I agree. The response to a HEAD request should be identical to that to a GET request. A developer may implement custom routes for HEAD but the default behaviour should be to forward to GET.

PR #3147, opened 2024-01-30
Remove HeadResponder

### LINKED ISSUE #2680: HEAD route ignores Middlewares
### Describe the bug

DefaultResponder will create a HEAD route for any constant-path GET route by default. But this behaviour prohibits developers from inserting middleware to GET route without differentiating the behaviour of GET and HEAD.

```swift
// If the route isn't explicitly a HEAD route,
// and it's made up solely of .constant components,
// register a HEAD route with the same path
if route.method == .GET &&
route.path.allSatisfy({ component in
    if case .constant(_) = component { return true }
    return false
}) {
    let headRoute = Route(
        method: .HEAD,
        path: cached.route.path,
        responder: middleware.makeResponder(chainingTo: HeadResponder()),
        requestType: cached.route.requestType,
        responseType: cached.route.responseType)
    
    let headCachedRoute = CachedRoute(route: headRoute, responder: middleware.makeResponder(chainingTo: HeadResponder()))
    
    router.register(headCachedRoute, at: [.constant(HTTPMethod.HEAD.string)] + path)
}
```

### To Reproduce

1. Create a GET route with a Middleware group that inserts additional headers.

```swift
public struct RandomHeaderMiddleware: Middleware {
    public func respond(to request: Request, chainingTo next: Responder) -> EventLoopFuture<Response> {
        next.respond(to: request).map {
            response.headers.add(name: "x-random", value: "\(Int.random(in: 0...100))")
            return response
        }
    }
}
```

```swift
routes.grouped(RandomHeaderMiddleware()).get {
    return "Hello World!"
}
```

2. When a HEAD request is sent to the endpoint, ```x-random``` header will not exist. By contrast, when a GET request is sent, it will exist on the response.

### Expected behavior

1. Middleware will be accessible to HEAD method as well.
2. HEAD and GET methods will respond with same headers.

### Environment

* Vapor Framework version: 4.48.3
* Vapor Toolbox version: 18.3.3
* OS version: macOS Big Sur 11.5.2

### Additional context

If the fix I suggested is not feasible, I think at least there should be an option to not auto-generate HEAD route when the developer needs. Currently, there isn't one without making dynamic path with wildcards or parameters.

I could just override the HEAD route, but it feels like an undefined behaviour that needs to be amended.

baarde:
The HEAD method is identical to GET except that the server must not send content in the response ([RFC 9110, section 9.3.2](https://datatracker.ietf.org/doc/html/rfc9110#HEAD)).

The previous default behaviour of returning 200 OK to every HEAD request to a constant route is not standard-compliant.

The new behaviour is to always forward the request to the GET route, unless the developer explicitely configured a custom HEAD route.

This PR fixes #2680 and #2749.

### DIFF
--- a/Sources/Vapor/Responder/DefaultResponder.swift
+++ b/Sources/Vapor/Responder/DefaultResponder.swift
@@ -39,26 +39,6 @@ internal struct DefaultResponder: Responder {
                 }
             }
             
-            // If the route isn't explicitly a HEAD route,
-            // and it's made up solely of .constant components,
-            // register a HEAD route with the same path
-            if route.method == .GET &&
-                route.path.allSatisfy({ component in
-                    if case .constant(_) = component { return true }
-                    return false
-            }) {
-                let headRoute = Route(
-                    method: .HEAD,
-                    path: route.path,
-                    responder: middleware.makeResponder(chainingTo: HeadResponder()),
-                    requestType: route.requestType,
-                    responseType: route.responseType)
-
-                let headCachedRoute = CachedRoute(route: headRoute, responder: middleware.makeResponder(chainingTo: HeadResponder()))
-
-                router.register(headCachedRoute, at: [.constant(HTTPMethod.HEAD.string)] + path)
-            }
-            
             router.register(cached, at: [.constant(route.method.string)] + path)
         }
         self.router = router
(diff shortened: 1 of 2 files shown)


gwynne (approved):
LGTM, @0xTim?

0xTim (changes requested):
Sorry for the delay in this. I think this is good, but I'd like to see a test for the explicit override of a HEAD request

baarde:
@0xTim I think this is already covered by ~~`testExplicitHeadRouteOverridesGeneratedHandler`~~ `testExplicitHeadRouteOverridesForwardingToGet` (I've renamed the method to reflect the implementation change).

0xTim (approved):
Ah I missed that. Ok this looks fine to me

@owainhunt since you did the original PR, does the overriding as described work for you?

owainhunt:
LGTM 👍

PR #3156, opened 2024-03-07
Fix EndpointCache

MarcusSmith:
EndpointCache's `get(on: Request)` convenience function was calling through directly to `download`, which was completely bypassing the cache. Updated this function to call through to `get(using: Client, on: EventLoop)` instead which should check cache correctly. 

Before this get merged, anyone else running into this can work around the bug by calling `cache.get(using: request.client, on: request.eventLoop)` instead of `cache.get(on: request)`

### DIFF
--- a/Sources/Vapor/HTTP/EndpointCache.swift
+++ b/Sources/Vapor/HTTP/EndpointCache.swift
@@ -37,7 +37,7 @@ public final class EndpointCache<T>: Sendable where T: Decodable & Sendable {
     ///   - request: The `Request` which is initiating the download.
     ///   - logger: An optional logger
     public func get(on request: Request, logger: Logger? = nil) -> EventLoopFuture<T> {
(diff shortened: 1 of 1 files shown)


gwynne (changes requested):
At a glance, this does look correct to me, but I'm not fully familiar with this particular logic; @0xTim can you chime in?

In the meantime, @MarcusSmith can you add a test to `EndpointCacheTests.swift` for this issue?

MarcusSmith:
Is there a recommended way to get ahold of a `Request` to pass into it?

Best I've been able to come up with is define a new middleware that uses an endpoint cache for something and trigger that multiple times through a get request and assert that it only calls that cache's endpoint once.

Just did a live test of our hackaround to use the other init and it fixed our production issue, so I'm pretty confident this was the issue

0xTim:
@MarcusSmith you can stub out the client with something similar to #2928 to make it possible to make assertions on that (and avoid any network requests)

PR #3158, opened 2024-03-12
Fix some Sendable warnings on 5.10

sidepelican:
Fix a number of warnings in Swift 5.10 like below.

Fix simple issues that can be addressed by simply adding `Sendable`.

### DIFF
--- a/Sources/Development/routes.swift
+++ b/Sources/Development/routes.swift
@@ -268,7 +268,7 @@ public func routes(_ app: Application) throws {
         return [cred1]
     }
     
-    func opaqueRouteTester(_ req: Request) async throws -> some AsyncResponseEncodable {
+    @Sendable func opaqueRouteTester(_ req: Request) async throws -> some AsyncResponseEncodable {
         "Hello World"
     }
     asyncRoutes.get("opaque", use: opaqueRouteTester)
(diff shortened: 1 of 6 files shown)


Sources/Development/routes.swift
```diff
@@ -268,7 +268,7 @@ public func routes(_ app: Application) throws {
         return [cred1]
     }
     
-    func opaqueRouteTester(_ req: Request) async throws -> some AsyncResponseEncodable {
+    @Sendable func opaqueRouteTester(_ req: Request) async throws -> some AsyncResponseEncodable {
```
0xTim, line 271: This unfortunately masks any issues as is the incorrect usage for functions (it's confusing and very counter intuitive) - could you remove?

0xTim (changes requested):
Just need to unwind one of the route tests but otherwise LGTM!

Sources/Development/routes.swift
sidepelican: Ok. Removed. https://github.com/vapor/vapor/pull/3158/commits/e806eafb682f7a1030911293b548dc4cce3df3fb

0xTim (approved):
Thanks!

PR #3160, opened 2024-03-13
Patch configuration and log actual port on startup

### LINKED ISSUE #3159: Log actual port when it's picked by the OS
### Describe the issue

Startup log message prints configured port instead of actual one.

### Vapor version

4.92.4

macOS 14.3.1

### Swift version

Swift Package Manager - Swift 5.10.0-dev

### Steps to reproduce

1. Start the server on port 0. This lets the OS pick an available port for you.
2. Observe the console. It has a line: `[Vapor] Server starting on http://127.0.0.1:0`.
3. Inspect `app.http.server.shared.localAddress!.port` to see the port decided by the OS.

### Outcome

The log message indicates that the server is running on port 0 which is impossible. It should say the actual port.

### Additional notes

One day I will be a bieautiful butterfly!

bisgardo:
Before this change, the application

```swift
defer { app.shutdown() }
try app.server.start(hostname: nil, port: 0)
defer { app.server.shutdown() }
```

would log the following message *before* starting the server:

```
[Vapor] Server starting on http://127.0.0.1:0
```

After this change it instead logs a message like the following *after* starting the server:

```
[Vapor] Server starting on http://127.0.0.1:57935
```

The input configuration is also patched such that `app.http.server.configuration.port` will hold the actual port after startup. Currently if it has value 0 it will keep that value (only `app.http.server.shared.localAddress?.port` will have the correct one).

Fixes #3159.


bisgardo:
I have no experience with this code base so I'm just guessing that it's safe to force unwrap `self.localAddress` after the server has successfully started.

I also realize that the log message is now less readable and doesn't include the protocol. This was just the simplest possible solution I could come up with. I'll be happy to change it.

Looking forward to your feedback.

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -331,6 +319,9 @@ public final class HTTPServer: Server, Sendable {
             ).wait()
         }
 
+        /// Print starting message.
+        configuration.logger.notice("Server started on \(self.localAddress!.description)")
```
0xTim, line 323: As a general rule force unwrapping should be avoided. What I'd do is just move the whole block down and use the port from the local address and fall back to the existing behaviour if it's nil

0xTim (changes requested):

Sources/Vapor/HTTP/Server/HTTPServer.swift
0xTim: Hmm actually I remember why it's not like this currently - if it fails to bind to the specified you get no log message to help debug so we need the original message. I think we could probably change the original to an info and keep this one as notice.

What do you think @gwynne ?

bisgardo: @0xTim just pushed a commit to address your first comment, didn't see the second one until now. Isn't the correct solution to ensure that an error is logged (including the configured port) when binding fails?

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -331,6 +319,28 @@ public final class HTTPServer: Server, Sendable {
...
 
+        /// Override configuration with actual address.
+        /// This may differ from the provided configuation if port 0 was provided, for example.
+        if let localAddress = self.localAddress {
+            if let port = localAddress.port {
+                configuration.port = port
+            }
+            if let hostname = localAddress.hostname {
+                configuration.hostname = hostname
+            }
```
bisgardo, line 330: Unsure if this is relevant. Can the actual hostname differ from the configured one?

Sources/Vapor/HTTP/Server/HTTPServer.swift
bisgardo: Or maybe just wrap the thrown error with one that includes `configuration.address`? For reference, if I try to start on a port that's already in use, the error is
```
failed: caught error: "bind(descriptor:ptr:bytes:): Address already in use (errno: 48)"
```

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -331,6 +319,28 @@ public final class HTTPServer: Server, Sendable {
...
+        }
+
+        /// Print starting message.
```
bisgardo, line 333: Out of curiosity, may I ask why comments are written with tripple slashes? Isn't that style for docstrings only?

dimitribouniol: I bet those are mine haha… Xcode renders them in a proportional font with formatting by default, which is easier on the eyes for prose comments

Sources/Vapor/HTTP/Server/HTTPServer.swift
dimitribouniol: My memory is spotty here, but the configuration, set in code, can be different than the listening port passed in the command line with the `listen` command

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -308,18 +308,6 @@ public final class HTTPServer: Server, Sendable {
...
-        let scheme = configuration.tlsConfiguration == nil ? "http" : "https"
-        let addressDescription: String
-        switch configuration.address {
-        case .hostname(let hostname, let port):
-            addressDescription = "\(scheme)://\(hostname ?? configuration.hostname):\(port ?? configuration.port)"
-        case .unixDomainSocket(let socketPath):
-            addressDescription = "\(scheme)+unix: \(socketPath)"
-        }
-        
-        self.configuration.logger.notice("Server starting on \(addressDescription)")
```
0xTim, line 321: Let's keep this but at the debug level

0xTim (changes requested):

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -307,19 +307,18 @@ public final class HTTPServer: Server, Sendable {
...
-        /// Print starting message.
+
+        /// Log starting message for debugging before attempting to start the server.
         let scheme = configuration.tlsConfiguration == nil ? "http" : "https"
-        let addressDescription: String
+        var addressDescription: String
         switch configuration.address {
         case .hostname(let hostname, let port):
-            addressDescription = "\(scheme)://\(hostname ?? configuration.hostname):\(port ?? configuration.port)"
+            addressDescription = "\(scheme)://\(hostname ?? Configuration.defaultHostname):\(port ?? Configuration.defaultPort)"
```
bisgardo, line 316: Same as before but a little more concise. At least I found `hostname ?? configuration.hostname` a little puzzling because that's always the same as just `configuration.hostname`.

Sources/Vapor/HTTP/Server/HTTPServer.swift
bisgardo: Done.

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -331,6 +330,19 @@ public final class HTTPServer: Server, Sendable {
             ).wait()
         }
 
+        /// Override configuration with actual address.
+        /// This may differ from the provided configuation if port 0 was provided, for example.
+        if let localAddress = self.localAddress {
+            if let hostname = localAddress.hostname, let port = localAddress.port {
```
bisgardo, line 336: `.hostname` and `.port` are always (non) `nil` together.

Is it relevant to do the same for `pathname` (for Unix sockets)?

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -331,6 +330,19 @@ public final class HTTPServer: Server, Sendable {
...
+                configuration.port = port
+                configuration.hostname = hostname
+                addressDescription = "\(scheme)://\(hostname):\(port)"
```
bisgardo, line 339: Don't know about this? We could also make `addressDescription` a method on `Configuration`.

0xTim: I suggest we just have a new variable to avoid the mutating variable that might get confusing later on

Sources/Vapor/HTTP/Server/HTTPServer.swift
0xTim: We definitely want to print the socket address at the notice log level once the server has started

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -307,19 +307,18 @@ public final class HTTPServer: Server, Sendable {
             /// Override the socket path.
             configuration.address = address!
         }
-        
-        /// Print starting message.
+
+        /// Log starting message for debugging before attempting to start the server.
         let scheme = configuration.tlsConfiguration == nil ? "http" : "https"
-        let addressDescription: String
+        var addressDescription: String
```
0xTim, line 313: Let's have a separate variable for pre and post app startup

0xTim (changes requested):
Thanks! Just a couple of nits left

Sources/Vapor/HTTP/Server/HTTPServer.swift
bisgardo: That also happened with this. The question is whether it's relevant to patch unix socket addres with the actual pathname (i.e. whether it can differ from the configured one). I added that in the latest commit, but can remove it again if it's irrelevant.

Sources/Vapor/HTTP/Server/HTTPServer.swift
bisgardo: I hope you agree that a computed property on Configuration is cleaner than duplicating the code block, so I did that. Is that ok?

bisgardo:
Ping @0xTim :)

bisgardo:
@0xTim Is there anything that needs to change before this can get merged? If you disagree with extracting `addressDescription` I can inline it, no problem.

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -48,23 +48,34 @@ public final class HTTPServer: Server, Sendable {
...
+                case .hostname(let hostname, _):
+                    address = .hostname(hostname, port: newValue)
+                default:
+                    address = .hostname(nil, port: newValue)
+                }
+            }
+        }
+        
+        /// A human-readable description of the configured address. Used in log messages when starting server.
+        public var addressDescription: String {
```
0xTim, line 70: Let's make this private, we can always open it up if needed but since we only use it internally it's better to keep it like that for now

0xTim (changes requested):
One minor change then good to go!

Sources/Vapor/HTTP/Server/HTTPServer.swift
bisgardo: Done (assuming you meant non-public and not actually `private` as that doesn't compile).

0xTim (changes requested):
Sorry one last thing to add - can you add a test for the configuration? Essentially run the server set to port 0 and then try and read the port from the configuration and make sure it isn't 0

0xTim (approved):
Thanks!

0xTim:
@bisgardo the new test is failing

bisgardo:
> @bisgardo the new test is failing

It was an existing test, not the new one. I've pushed a fix.

PR #3161, opened 2024-03-15
Add XCTHTTPClient

AFutureD:
This PR allow user to tests Services or Method without perform actual requests.

Here is an example.
```Swift
final class AppleTests: XCTestCase {
    
    var client = XCTHTTPClient(eventLoop: nil)
   
    func testSomething() async throws {
        
        var resp: ClientResponse = .init(status: .ok)
        try resp.content.encode(content, as: HTTPMediaType.json)
        
        client.mock(
            request: .init(method: .POST, url: .init(string: AppleApi.ValidationTokenRequest.method)),
            response: resp
        )
        
        let service = Service(signers: signers, client: client , properties: prop, log: log)
        
        let req = AppleApi.ValidationTokenRequest.init("...")
        let result = try await service.validateToken(request: req)
        
        guard case let .success(received) = result else {
            XCTFail()
            return
        }
        
        XCTAssertEqual(...)
    }
}
```

### DIFF
--- a/Sources/XCTVapor/XCTHTTPClient.swift
+++ b/Sources/XCTVapor/XCTHTTPClient.swift
@@ -1,0 +1,49 @@
+import Foundation
+import Vapor
+import NIOCore
+
+extension ClientResponse: @unchecked Sendable  {
+}
+
+struct XCTHTTPClient: Client {
+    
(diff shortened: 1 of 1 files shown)


AFutureD:
Hi @0xTim, I face the problem that there is no place for XCTHTTPClient  to write test code.
Can you give me some advice?

AFutureD:
close as there is no progress.

PR #3164, opened 2024-04-01
Fix decoding 'flag' URL query params via `.decode(StructType.self)`

### LINKED ISSUE #3163: 'Flag' URL Query params don't decode into structs
### Describe the issue

req.query.decode(Struct.self) fails to decode '?flag1' query into 'let flag1: Bool?' property

### Vapor version

4.92.5

mac OS 14.3.1

### Swift version

Apple Swift version 5.9.2

### Steps to reproduce

1. Add the following to "routes.swift" in a template Vapor project created with `vapor new hello -n`. 

```	
struct QueryStruct: Content {
    var flag1: Bool?
}
app.get("singlevaluedecode") { req async throws -> String in
    var queryStruct = QueryStruct()
    queryStruct.flag1 = req.query[Bool.self, at: "flag1"]
    return "\(queryStruct)"
}
app.get("structdecode") { req async throws -> String in
    let queryStruct = try req.query.decode(QueryStruct.self)
    return "\(queryStruct)"
}
```

2. Run curl on the two routes, with a `?flag1` query:

```
% curl "http://127.0.0.1:8080/singlevaluedecode?flag1" 
QueryStruct(flag1: Optional(true))

% curl "http://127.0.0.1:8080/structdecode?flag1"
QueryStruct(flag1: nil)
```

### Outcome

It's not feasible for a generalized decoder to capture all the possible semantic meanings of a query; a query that relied on order of query params couldn't use either single-value decoding nor struct decoding. There's no standard here; the query coders are a utility to help with commonly-used cases.

But single-value and struct decoding should work the same, as much as they can. 4.92.2 fixed flag query params for the single-value decode method; struct decoding (into optional Bools) ought to have parity in this case.

### Additional notes

Cause of this issue appears to be that `URLEncodedFormParser` parses flags into `URLEncodedFormData.values` and key-value params into `URLEncodedFormData.children` while `URLEncodedFormDecoder`'s `_Decoder.KeyedContainer.contains()` only checks `.children`. This makes `decodeIfPresent()` always return nil for flags.

`_Decoder.KeyedContainer.decode()` contains a case for decoding flag values into Bools, so decoding into a non-optional Bool works.

PR incoming.

challfry:
Fixes #3163.

The code:
```
struct QueryStruct: Content {
    var flag1: Bool?
}
let queryStruct = try req.query.decode(QueryStruct.self)
```

produces `queryStruct.flag1 == true` when decoding the URL query "?flag1", matching the behavior of `req.query[Bool.self, at: "flag1"]`.


0xTim (approved):
Thanks!

@gwynne any thoughts?

0xTim:
@challfry could you fix the conflicts?

gwynne (approved):
I'm okay with it, once the conflicts are fixed

challfry:
Merge conflicts fixed.

For the failing vapor/jwt test, it appears jwt-kit changed part of their API last week, and jwt's JWTTests.swift is yet to be updated.

PR #3167, opened 2024-04-04
Migrate to Async NIOFileIO APIs

0xTim:
This migrates `collectFile(at:)` and `writeFile(_:at:)` to use NIO's async NIOFileIO APIs introduced in https://github.com/apple/swift-nio/releases/tag/2.63.0

Also adds a new API for streaming files using a `AsyncSequence` based on the new `NIOFileSystem`.

This work is required to move the `DotEnv` support over to an async API to avoid calling `wait()`s in an async context which can cause issues


Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -279,6 +293,99 @@ public struct FileIO: Sendable {
...
+    ///
+    ///     let data = try await req.fileio.collectFile(file: "/path/to/file.txt")
+    ///     print(data) // file data
+    ///
+    /// - parameters:
+    ///     - path: Path to file on the disk.
+    /// - returns: `ByteBuffer` containing the file data.
+    public func collectFile(at path: String) async throws -> ByteBuffer {
+        guard
+            let attributes = try? FileManager.default.attributesOfItem(atPath: path),
```
ptoffy, line 308: While IMO this should use `NIOFileSystem` rather than `FileManager` I get that it probably makes sense to wait for it to be more stable

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -279,6 +293,99 @@ public struct FileIO: Sendable {
...
+        }
+
+        public func makeAsyncIterator() -> FileChunksIterator {
+            FileChunksIterator(wrapping: fileChunks.makeAsyncIterator(), fileHandle: fileHandle)
+        }
+    }
+
+    /// Reads the contents of a file at the supplied path in chunks.
+    ///
+    ///    for chunk in try await req.fileio.readFile(at: "/path/to/file.txt") {
```
ptoffy, line 353: Oops
```suggestion
    ///    for try await chunk in try await req.fileio.readFile(at: "/path/to/file.txt") {
```

ptoffy (approved):
Nothing blocking

PR #3168, opened 2024-04-04
Mark all functions that use `wait` as `noasync`

0xTim:
⚠️ **WARNING**: If you have strict concurrency checking enabled you should migrate to the async `Application.make()`

NIO's `EventLoopFuture.wait()` is marked as `noasync` because is can cause issues when used in a concurrency context. All places where we call `.wait()` should also be marked as `noasync` to avoid this issue.

This adds `async` alternatives for those functions and adds `noasync` annotations where appropriate.

Also adds an `async` `Application.make` to replace the old initialiser that is now `noasync`


0xTim:
This is reliant on #3167 

Todo:

- [ ] Migrate the commands to `AsyncCommands`
- [ ] Set up tests to ensure we don't add any more back
- [ ] Make NIO the global executor (this may need to be done in the template, still to investigate)

0xTim:
Test failures are expected due to the use of string concurrency

0xTim:
@gwynne I think this is ready for review now, should be non-breaking

gwynne (approved):

PR #3169, opened 2024-04-05
Session cookie creation flag

bgisme:
I ran into a problem where setting the session id prevented it from being saved by SessionsMiddleware.

In line 69 of SessionsMiddleware.swift...
[https://github.com/bgisme/vapor/blob/11cdb29614a5c7f8c5289f3c97b3398c3d89b395/Sources/Vapor/Sessions/SessionsMiddleware.swift#L69](url)
...the session is created depending on whether or not it has an id.

So if you set the session id for the first time it will not be saved.

This pull request is just a simple attempt at addressing the issue with an internal boolean flag for whether the session was created or not. I'll be the first to admit it may not be the best way. Just a suggestion. In case anyone thinks it's a problem.

### DIFF
--- a/Sources/Vapor/Sessions/Session.swift
+++ b/Sources/Vapor/Sessions/Session.swift
@@ -27,12 +27,23 @@ public final class Session: Sendable {
         }
     }
 
+    /// whether created by SessionsMiddleware
+    internal var isCreated: Bool {
+        get {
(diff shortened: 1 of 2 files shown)


0xTim:
I'm going to close this since it's a pretty edge case and you shouldn't be using the Session ID for the state. I believe the underlying issue has been solved on Discord as well

PR #3170, opened 2024-04-09
Add AsyncSequence-based file read method

ptoffy:


Tests/VaporTests/FileTests.swift
```diff
@@ -484,4 +484,23 @@ final class FileTests: XCTestCase {
...
+        defer { app.shutdown() }
+
+        let request = Request(application: app, on: app.eventLoopGroup.next())
+
+        let path = "/" + #filePath.split(separator: "/").dropLast().joined(separator: "/") + "/Utilities/foo.txt"
+
+        let content = try String(contentsOfFile: path)
+
+        var readContent = ""
+        let file = try await request.fileio.readFile(at: path)
```
0xTim, line 499: Just to double check - what's the default chunk size? Are we reading the whole file in one chunk? In which case we should make the chunk size smaller or use a bigger file

ptoffy: I added in a long test file just to be sure

0xTim (approved):

PR #3172, opened 2024-04-12
Use "best for platform" `EventLoopGroup` instead of always using `MultiThreadedEventLoopGroup`

MahdiBM:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,7 +16,7 @@ let package = Package(
     ],
     dependencies: [
         // HTTP client library built on SwiftNIO
-        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.19.0"),
+        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.21.0"),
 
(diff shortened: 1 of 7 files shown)


Package.swift
```diff
@@ -16,7 +16,7 @@ let package = Package(
     ],
     dependencies: [
         // HTTP client library built on SwiftNIO
-        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.19.0"),
+        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.21.0"),
```
MahdiBM, line 19: It is kind of unrelated to the goal of this PR but it should be fine considering the version comes with the `HTTPClient.shared` and is going to be needed by _something_ if not Vapor.

gwynne:
I think we've deliberately avoided doing this in the past for various reasons; NIOTS is really only valuable on iOS.

0xTim:
Yeah we need to do a bit of rearchitecting if we want to support NIOTS properly and frankly that's going to wait until Vapor 5

PR #3175, opened 2024-04-20
Enabled Request Decompression By Default

dimitribouniol:
This enables request decompression by default and sets the decompression limit to 25×.


Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -119,7 +119,7 @@ public final class HTTPServer: Server, Sendable {
 
             /// Enables decompression with default configuration.
             public static var enabled: Self {
-                .enabled(limit: .ratio(10))
+                .enabled(limit: .ratio(25))
             }
```
MahdiBM, line 123: Although this will be a change of behavior for the current users of the variable, i don't think it's for the worse all things considered (the things we talked about on Discord), so should be fine.

MahdiBM (approved):

gwynne (approved):

PR #2998, opened 2023-04-14
Add support for asynchronous body stream writing

Joannis:
- Fixes #2930 - a crash when users try to write a body from within a task towards the ELF APIs.
- Introduces a new API for writing chunked HTTP response bodies
- Adds a helper that automatically manages failing and closing streams

PR #3183, opened 2024-04-30
Fix test failures

0xTim:
Fix parallel test failures introduced in #2998 - not sure how this got through CI ¯\_(ツ)_/¯

### DIFF
--- a/Tests/VaporTests/PipelineTests.swift
+++ b/Tests/VaporTests/PipelineTests.swift
@@ -4,6 +4,7 @@ import XCTest
 import AsyncHTTPClient
 import NIOEmbedded
 import NIOCore
+import NIOConcurrencyHelpers
 
 final class PipelineTests: XCTestCase {
     var app: Application!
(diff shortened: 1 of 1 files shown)


Tests/VaporTests/PipelineTests.swift
```diff
@@ -86,7 +87,9 @@ final class PipelineTests: XCTestCase {
             }))
         }
         
-        try app.start()
+        app.environment.arguments = ["serve"]
```
gwynne, line 90: Why is this needed? Are we mishandling the default command or racing against other parallel tests that chance the commandline arguments?

gwynne (changes requested):
See comment

Tests/VaporTests/PipelineTests.swift
0xTim: It's always been like this - something about running the tests from CI seems to mishandle a default command

PR #3184, opened 2024-04-30
Provide AsyncFileStreaming API

0xTim:
Builds on the work of #2998, #3170 and #3167 to provide a full async streaming API that can be used in Swift Concurrency environments:

* Provides a new `asyncStreamFile(at:chunkSize:mediaType:advancedETagComparison:onCompleted:)` that takes advantage of the full async response streaming
* Fixes a number of bugs with the async Response body streaming
* `FileMiddleware` is now an `AsyncMiddleware`
* Replaces usages of `FileManager` with `NIOFileSystem` apart from in one deprecated API that can't be async
* Correctly marks `XCTVapor` functions as `noasync` where they use `.wait()` and provides proper async alternatives


0xTim:
The stream file functions could use some tidying up as there's a lot of duplicated code there

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
```diff
@@ -59,6 +59,10 @@ final class HTTPServerHandler: ChannelInboundHandler, RemovableChannelHandler {
                 case .failure(let error):
                     if case .stream(let stream) = response.body.storage {
                         stream.callback(ErrorBodyStreamWriter(eventLoop: request.eventLoop, error: error))
+                    } else if case .asyncStream(let stream) = response.body.storage {
+                        Task {
+                            try await stream.callback(ErrorBodyStreamWriter(eventLoop: request.eventLoop, error: error))
```
gwynne, line 64: Probably better to use `try?` here rather than let errors propagate, we're already in an error-handling path as it is. (Yes, I realize it doesn't really matter with this writer, but as a matter of good form.)

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
```diff
@@ -77,10 +81,13 @@ final class HTTPServerHandler: ChannelInboundHandler, RemovableChannelHandler {
     }
 }
 
-struct ErrorBodyStreamWriter: BodyStreamWriter {
+struct ErrorBodyStreamWriter: BodyStreamWriter, AsyncBodyStreamWriter {
     var eventLoop: EventLoop
     var error: Error
```
gwynne, line 86: These don't need to be mutable 🙄

Sources/Vapor/Response/Response+Body.swift
```diff
@@ -242,7 +249,7 @@ extension Response {
 // Since all buffer mutation is done on the event loop, we can be unchecked here.
 // This removes the need for a lock and performance hits from that
 // Any changes to this type need to be carefully considered
-private final class ResponseBodyCollector: BodyStreamWriter, @unchecked Sendable {
+private final class ResponseBodyCollector: BodyStreamWriter, AsyncBodyStreamWriter, @unchecked Sendable {
```
gwynne, line 252: Really, this should look a lot more similar to the `ChannelResponseBodyStream` writer - at least inasmuch as it should maintain enough state to defensively complain on API misuse and so forth. But I don't think it's worth holding up this PR further over it.

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -93,20 +94,21 @@ public struct FileIO: Sendable {
...
+            guard let fileSize = fileSize else {
+                return self.request.eventLoop.makeFailedFuture(Abort(.internalServerError))
+            }
+            return self.read(
+                path: path,
+                fromOffset: 0,
+                byteCount: Int(fileSize),
+                chunkSize: chunkSize,
+                onRead: onRead
+            )
```
gwynne, line 110:
```suggestion
        self.request.eventLoop.makeFutureWithTask {
            guard let fileSize = try await FileSystem.shared.info(forFileAt: .init(path))?.size else {
                throw Abort(.internalServerError)
            }
            try await self.read(
                path: path,
                fromOffset: 0,
                byteCount: Int(fileSize),
                chunkSize: chunkSize,
                onRead: onRead
            ).get()
        }
```
Saves at least one extra thread hop, especially in the case of the file not existing (where it saves more like two or three).

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -531,6 +455,137 @@ public struct FileIO: Sendable {
...
+    public func asyncStreamFile(
+        at path: String,
+        chunkSize: Int = NonBlockingFileIO.defaultChunkSize,
+        mediaType: HTTPMediaType? = nil,
+        advancedETagComparison: Bool = false,
+        onCompleted: @escaping @Sendable (Result<Void, Error>) async throws -> () = { _ in }
+    ) async throws -> Response {
+        // Get file attributes for this file.
+        guard let fileInfo = try await FileSystem.shared.info(forFileAt: .init(path)) else {
+            return Response(status: .internalServerError)
```
gwynne, line 489: Wouldn't we prefer to `throw Abort(.internalServerError)` here to get the benefit of any error middleware?

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -531,6 +455,137 @@ public struct FileIO: Sendable {
...
+            if rangeFromHeaders.unit == .bytes && rangeFromHeaders.ranges.count == 1 {
+                contentRange = rangeFromHeaders
+            } else {
+                contentRange = nil
+            }
+        } else if request.headers.contains(name: .range) {
+            // Range header was supplied but could not be parsed i.e. it was invalid
+            request.logger.debug("Range header was provided in request but was invalid")
+            let response = Response(status: .badRequest)
+            return response
```
gwynne, line 503: Again, shouldn't we throw?

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -531,6 +455,137 @@ public struct FileIO: Sendable {
...
+            response.status = .partialContent
+            response.headers.add(name: .accept, value: contentRange.unit.serialize())
+            if let firstRange = contentRange.ranges.first {
+                do {
+                    let range = try firstRange.asResponseContentRange(limit: Int(fileInfo.size))
+                    response.headers.contentRange = HTTPHeaders.ContentRange(unit: contentRange.unit, range: range)
+                    (offset, byteCount) = try firstRange.asByteBufferBounds(withMaxSize: Int(fileInfo.size), logger: request.logger)
+                } catch {
+                    let response = Response(status: .badRequest)
+                    return response
```
gwynne, line 548: Again, shouldn't we throw?

gwynne (changes requested):
A few nits about behavior

Sources/Vapor/Utilities/FileIO.swift
0xTim: Yeah no idea why I picked that

gwynne (approved):

PR #3186, opened 2024-05-04
Make # of connections accepted per event loop cycle configurable, and raise the default

gwynne:
As per @weissi's suggestion in [this forums post](https://forums.swift.org/t/standard-vapor-website-drops-1-5-of-requests-even-at-concurrency-of-100/71583/49), we raise the default maximum number of connections accepted per cycle of the server's event loop from 4 to 256, and the value is now user-configurable.

There are no new tests for this because I'm not sure if there's a way to measure the effect of changing this value that doesn't involve nondeterministic timing measurements.

Also takes the opportunity/excuse to add the missing `customCertificateVerifyCallback` parameter to the initializers of `HTTPServer.Configuration`.

### DIFF
--- a/Sources/Vapor/HTTP/Server/HTTPServer.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServer.swift
@@ -171,6 +171,14 @@ public final class HTTPServer: Server, Sendable {
         /// This is the same as `NIOSSLCustomVerificationCallback` but just marked as `Sendable`
         @preconcurrency
         public var customCertificateVerifyCallback: (@Sendable ([NIOSSLCertificate], EventLoopPromise<NIOSSLVerificationResult>) -> Void)?
+        
+        /// The number of incoming TCP connections to accept per "tick" (i.e. each time through the server's event loop).
+        ///
+        /// Most users will never need to change this value; its primary use case is to work around benchmarking
+        /// artifacts where bursts of connections are created within extremely small intervals. See
+        /// https://forums.swift.org/t/standard-vapor-website-drops-1-5-of-requests-even-at-concurrency-of-100/71583/49
+        /// for additional information.
+        public var connectionsPerServerTick: UInt
 
         public init(
             hostname: String = Self.defaultHostname,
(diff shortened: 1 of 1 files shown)


Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -388,7 +404,8 @@ private final class HTTPServerConnection: Sendable {
     ) -> EventLoopFuture<HTTPServerConnection> {
         let quiesce = ServerQuiescingHelper(group: eventLoopGroup)
         let bootstrap = ServerBootstrap(group: eventLoopGroup)
-            /// Specify backlog and enable `SO_REUSEADDR` for the server itself.
+            /// Specify accepts per loop and backlog, and enable `SO_REUSEADDR` for the server itself.
+            .serverChannelOption(ChannelOptions.maxMessagesPerRead, value: UInt(configuration.requestsPerServerTick))
```
MahdiBM, line 408: Is the conversion needed?!

```suggestion
            .serverChannelOption(ChannelOptions.maxMessagesPerRead, value: configuration.requestsPerServerTick)
```

MahdiBM (approved):
Looks good to me. Reminded me of TCP congestion control, but that's per-connection so not exactly the same as this.

MahdiBM:
I'd wait for @0xTim too, since he works more with these lower-level stuff.

gwynne:
> I'd wait for @0xTim too, since he works more with these lower-level stuff.

What, you think I'd merge _this_ on _your_ say-so alone? 😂😉

Sources/Vapor/HTTP/Server/HTTPServer.swift
gwynne: Nope, good catch!

MahdiBM:
@gwynne 🫠 😂 of course you would, I know you count every second before you can hit the merge buttons.

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -171,6 +171,14 @@ public final class HTTPServer: Server, Sendable {
...
         @preconcurrency
         public var customCertificateVerifyCallback: (@Sendable ([NIOSSLCertificate], EventLoopPromise<NIOSSLVerificationResult>) -> Void)?
+        
+        /// The number of incoming requests to accept per "tick" (i.e. each time through the server's event loop).
+        ///
+        /// Most users will never need to change this value; its primary use case is to work around benchmarking
+        /// artifacts where bursts of connections are created within extremely small intervals. See
+        /// https://forums.swift.org/t/standard-vapor-website-drops-1-5-of-requests-even-at-concurrency-of-100/71583/49
+        /// for additional information.
+        public var requestsPerServerTick: UInt
```
weissi, line 181: This is technically the maximum number of TCP connections to accept per tick. So wouldn't write "request" here.

gwynne: Good call, fixed!

0xTim (approved):
LGTM!

PR #3187, opened 2024-05-07
Log source file and line info for errors in ErrorMiddleware when possible

gwynne:
Ever since the last changes to `ErrorMiddleware` (by me, naturally), the error logging fails to correctly report file/line/function information even when the error has that data available. We now correctly pass these along to the logging machinery. The error responses sent to clients are unchanged.

Additional changes:
- Restore recognition of the `DebuggableError` protocol (reason and source location information for such errors are now used again).
- Handle generating error responses slightly more efficiently.
- Include the original error message in the fallback text if encoding an error to JSON fails.
- Improve the correctness of the `reason` messages used for `DecodingError`s.


MahdiBM (approved):
all looks good

PR #3188, opened 2024-05-08
Support compiling against Musl

simonjbeaumont:
Vapor already makes some provision for compiling against Musl in the RFC1123 implementation, where `Glibc` is not assumed and is imported conditionally alongside a conditional import of `Musl`. However, there are a couple of other places where `Glibc` is still assumed when compiling for Linux.

This patch replaces these imports with the same `#if canImport(...)` pattern.


0xTim (approved):
Thanks!

0xTim:
Waiting on https://github.com/vapor/console-kit/pull/198 to discuss if we can test this in CI

PR #3189, opened 2024-05-10
Add async alternative for Application.shutdown

0xTim:
Adds an async alternative for `Application.shutdown()` and annotates `shutdown()` with `noasync`

### DIFF
--- a/Sources/Development/entrypoint.swift
+++ b/Sources/Development/entrypoint.swift
@@ -14,6 +14,7 @@ struct Entrypoint {
             try await app.asyncShutdown()
         } catch {
             try await app.asyncShutdown()
+            throw error
         }
     }
 }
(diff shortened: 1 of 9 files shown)


gwynne (approved):
Need to update the template with these new APIs.

Sources/Development/entrypoint.swift
```diff
@@ -7,11 +7,15 @@ struct Entrypoint {
...
-
-        try configure(app)
-        try await app.execute()
+        let app = try await Application.make(env)
+        do {
+            try configure(app)
+            try await app.execute()
+            try await app.asyncShutdown()
+        } catch {
+            try await app.asyncShutdown()
```
gwynne, line 16:
```suggestion
            try? await app.asyncShutdown()
```

0xTim:
Yep that's how I discovered this when updating the template

PR #3190, opened 2024-05-10
Async Serve Command

0xTim:
Migrate `ServeCommand` to an `AsyncCommand` to enable proper custom executor support and remove any calls to `wait()`

### DIFF
--- a/Sources/XCTVapor/XCTApplication.swift
+++ b/Sources/XCTVapor/XCTApplication.swift
@@ -90,41 +90,46 @@ extension Application {
         
         func performTest(request: XCTHTTPRequest) async throws -> XCTHTTPResponse {
             try await app.server.start(address: .hostname(self.hostname, port: self.port))
-            #warning("Fix")
-            defer { app.server.shutdown() }
-            
             let client = HTTPClient(eventLoopGroup: MultiThreadedEventLoopGroup.singleton)
-            defer { try! client.syncShutdown() }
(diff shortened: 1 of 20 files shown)


0xTim:
Need to work out why the configuration has broken in the tests

0xTim:
Verified this works with https://github.com/vapor/template/pull/131/files with Leaf and Fluent and all working

Sources/XCTVapor/XCTApplication.swift
```diff
@@ -89,41 +89,47 @@ extension Application {
...
+                if let query = request.url.query {
+                    url += "?\(query)"
+                }
+                var clientRequest = try HTTPClient.Request(
+                    url: url,
+                    method: request.method,
+                    headers: request.headers
+                )
+                clientRequest.body = .byteBuffer(request.body)
+                let response = try await client.execute(request: clientRequest).get()
```
gwynne, line 120: Why is `.get()` needed here? `HTTPClient` has an async version of `execute()` (with different argument labels).

gwynne (changes requested):
One nit

Sources/XCTVapor/XCTApplication.swift
0xTim: Done

0xTim:
@gwynne this is ready for you now I think

gwynne (approved):

PR #3193, opened 2024-05-14
Add Async Lifecycle Handlers

0xTim:
Adds new protocol functions to `LifecycleHandler`s to support async contexts. This is important because packages like [Redis](https://github.com/vapor/redis) use this to know when to shutdown their connection pool. In the shutdown function, these call `.wait()` which can cause application crashes if called when trying to use NIO's event loop concurrency executor.

This provides async alternatives to allow packages to provide full async calls through their stack to avoid these crashes

### DIFF
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -244,6 +244,12 @@ public final class Application: Sendable {
 
     @available(*, noasync, message: "This can block the thread and should not be called in an async context", renamed: "asyncShutdown()")
     public func shutdown() {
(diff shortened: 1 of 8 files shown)


gwynne (approved):
I absolutely hate this, but I don't have a better solution and it definitely needs addressing, so here we are 🤦‍♀️.

0xTim:
> I absolutely hate this, but I don't have a better solution and it definitely needs addressing, so here we are 🤦‍♀️.

Yep been my feelings for a lot of the async work recently! 😅

PR #3195, opened 2024-05-19
asyncBoot will no longer try booting server again if it is already booted

RussBaz:
The synchronous `boot` function skips running the lifecycle handlers if the server is already booted. However, the async version ignored this check. I have added a small fix to add this check again.

### DIFF
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -247,16 +247,14 @@ public final class Application: Sendable {
     
     /// Called when the applications starts up, will trigger the lifecycle handlers. The asynchronous version of ``boot()``
     public func asyncBoot() async throws {
-        let notBooted = self.isBooted.withLockedValue { booted in
-            guard booted else {
-                booted = true
-                return true
-            }
-
-            return false
+        /// Skip the boot process if already booted
+        guard !self.isBooted.withLockedValue({
+            var result = true
+            swap(&$0, &result)
+            return result
+        }) else {
+            return
         }
-
-        guard notBooted else { return }
 
         for handler in self.lifecycle.handlers {
             try await handler.willBootAsync(self)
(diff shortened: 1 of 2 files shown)


Sources/Vapor/Application.swift
```diff
@@ -247,12 +247,17 @@ public final class Application: Sendable {
...
+                booted = true
+                return true
             }
-            booted = true
+
+            return false
         }
+
+        guard notBooted else { return }
+
```
gwynne, line 260:
```suggestion
        guard !self.isBooted.withLockedValue({
            var result = true
            swap(&$0, &result)
            return result
        }) else {
            return
        }

Comment (RussBaz) [reply Sources/Vapor/Application.swift]: I agree, it made the code more concise and efficient. I merged the suggestion. Not only that, but I even learned that Swift has an inbuilt `swap` function. 😄  Although, I think it is now less obvious what this snippet does. I have added a comment to help with that a bit.

**Review by 0xTim — CHANGES_REQUESTED** [review]
Thanks! Could you add a test to ensure this works?

```swift
    func testBootDoesNotTriggerHandlerMultipleTimes() throws {
        defer { app.shutdown() }
        
        final class Handler: LifecycleHandler, Sendable {
            let bootCount = NIOLockedValueBox(0)
            func willBoot(_ application: Application) throws {
                bootCount.withLockedValue { $0 += 1 }
            }
        }
        
        let handler = Handler()
        app.lifecycle.use(handler)
        
        try app.boot()
        try app.boot()

        XCTAssertEqual(handler.bootCount.withLockedValue({ $0 }), 1)
    }
    
    func testAsyncBootDoesNotTriggerHandlerMultipleTimes() async throws {
        let app = try await Application.make(.testing)
        
        final class Handler: LifecycleHandler, Sendable {
            let bootCount = NIOLockedValueBox(0)
            func willBoot(_ application: Application) throws {
                bootCount.withLockedValue { $0 += 1 }
            }
        }
        
        let handler = Handler()
        app.lifecycle.use(handler)
        
        try await app.asyncBoot()
        try await app.asyncBoot()

        XCTAssertEqual(handler.bootCount.withLockedValue({ $0 }), 1)
        
        try await app.asyncShutdown()
    }
```

In ApplicationTests should work

RussBaz:
Done. Please have a look again.

0xTim (approved):
Thanks!

PR #3196, opened 2024-05-21
Add Async Storage shutdown

0xTim:
Currently running

```swift
Task {
    try? await Task.sleep(for: .seconds(5))
    app.running?.stop()
}
```

When you try and install NIO as the global executor will crash because the storage API didn't have any async entry points so stopping would trigger a synchronous shutdown, with a `wait()`. This fixes that

### DIFF
--- a/Sources/Vapor/Commands/ServeCommand.swift
+++ b/Sources/Vapor/Commands/ServeCommand.swift
@@ -103,12 +103,25 @@ public final class ServeCommand: AsyncCommand, Sendable {
         self.box.withLockedValue { $0 = box }
     }
 
+    @available(*, noasync, message: "Use the async asyncShutdown() method instead.")
     func shutdown() {
         var box = self.box.withLockedValue { $0 }
         box.didShutdown = true
         box.running?.stop()
         if let server = box.server {
             server.shutdown()
+        }
+        box.signalSources.forEach { $0.cancel() } // clear refs
+        box.signalSources = []
+        self.box.withLockedValue { $0 = box }
+    }
+    
+    func asyncShutdown() async {
+        var box = self.box.withLockedValue { $0 }
+        box.didShutdown = true
+        box.running?.stop()
+        if let server = box.server {
+            await server.shutdown()
         }
         box.signalSources.forEach { $0.cancel() } // clear refs
         box.signalSources = []
--- a/Sources/Vapor/Commands/ServeCommand.swift
+++ b/Sources/Vapor/Commands/ServeCommand.swift
@@ -120,9 +120,7 @@ public final class ServeCommand: AsyncCommand, Sendable {
         var box = self.box.withLockedValue { $0 }
         box.didShutdown = true
         box.running?.stop()
-        if let server = box.server {
-            await server.shutdown()
-        }
+        await box.server?.shutdown()
         box.signalSources.forEach { $0.cancel() } // clear refs
         box.signalSources = []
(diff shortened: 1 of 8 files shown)


Sources/Vapor/Commands/ServeCommand.swift
```diff
@@ -115,6 +116,18 @@ public final class ServeCommand: AsyncCommand, Sendable {
         self.box.withLockedValue { $0 = box }
     }
     
+    func asyncShutdown() async {
+        var box = self.box.withLockedValue { $0 }
+        box.didShutdown = true
+        box.running?.stop()
+        if let server = box.server {
+            await server.shutdown()
+        }
```
gwynne, line 125:
```suggestion
        await box.server?.shutdown()
```

Sources/Vapor/Utilities/Storage.swift
```diff
@@ -11,13 +11,21 @@ public struct Storage: Sendable {
...
                 logger.warning("Could not shutdown \(T.self): \(error)")
             }
         }
+        func asyncShutdown(logger: Logger) async {
+            do {
+                try await self.onAsyncShutdown?(self.value)
+            } catch {
+                logger.warning("Could not shutdown \(T.self): \(error)")
+            }
+        }
```
gwynne, line 28: I don't have a better suggestion, but this seems incredibly error-prone/easy to misuse 😕

Sources/Vapor/Utilities/Storage.swift
```diff
@@ -94,20 +102,49 @@ public struct Storage: Sendable {
...
+    ) async
+        where Key: StorageKey
+    {
+        let key = ObjectIdentifier(Key.self)
+        if let value = value {
+            self.storage[key] = Value(value: value, onShutdown: nil, onAsyncShutdown: onShutdown)
+        } else if let existing = self.storage[key] {
+            self.storage[key] = nil
+            await existing.asyncShutdown(logger: self.logger)
+        }
```
gwynne, line 122: What if the existing had a sync shutdown handler but no async one?

gwynne (approved):
A nit and a question, but nothing blocking. I find this _incredibly_ ugly and kinda concerningly fragile, but I have no better suggestions 😕

Sources/Vapor/HTTP/Client/Application+HTTP+Client.swift
```diff
@@ -34,6 +35,28 @@ extension Application.HTTP {
                 return new
             }
         }
+        
+        public var asyncShared: HTTPClient {
+            get async {
+                let lock = self.application.locks.lock(for: Key.self)
+                lock.lock()
```
0xTim, line 42: This makes me uneasy but I don't see a way around it - any thoughts?

Sources/Vapor/Utilities/Storage.swift
0xTim: This is handled in `asyncShutdown()` - it will run both

0xTim:
Ok @gwynne I've tested this with client calls as well and think I've fixed all the different areas without breaking the API too much

Sources/Vapor/Utilities/Storage.swift
0xTim: Ok now it will run the sync one if an async one doesn't exist

gwynne (approved):

PR #3197, opened 2024-05-25
Easy way to customize `Error` rendering

m1guelpf:
Customizing the response an exception returns is currently pretty involved (you need to nuke all middleware, then re-implement `ErrorMiddleware` from scratch.

This PR adds a `RenderableError` protocol, allowing developers to easily customize the response their errors produce.

### DIFF
--- a/Sources/Vapor/Error/RenderableError.swift
+++ b/Sources/Vapor/Error/RenderableError.swift
@@ -1,3 +1,3 @@
 protocol RenderableError: Error {
-    func render(_ req: Request) async -> Response
+    func render(_ req: Request) -> Response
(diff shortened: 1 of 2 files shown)


m1guelpf:
I'd love to allow the `render` function to be `async`, but not 100% how to approach this. would love some thoughts 😁

Naveen-C-Ramachandrappa (approved):

0xTim:
Thanks for the input! Given the ability to customise the `ErrorMiddleware` and the upcoming development of Vapor 5, this isn't something we'll probably take into Vapor 4

PR #3198, opened 2024-05-26
Implement Swift-Periphery

1rlan:
A dead-code analysis tool was added and unused entities were removed


0xTim:
Yeah we're not accepting this. The tool is not clever enough to pick up protocol conformances, default arguments that are used by the compiler and just seemingly deletes deprecated code? This would cause so many breakages in our API it's almost amusing

1rlan:
> Yeah we're not accepting this. The tool is not clever enough to pick up protocol conformances, default arguments that are used by the compiler and just seemingly deletes deprecated code? This would cause so many breakages in our API it's almost amusing

Hi, thanks for the feedback. 
To be frank, this pull request is part of an assignment at my university, the goal is to make some changes to an open-source project that can (potentially) improve the architecture of the project or other smaller aspects of it. 
Thank you very much for your attention to my code!

PR #3201, opened 2024-05-28
Exclude Query and Fragment from URI semicolon fix on Linux

fred-sch:
On Linux, `URLComponents` does not have 100% the same behavior like on macOS. Vapor accounts for [this unfixed bug](https://github.com/apple/swift-corelibs-foundation/issues/3352) by replacing percent-encoded semicolon `%3B` with `;` in `URI`s.

This is however not fully correct, because if a URI contains a percent encoded semicolon, this might have a different meaning, than when it is not percent encoded, compare the following sentence from RFC 3986:
```
A percent-encoding mechanism is used to represent a data octet in a
component when that octet's corresponding character is outside the
allowed set or is being used as a delimiter of, or within, the component.
```

This PR aims to limit the impact of the required semicolon fix by ensuring that query and fragments are not unnecessarily and incorrectly modified. 

Hopefully, in a future with the new swift-foundation this fix will not be needed anymore. But for now it would solve an issue on our side which is related to the concept of a signed request.

### DIFF
--- a/Sources/Vapor/Utilities/URI.swift
+++ b/Sources/Vapor/Utilities/URI.swift
@@ -197,9 +197,12 @@ public struct URI: CustomStringConvertible, ExpressibleByStringInterpolation, Ha
     public var string: String {
         if urlPathAllowedIsBroken {
             // On Linux and in older Xcode versions, URLComponents incorrectly treats `;` as *not* allowed in the path component.
-            let modify = self.components?.string ?? ""
-            let replaceUntil = modify.firstIndex(where: { $0 == "?" || $0 == "#" }) ?? modify.endIndex
-            return modify[modify.startIndex..<replaceUntil].replacingOccurrences(of: "%3B", with: ";") + modify[replaceUntil..<modify.endIndex]
+            let string = self.components?.string ?? ""
+            return string.replacingOccurrences(
+                of: "%3B", with: ";",
+                options: .literal, // N.B.: `rangeOfPath` never actually returns `nil`
+                range: self.components?.rangeOfPath ?? (string.startIndex..<string.startIndex)
+            )
         } else {
             return self.components?.string ?? ""
         }
@@ -335,6 +338,6 @@ extension CharacterSet {
 }
 
 /// On Linux and in older Xcode versions, URLComponents incorrectly treats `;` as *not* allowed in the path component.
-fileprivate let urlPathAllowedIsBroken: Bool = {
-    return CharacterSet.urlPathAllowed != CharacterSet.urlCorrectPathAllowed
+private let urlPathAllowedIsBroken: Bool = {
+    CharacterSet.urlPathAllowed != CharacterSet.urlCorrectPathAllowed
 }()
--- a/Sources/Vapor/Utilities/URI.swift
+++ b/Sources/Vapor/Utilities/URI.swift
@@ -195,14 +195,14 @@ public struct URI: CustomStringConvertible, ExpressibleByStringInterpolation, Ha
     }
 
     public var string: String {
-        #if canImport(Darwin)
-        self.components?.string ?? ""
-        #else
-        // On Linux, URLComponents incorrectly treats `;` as *not* allowed in the path component.
-        let modify = self.components?.string ?? ""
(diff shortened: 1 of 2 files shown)


fred-sch:
The failing test indicates, that the faulty `URLComponents` behavior also affects macOS-13. Unfortunately, I cannot really reproduce that locally.

Sources/Vapor/Utilities/URI.swift
```diff
@@ -195,12 +195,14 @@ public struct URI: CustomStringConvertible, ExpressibleByStringInterpolation, Ha
...
-        self.components?.string ?? ""
-        #else
-        // On Linux, URLComponents incorrectly treats `;` as *not* allowed in the path component.
-        self.components?.string?.replacingOccurrences(of: "%3B", with: ";") ?? ""
-        #endif
+        if urlPathAllowedIsBroken {
+            // On Linux and in older Xcode versions, URLComponents incorrectly treats `;` as *not* allowed in the path component.
+            let modify = self.components?.string ?? ""
+            let replaceUntil = modify.firstIndex(where: { $0 == "?" || $0 == "#" }) ?? modify.endIndex
+            return modify[modify.startIndex..<replaceUntil].replacingOccurrences(of: "%3B", with: ";") + modify[replaceUntil..<modify.endIndex]
```
gwynne, line 202:
```suggestion
            let string = self.components?.string ?? ""
            return string.replacingOccurrences(
                of: "%3B", with: ";",
                options: .literal, // N.B.: `rangeOfPath` never actually returns `nil`
                range: self.components?.rangeOfPath ?? (string.startIndex..<string.startIndex)
            )
```

Sources/Vapor/Utilities/URI.swift
```diff
@@ -331,3 +333,8 @@ extension CharacterSet {
         .urlPathAllowed.union(.init(charactersIn: ";"))
     }
 }
+
+/// On Linux and in older Xcode versions, URLComponents incorrectly treats `;` as *not* allowed in the path component.
+fileprivate let urlPathAllowedIsBroken: Bool = {
+    return CharacterSet.urlPathAllowed != CharacterSet.urlCorrectPathAllowed
+}()
```
gwynne, line 340:
```suggestion
private let urlPathAllowedIsBroken: Bool = {
    CharacterSet.urlPathAllowed != CharacterSet.urlCorrectPathAllowed
}()
```

Tests/VaporTests/URITests.swift
```diff
@@ -132,6 +132,19 @@ final class URITests: XCTestCase {
...
+        XCTAssertURIString(
+            "https://user:pass@vapor.codes:1234/foo;?bar=abcd%C3%A4%2B#qux",
+            hasScheme: "https",
+            hasUserinfo: "user:pass",
+            hasHost: "vapor.codes",
+            hasPort: 1234,
+            hasPath: "/foo;",
+            hasQuery: "bar=abcd%C3%A4%2B",
+            hasFragment: "qux"
+        )
```
gwynne, line 145:
```suggestion
    func testSemicolonIsNotEncodedInPathComponent() {
        XCTAssertURIString(
            "https://user:pass@vapor.codes:1234/foo;?bar=abcd%3B%C3%A4%2B#qux%3B",
            hasScheme: "https",
            hasUserinfo: "user:pass",
            hasHost: "vapor.codes",
            hasPort: 1234,
            hasPath: "/foo;",
            hasQuery: "bar=abcd%3B%C3%A4%2B",
            hasFragment: "qux%3B"
        )
```
^ so that we test that we're _specifically_ only mutating the path component.

gwynne (changes requested):
Good catch on the issue, some nits regarding the fix

Sources/Vapor/Utilities/URI.swift
fred-sch: I totally agree that this is the better solution!

Tests/VaporTests/URITests.swift
fred-sch: I also slightly adapted the query so that we test that we can have `;` and `%3B` at the same time:

```
hasQuery: "bar=abcd%3B%C3%A4%2B;efg",
```

gwynne (approved):
Looks great, thanks!

PR #3208, opened 2024-06-19
Add Sendable conformance to XCTApplicationTester, XCTHTTPRequest/Response, and some others

gwynne:
This prevents compiler errors in the Swift 6 language mode for test methods isolated to global actors (especially `@MainActor`). Also adds `Sendable` conformance to `XCTHTTPRequest` and `XCTHTTPResponse`, for much the same reason.

Also fixes lots and lots of various `Sendable` warnings in Vapor in general, including adding `Sendable` conformance to `ContentConfiguration`, `ContentEncoder`, `ContentDecoder`, `URLQueryEncoder`, `URLQueryDecoder`, `URLEncodedFormEncoder`, and `URLEncodedFormDecoder`.


0xTim (approved):

Sources/XCTVapor/XCTHTTPRequest.swift
```diff
@@ -1,18 +1,39 @@
...
 import Vapor
 
-public struct XCTHTTPRequest {
-    public var method: HTTPMethod
-    public var url: URI
-    public var headers: HTTPHeaders
-    public var body: ByteBuffer
-    
+public struct XCTHTTPRequest: Sendable {
+    private struct Box: Sendable {
```
0xTim, line 7: Why do we need a box? This is a struct

Sources/XCTVapor/XCTHTTPResponse.swift
```diff
@@ -1,12 +1,35 @@
...
 import NIOHTTP1
+import NIOConcurrencyHelpers
 import Vapor
 import XCTest
 
-public struct XCTHTTPResponse {
-    public var status: HTTPStatus
-    public var headers: HTTPHeaders
-    public var body: ByteBuffer
+public struct XCTHTTPResponse: Sendable {
```
0xTim, line 7: As above

Sources/Vapor/Utilities/RFC1123.swift
```diff
@@ -16,18 +16,9 @@ import NIOConcurrencyHelpers
...
     /// A static RFC1123 helper instance
     static var shared: RFC1123 {
-        if let existing = thread.currentValue {
-            return existing
-        } else {
-            let new = RFC1123()
-            thread.currentValue = new
-            return new
-        }
+        .init()
```
0xTim, line 21: Doesn't `RFC1123` need to be `Sendable`? This is my one concern with this is this is used for the `Response` so could introduce a slow down but I don't think we have much choice with a/a

Sources/Vapor/Content/PlaintextEncoder.swift
```diff
@@ -49,7 +49,7 @@ public struct PlaintextEncoder: ContentEncoder {
 
 // MARK: Private
 
-private final class _PlaintextEncoder: Encoder, SingleValueEncodingContainer {
+private final class _PlaintextEncoder: Encoder, SingleValueEncodingContainer, @unchecked Sendable {
     public var codingPath: [CodingKey] = []
```
0xTim, line 53: This is confusing - the properties shouldn't be public since it's a private type. Could we make these `(private) set` to avoid any potential data races?

0xTim (changes requested):
Need to update the title and this should probably be a semver minor now as well

Sources/XCTVapor/XCTHTTPRequest.swift
gwynne: Yep, good call, I'd been doing a bunch of boxes for classes in another project when I hit the need for these to be Sendable so I did it by reflex 🤦‍♀️

Sources/Vapor/Utilities/RFC1123.swift
gwynne: The compiler didn't complain about it not being `Sendable` 🤷‍♀️. I marked it as such anyway, though.

Sources/Vapor/Content/PlaintextEncoder.swift
gwynne: I think the `public` access levels on those declarations are holdovers from when the code got restructured a bit. Turns out there's a good bit going on here that doesn't need to be anyhow, I've cleaned it up.

0xTim (approved):

PR #3211, opened 2024-07-05
fix: support capital letters in the domain part of email validation

### LINKED ISSUE #2889: Fix handling of uppercase domain names when validating email addresses
`Validator.email` misbehaves when asked to validate an email address whose domain name contains uppercase characters. The address will be reported as invalid, which is not correct by spec (lowercase is the recommended canonical form, but isn't an actual requirement). See https://github.com/vapor/vapor/blob/main/Sources/Vapor/Validation/Validators/Email.swift#L56

The correct fix is for the regular expression to either allow uppercase, or, preferably, be replaced with something that's not guaranteed to have other issues 🙂.

Austinpayne:
Fixes #2889


0xTim (approved):
Thanks!

PR #3214, opened 2024-07-10
Incorrect HTTPMediaType/parameters Documentation Formatting.

dimitribouniol:
Fixed an issue where docs for HTTPMediaType/parameters were incorrectly formatted. This was because `parameters` was being interpreted as the argument list, and wouldn't render.

### Before

image

### After

image


0xTim (approved):
Thanks!

PR #3215, opened 2024-07-10
Conditional Content Response Compression

dimitribouniol:
Added support for conditionally compressing responses based on content type or marker headers. This is necessary because some resource types, such as images, don't compress much, and end up maxing out CPU time on the thread and sometimes block the entire channel while they compress. This results in pipelined resources taking a long time to load even on fast connections.

This change comes with three knobs to control this:
- A global disabled/enabled by default state for the entire server.
- Allow and disallow lists of content types for automatic configurations.
- A marker header for explicitly configuring certain requests or routes to compress or not.

Notably, since the response compression handler now takes a predicate, it is always installed, and the predicate statelessly determines if a response should be compressed based on the server configuration:
- If a response compression configuration was never set, this continues to be a no-op and no compression will occur, though routes can now explicitly enable it as needed.
- If response compression is enabled, compression will be supported for all types except those that are known to be incompressible, fixing poor performance when there is no gain to be had by compressing resources.
- If any of the new permutations are used, the user gains more control over the process as needed.

This change relies on a feature recently added to [NIOExtras](https://github.com/apple/swift-nio-extras): https://github.com/apple/swift-nio-extras/pull/225

### DIFF
--- a/Sources/Vapor/HTTP/Headers/HTTPMediaType.swift
+++ b/Sources/Vapor/HTTP/Headers/HTTPMediaType.swift
@@ -167,6 +167,57 @@ public struct HTTPMediaType: Hashable, CustomStringConvertible, Equatable, Senda
     }
 }
 
+/// A collection for efficiently determining if a set of types contains another type.
+public struct HTTPMediaTypeSet: Sendable {
+    let mediaTypeLookup: [String : [String : Set<HTTPMediaType>]]
+    let allowsAny: Bool
+    let allowsNone: Bool
+    
+    public init(mediaTypes: some Sequence<HTTPMediaType>) {
+        var mediaTypeLookup: [String : [String : Set<HTTPMediaType>]] = [:]
+        for mediaType in mediaTypes {
+            mediaTypeLookup[mediaType.type, default: [:]][mediaType.subType, default: Set()].insert(mediaType)
+        }
+        self.mediaTypeLookup = mediaTypeLookup
+        self.allowsAny = mediaTypeLookup["*"] != nil
+        self.allowsNone = mediaTypeLookup.isEmpty
+    }
+    
+    /// Check to see if a media type is contained within the set.
+    public func contains(_ mediaType: HTTPMediaType) -> Bool {
+        /// If we allow any type or no types, stop here. These are uncommon cases, so we cache the results ahead of time to avoid the extra dictionary checks.
+        if allowsAny { return true }
+        if allowsNone { return false }
+        
+        /// Make sure we have an entry for the specific type:
+        guard let mediaSubTypeLookup = mediaTypeLookup[mediaType.type]
+        else { return false }
+        
+        /// If we alow any of the subtypes, stop here.
+        if mediaSubTypeLookup["*"] != nil { return true }
+        
+        /// Make sure we have an entry for the specific sub type:
+        guard let mediaTypes = mediaSubTypeLookup[mediaType.subType]
+        else { return false }
+        
+        /// Check and return true if either the type as is (with potential parameters), or the parameter-less type is in the set:
+        return mediaTypes.contains(mediaType) || mediaTypes.contains(HTTPMediaType(type: mediaType.type, subType: mediaType.subType))
+    }
+    
+    /// The super set of all ``HTTPMediaType``s.
+    public static let all = HTTPMediaTypeSet(mediaTypes: [.any])
+    
+    /// The empty set of ``HTTPMediaType``s.
+    public static let none = HTTPMediaTypeSet(mediaTypes: [])
+}
+
+extension HTTPMediaTypeSet: ExpressibleByArrayLiteral {
+    public init(arrayLiteral elements: HTTPMediaType...) {
+        self = HTTPMediaTypeSet(mediaTypes: elements)
+    }
+}
+
+// MARK: - Media Type Dataset
 public extension HTTPMediaType {
     /// Any media type (*/*).
     static let any = HTTPMediaType(type: "*", subType: "*")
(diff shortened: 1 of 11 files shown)


Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -89,33 +89,161 @@ public final class HTTPServer: Server, Sendable {
         /// Response compression configuration.
         public var responseCompression: CompressionConfiguration
 
-        /// Supported HTTP compression options.
+        /// Supported HTTP response compression options.
         public struct CompressionConfiguration: Sendable {
```
dimitribouniol, line 93: Any qualms we me:
a) Moving this whole thing to a `HTTPServerConfiguration+ResponseCompressionConfiguration.swift` file, and
b) Renaming this to `ResponseCompressionConfiguration` (adding a type alias for `CompressionConfiguration`, and deprecating it for backwards compatibility)?

dimitribouniol:
Tested in my project, and noticed a 2-3× load time improvement on the same host when compression is enabled with default settings (182KB JSON compressed to 28KB, 90 or so images left untouched, load times went from 5-6s to 1.2-1.9s).

0xTim:
@dimitribouniol is this ready to go?

dimitribouniol:
@0xTim Yep!

Sources/Vapor/HTTP/Headers/HTTPMediaType.swift
```diff
@@ -777,3 +828,102 @@ let fileExtensionMediaTypeMapping: [String: HTTPMediaType] = [
...
+        HTTPMediaType(type: "application", subType: "x-gtar-compressed"),
+        HTTPMediaType(type: "application", subType: "x-lha"),
+        HTTPMediaType(type: "application", subType: "x-lzh"),
+        HTTPMediaType(type: "application", subType: "x-lzx"),
+        HTTPMediaType(type: "application", subType: "x-tar"),
+        HTTPMediaType(type: "audio", subType: "*"),
+        HTTPMediaType(type: "image", subType: "gif"),
+        HTTPMediaType(type: "image", subType: "ief"),
+        HTTPMediaType(type: "image", subType: "jp2"),
+        HTTPMediaType(type: "image", subType: "jpeg"),
```
0xTim, line 896: I would have thought images would compress nicely

Sources/Vapor/HTTP/Server/HTTPServer.swift
0xTim: Nope I think this is sensible - you can rename `DecompressionConfiguration` to `RequestDecompressionConfiguration` at the same time.

The only thing to check is to make sure we're not using this in the `Client` anywhere for decompressing client responses (and the same with client requests in the inverse)

Sources/Vapor/Middleware/ResponseCompressionMiddleware.swift
```diff
@@ -0,0 +1,64 @@
...
+    /// - SeeAlso: Please see ``responseCompressionOverride`` for more details.
+    public init(override: HTTPHeaders.ResponseCompression, force shouldForce: Bool = false) {
+        self.responseCompressionOverride = override
+        self.shouldForce = shouldForce
+    }
+    
+    public func respond(to request: Request, chainingTo next: any AsyncResponder) async throws -> Response {
+        let response = try await next.respond(to: request)
+        /// Only set the header if it is unset, and prefer the next responder's header over our own override, as _it_ is overriding ours.
+        if response.headers.responseCompression == .unset || shouldForce {
```
0xTim, line 38: Is it worth checking the `Accept` header here as well?

0xTim (commented):
Overall, looks good! Couple of comments/queries

Sources/Vapor/HTTP/Server/HTTPServer.swift
dimitribouniol: Sweet, I'll make this change likely by tomorrow!

Sources/Vapor/HTTP/Headers/HTTPMediaType.swift
dimitribouniol: Unfortunately, they compressed very poorly (oftentimes ending up larger than the originals, though I had ImageOptim-ized images), and consumed a full CPU core (often the _same_ CPU core for all resources on a page, since they could end up using the same connection on http1.1/h2) to do essentially nothing other than deliver resources slowly. It could be worth giving /Public compression on build capabilities for static, pre-compressed resources, but out of scope here 😔

Sources/Vapor/Middleware/ResponseCompressionMiddleware.swift
dimitribouniol: `HTTPResponseCompressor` already checks for this, and always calls our predicate whether it's supported or not, so we can clean up private headers we've attached for this purpose. So either:
- `Accept` is specified, the compressor asks us if we still want compression, we check the headers to enable it or not, and we delete the private header, or
- `Accept` does not match a compression type we support or is missing, the compressor tells us that its not supported, and we delete the private header.

Sources/Vapor/HTTP/Server/HTTPServer.swift
dimitribouniol: @0xTim This change has been made: https://github.com/vapor/vapor/pull/3215/commits/a956312001747cd429675f3b0779557b4446d03c

I think we should be good with semver, but I'll keep an eye on CI to see if I messed up the rename.

0xTim (approved):

Sources/Vapor/HTTP/Server/HTTPServer.swift
```diff
@@ -23,7 +23,7 @@ public final class HTTPServer: Server, Sendable {
         public static let defaultHostname = "127.0.0.1"
         public static let defaultPort = 8080
         
-        /// Address the server will bind to. Configuring an address using a hostname with a nil host or port will use the default hostname or port respectively.
+        /// Address the server will bind to. Configuring an addrexss using a hostname with a nil host or port will use the default hostname or port respectively.
```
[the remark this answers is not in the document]
dimitribouniol: Oops, this must have been the cat waking over the keyboard while I was moving things over, sorry about that 🤣

PR #3224, opened 2024-08-20
Use Configured JSONEncoder in ErrorMiddleware

### LINKED ISSUE #3218: Use global JSONEncoder (via ContentConfiguration) in ErrorMiddleware
I've configured the default JSONEncoder to pretty print output. This works everywhere, except for `ErrorMiddleware` which creates its own `JSONEncoder`.

`ErrorMiddleware` should instead use `ContentConfiguration.global.requireEncoder(for: .json)`

0xTim:
Use the prescribed content encoder for the body of `ErrorMiddleware` instead of a contained `JSONEncoder` that can't be overridden.

You can set the encoder with

```swift
ContentConfiguration.global.use(encoder: myCustomEncoder, for: .json)
```

Resolves #3218

### DIFF
--- a/Sources/Vapor/Middleware/ErrorMiddleware.swift
+++ b/Sources/Vapor/Middleware/ErrorMiddleware.swift
@@ -49,6 +49,7 @@ public final class ErrorMiddleware: Middleware {
                 let encoder = try ContentConfiguration.global.requireEncoder(for: .json)
                 var byteBuffer = req.byteBufferAllocator.buffer(capacity: 0)
                 try encoder.encode(ErrorResponse(error: true, reason: reason), to: &byteBuffer, headers: &headers)
+                
                 body = .init(
                     buffer: byteBuffer,
                     byteBufferAllocator: req.byteBufferAllocator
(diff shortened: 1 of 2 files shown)


Sources/Vapor/Middleware/ErrorMiddleware.swift
```diff
@@ -46,11 +46,13 @@ public final class ErrorMiddleware: Middleware {
             // attempt to serialize the error to json
             let body: Response.Body
             do {
+                let encoder = try ContentConfiguration.global.requireEncoder(for: .json)
```
dimitribouniol, line 49: Would it be more flexible to load an encoder specifically for error types, or allow this type to be configured on the error middleware, so the user can more finely control _which_ encoder gets used here?

dimitribouniol: (I guess the middleware can take any response serialization closure, so maybe not the biggest concern)

0xTim: Yeah that was my thinking - if you want that amount of customisation then you can just switch out the closure or even provide your own middleware. It's reasonable to assume that any JSONEncoder you set for `ContentConfiguration` will be used for encoding the error JSON

dimitribouniol (approved):

PR #3225, opened 2024-08-20
Support Swift 6

0xTim:
* Drop support for Swift 5.7
* Fix warnings in Vapor with latest Swift 6 changes
* Fix tests on Linux nightlies

### DIFF
--- a/Sources/Vapor/Logging/LoggingSystem+Environment.swift
+++ b/Sources/Vapor/Logging/LoggingSystem+Environment.swift
@@ -2,7 +2,8 @@ import Logging
 import ConsoleKit
 
 extension LoggingSystem {
-    public static func bootstrap(from environment: inout Environment, _ factory: (Logger.Level) -> (String) -> LogHandler) throws {
+    @preconcurrency
+    public static func bootstrap(from environment: inout Environment, _ factory: @Sendable (Logger.Level) -> (@Sendable (String) -> LogHandler)) throws {
         let level = try Logger.Level.detect(from: &environment)
 
         // Bootstrap logger with a factory created by the factoryfactory.
--- a/Tests/VaporTests/ServerTests.swift
+++ b/Tests/VaporTests/ServerTests.swift
@@ -349,8 +349,7 @@ final class ServerTests: XCTestCase, @unchecked Sendable {
             )
         }
         
-        try app.server.start()
-        defer { app.server.shutdown() }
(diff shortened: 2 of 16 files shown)


Tests/VaporTests/ServerTests.swift
```diff
@@ -42,6 +46,8 @@ final class ServerTests: XCTestCase, @unchecked Sendable {
         XCTAssertEqual(res.body?.string, "bar")
     }
     
+    // `httpUnixDomainSocket` is currently broken in 6.0
```
dimitribouniol, line 49: Should we make a ticket for this and link it here?

Tests/VaporTests/URITests.swift
```diff
@@ -183,13 +183,16 @@ final class URITests: XCTestCase {
         XCTAssertEqual(uri.string, "foobar://host:1/test?query#fragment")
     }
     
+    // Disable tests on Linux 6.0 until behaviour is fixed
+    #if compiler(<6.0)
```
dimitribouniol, line 187: Same here

dimitribouniol (approved):
🚀

Tests/VaporTests/ServerTests.swift
0xTim: It's fixed in the 6.0 branch, just waiting on a release

Sources/Vapor/Logging/LoggingSystem+Environment.swift
```diff
@@ -2,7 +2,8 @@ import Logging
 import ConsoleKit
 
 extension LoggingSystem {
-    public static func bootstrap(from environment: inout Environment, _ factory: (Logger.Level) -> (String) -> LogHandler) throws {
+    @preconcurrency
+    public static func bootstrap(from environment: inout Environment, _ factory: @Sendable (Logger.Level) -> (@Sendable (String) -> LogHandler)) throws {
```
dimitribouniol, line 6: Interesting that the API checker is flagging this as different — the parentheses are implied in the same spot on the old version, aren't they?

ptoffy (approved):

Sources/Vapor/Logging/LoggingSystem+Environment.swift
0xTim: Yeah but I don't think the API checker is clever enough to work that out

PR #3228, opened 2024-09-02
Add response order pipelining test

ptoffy:
This adds a test which makes sure that Vapor conforms to the pipelining spec by returning responses in the order they were made rather than the processing order

### DIFF
--- a/Tests/VaporTests/PipelineTests.swift
+++ b/Tests/VaporTests/PipelineTests.swift
@@ -331,7 +331,7 @@ final class PipelineTests: XCTestCase {
 
         try await channel.writeInbound(ByteBuffer(string: "GET /sleep/100 HTTP/1.1\r\n\r\nGET /sleep/0 HTTP/1.1\r\n\r\n"))
 
-        // We expect 6 responses to be there, however if there are less
+        // We expect 6 writes to be there - three parts (the head, body and separator for each request). However, if there are less
         // we need to have a timeout to avoid hanging the test
         let deadline = NIODeadline.now() + .milliseconds(200)
         var responses: [String] = []
--- a/Tests/VaporTests/PipelineTests.swift
+++ b/Tests/VaporTests/PipelineTests.swift
@@ -382,7 +382,8 @@ final class PipelineTests: XCTestCase {
             var sleeps = 0
             for try await chunk in inbound {
                 data.writeImmutableBuffer(chunk)
-                data.writeImmutableBuffer(ByteBuffer(string: "\r\n"))
+                data.writeString("\r\n")
+                
                 if String(decoding: chunk.readableBytesView, as: UTF8.self).components(separatedBy: "\r\n").contains(where: { $0.hasPrefix("slept") }) {
                     sleeps += 1
                 }
--- a/Tests/VaporTests/PipelineTests.swift
+++ b/Tests/VaporTests/PipelineTests.swift
@@ -1,10 +1,11 @@
 @testable import Vapor
-import enum NIOHTTP1.HTTPParserError
+import NIOHTTP1
 import XCTest
 import AsyncHTTPClient
 import NIOEmbedded
 import NIOCore
 import NIOConcurrencyHelpers
+import class NIOPosix.ClientBootstrap
 
 final class PipelineTests: XCTestCase {
     var app: Application!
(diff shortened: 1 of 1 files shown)


Joannis:
This is exactly what I found

Joannis (approved):

0xTim:
@weissi could you take a look and confirm we're not missing anything and the test is correct in showing that pipelining is working as expected?

Tests/VaporTests/PipelineTests.swift
```diff
@@ -311,6 +311,39 @@ final class PipelineTests: XCTestCase {
...
+        let app = self.app!
+        _ = try await (channel.eventLoop as! NIOAsyncTestingEventLoop).executeInContext {
+            channel.pipeline.addVaporHTTP1Handlers(
+                application: app,
+                responder: app.responder,
+                configuration: app.http.server.configuration
+            )
+        }
+
+        try await channel.writeInbound(ByteBuffer(string: "GET /sleep/100 HTTP/1.1\r\n\r\nGET /sleep/50 HTTP/1.1\r\n\r\n"))
```
weissi, line 331: nit: but ideally you want to make the second request as quick as possible. So maybe change the second one to `/sleep/0` or `/sleep/1`, that widens the window during which a reorder could happen.

Tests/VaporTests/PipelineTests.swift
```diff
@@ -311,6 +311,39 @@ final class PipelineTests: XCTestCase {
...
+
+        try await Task.sleep(for: .milliseconds(1000))
```
weissi, line 333: This sleep is racy and shouldn't be necessary.

@franzbusch what's the best way to wait for an outbound message in `NIOAsyncTestingChannel`?

Tests/VaporTests/PipelineTests.swift
```diff
@@ -311,6 +311,39 @@ final class PipelineTests: XCTestCase {
...
+    func testCorrectResponseOrder() async throws {
+        app.get("sleep", ":ms") { req -> String in
+            let ms = try req.parameters.require("ms", as: Int64.self)
+            try await Task.sleep(for: .milliseconds(ms))
+            return "slept \(ms)ms"
+        }
+
+        let channel = NIOAsyncTestingChannel()
+        let app = self.app!
+        _ = try await (channel.eventLoop as! NIOAsyncTestingEventLoop).executeInContext {
```
weissi, line 323: That's a good idea to test this with async testing channel. Additionally I'd recommend also writing an end to end test against actual Vapor.

ptoffy: Yeah I couldn't really get it to work with EmbeddedChannel, not even with EVL based sleeps

weissi: `EmbeddedChannel` won't work I don't think because Vapor may require actual threads.

I meant an actual Vapor, connecting over (localhost TCP) using `ClientBootstrap(...).connect()`

Tests/VaporTests/PipelineTests.swift
ptoffy: I figured but without it the read returns nothing so I figured we'd have to wait for the message to be returned by the endpoint. Maybe using shorter sleeps times as you suggested it might work without it

Tests/VaporTests/PipelineTests.swift
ptoffy: Yeah I meant I tried using EmbeddedChannel instead of the NIOAsyncTestChannel

Tests/VaporTests/PipelineTests.swift
ptoffy: @weissi FYI I'm guessing we were looking for [`waitForOutboundWrite(as:)`](https://swiftpackageindex.com/apple/swift-nio/2.67.0/documentation/nioembedded/nioasynctestingchannel/waitforoutboundwrite(as:))

Tests/VaporTests/PipelineTests.swift
```diff
@@ -311,6 +312,92 @@ final class PipelineTests: XCTestCase {
...
+            channel.pipeline.addVaporHTTP1Handlers(
+                application: app,
+                responder: app.responder,
+                configuration: app.http.server.configuration
+            )
+        }
+
+        try await channel.writeInbound(ByteBuffer(string: "GET /sleep/100 HTTP/1.1\r\n\r\nGET /sleep/0 HTTP/1.1\r\n\r\n"))
+
+        // We expect 6 responses to be there, however if there are less
```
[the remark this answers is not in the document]
ptoffy: Don't ask me why but the channel consistently contains 
```
([String]) 6 values {
  [0] = "HTTP/1.1 200 OK\r\ncontent-type: text/plain; charset=utf-8\r\ncontent-length: 11\r\nconnection: keep-alive\r\ndate: Fri, 06 Sep 2024 16:01:35 GMT\r\n\r\n"
  [1] = "slept 100ms"
  [2] = ""
  [3] = "HTTP/1.1 200 OK\r\ncontent-type: text/plain; charset=utf-8\r\ncontent-length: 9\r\nconnection: keep-alive\r\ndate: Fri, 06 Sep 2024 16:01:35 GMT\r\n\r\n"
  [4] = "slept 0ms"
  [5] = ""
}
```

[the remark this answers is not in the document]
0xTim: Ah ok, head, body and separator for each request 👍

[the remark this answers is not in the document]
0xTim:
```suggestion
        // We expect 6 writes to be there - three parts (the head, body and separator for each request). However, if there are less
```

PR #3232, opened 2024-09-08
Fix `.noSignalReceived` body streaming crash

ptoffy:
Update SwiftNIO's minimum version to fix the `preconditionFailure` being thrown in `AsyncSequenceProducerDelegate`


Tests/VaporTests/AsyncRequestTests.swift
```diff
@@ -195,6 +195,42 @@ final class AsyncRequestTests: XCTestCase {
...
+        guard 
+            let localAddress = app.http.server.shared.localAddress,
+            let ip = localAddress.ipAddress,
+            let port = localAddress.port 
+        else {
+            XCTFail("couldn't get ip/port from \(app.http.server.shared.localAddress.debugDescription)")
+            return
+        }
+
+        let fiftyMB = ByteBuffer(repeating: 0x41, count: 600 * 1024 * 1024)
```
dimitribouniol, line 222: Seems like you changed it to 600MB
```suggestion
        let bigBody = ByteBuffer(repeating: 0x41, count: 600 * 1024 * 1024)
```

ptoffy: Oops, thanks!

mrstegeman:
Is anything preventing this from being merged? We’ve been using this patch for a few weeks with no issues and it has solved our upload crashes.

ptoffy:
@mrstegeman main issue is that while this patch _works_, we're not sure this is the correct approach as there might be something wrong with the state machine itself, and the state it's crashing in shouldn't even be reached (so if the `preconditionFailure` is actually correct to be left in)

lmcd:
There should be really be a note in the documentation to not use these features until resolved.
The handling of this is really making us want to migrate away from Vapor. We're running a whole file transfer infrastructure using Vapor and was very close to getting caught in a production nightmare with this 😳

The code isn't documented either, which makes investigating ourselves and contributing a fix difficult.

Volodymyr-13:
Will this be merged anytime soon? We have a lot of crashed with this.

ptoffy:
We're just waiting on the NIO release including the fix

ptoffy:
@IgorMuzyka where do you see that? Latest release is 2.76.1 https://github.com/apple/swift-nio/releases

IgorMuzyka:
> @IgorMuzyka my mistake, not yet

0xTim (approved):

PR #3233, opened 2024-09-10
Pass the method, url, and user agent from the ErrorMiddleware to the logger in the metadata.

maciejtrybilo:
The ErrorMiddleware reports an error to the logger of the request. Sometimes it might be useful to know the endpoint that failed.

This PR adds the method, url, and user agent of the request in the metadata of the report.


petrpavlik:
+1 to this, it'd allow me to afaik update https://github.com/petrpavlik/swift-sentry to attach this data to logged errors, right now it only logs a request id.

Sources/Vapor/Middleware/ErrorMiddleware.swift
```diff
@@ -41,7 +41,12 @@ public final class ErrorMiddleware: Middleware {
             }
             
             // Report the error
-            req.logger.report(error: error, file: source.file, function: source.function, line: source.line)
+            req.logger.report(error: error,
+                              metadata: ["method" : .string(req.method.string),
+                                         "url" : .string(req.url.string)],
```
petrpavlik, line 46: A nitpick, but the docs ask to prefer not using the `.string(...)` syntax? https://apple.github.io/swift-log/docs/current/Logging/Structs/Logger/MetadataValue.html

maciejtrybilo: @petrpavlik It does, but it doesn't compile. I'm not sure force casting is better?
Screenshot 2024-09-10 at 12 47 41

petrpavlik: Okay, never mind what I said than. I'll just out of curiosity try to look into it later.

ptoffy: You'd need to use 
```swift
"method": "\(req.method.string)"
```
to convert it to a String literal

gwynne:
It's not really breaking as written; the new parameter is defaulted. (Yes, technically that can still be source-breaking, but it's a rather low risk with this particular API.)

Sources/Vapor/Middleware/ErrorMiddleware.swift
maciejtrybilo: Ok, changed! I'm not sure if there is any trick to handle the .array in this way?

maciejtrybilo:
I also think it improves the API. Logger has the concept of metadata already, so it looks like an omission not to have it in `report()`?

0xTim (approved):
Yes I think this makes sense

PR #3239, opened 2024-10-03
Use application's logger in InMemory tester

sidepelican:
Usually, a copy of the application's Logger is passed to the `Request`. However, for the `InMemory` part of `XCTApplicationTester`, it seems the default Logger is used.
This PR fix the behavior so that the application's Logger is used instead of the default one.


sidepelican:
> /Users/runner/work/vapor/vapor/Tests/VaporTests/PipelineTests.swift:340: error: -[VaporTests.PipelineTests testCorrectResponseOrder] : failed - Timed out waiting for responses

🤔

Naveen-C-Ramachandrappa (approved):

0xTim (approved):
Thanks!

PR #3241, opened 2024-10-12
Test MUSL and iOS compilation

0xTim:
Adds a CI step to ensure Vapor compiles with MUSL and iOS

@gwynne should we just add this to vapor/ci behind a flag so any package can opt in?

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -119,6 +119,7 @@ let package = Package(
             dependencies: [
                 .product(name: "NIOTestUtils", package: "swift-nio"),
                 .target(name: "XCTVapor"),
+                "Vapor",
             ],
             resources: [
                 .copy("Utilities/foo.txt"),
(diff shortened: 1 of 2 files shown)


gwynne:
I would say yes, we should

0xTim:
Well it works, I guess XCTest isn't exposed in the static SDK. Might need a custom build flag then to ensure we only build the Vapor product

Package.swift
```diff
@@ -119,6 +119,7 @@ let package = Package(
             dependencies: [
                 .product(name: "NIOTestUtils", package: "swift-nio"),
                 .target(name: "XCTVapor"),
+                "Vapor",
```
MahdiBM, line 122: Is this needed?

MahdiBM (approved):
Looks Good.

Package.swift
0xTim: We should have it (we get around it since XCTVapor exports Vapor) but it's technically relying on it being there instead of being a transitive dependency

PR #3243, opened 2024-10-23
Omit ACAO header instead of empty value

grahamburgsma:
_For context, Vapor currently sends an empty Access-Control-Allow-Origin (ACAO) header when the origin does not match or is set to none._

We recently had a pentest done against our Vapor server and the tester reported the following regarding the empty ACAO header:

> When the header is empty, browsers might reject the request without detailed error messages, making it harder for developers to debug or even realize there is a problem. This lack of transparency can lead to extended periods of vulnerability before the issue is discovered.

Looking at some other sources as well, an empty header doesn't appear to be a valid value and so could result in unexpected behaviour.

https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#the_http_response_headers
https://fetch.spec.whatwg.org/#http-access-control-allow-origin

### DIFF
--- a/Sources/Vapor/Middleware/CORSMiddleware.swift
+++ b/Sources/Vapor/Middleware/CORSMiddleware.swift
@@ -139,7 +139,10 @@ public final class CORSMiddleware: Middleware {
             // Modify response headers based on CORS settings
             let originBasedAccessControlAllowHeader = self.configuration.allowedOrigin.header(forRequest: request)
             response.responseBox.withLockedValue { box in
-                box.headers.replaceOrAdd(name: .accessControlAllowOrigin, value: originBasedAccessControlAllowHeader)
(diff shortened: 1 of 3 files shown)


vzsg:
There's a unit test that's still asserting the previous behavior, and is now failing. https://github.com/vapor/vapor/blob/main/Tests/VaporTests/AsyncMiddlewareTests.swift#L92-L106

grahamburgsma:
> There's a unit test that's still asserting the previous behavior, and is now failing. https://github.com/vapor/vapor/blob/main/Tests/VaporTests/AsyncMiddlewareTests.swift#L92-L106

Oops I missed that one, thanks @vzsg.

0xTim (approved):
I agree looking at the specs this seems wrong, thanks for the fix!

PR #3244, opened 2024-10-23
[DNM] Test updated macOS CI

MahdiBM:


aniruddhaadak80:
Testing updated macOS runners is crucial given the recent silicon architecture shifts. Ensuring the test matrix covers standard arm64 compilation is greatly appreciated.

PR #3249, opened 2024-11-01
fix `HTTPMethod.RAW(value: String)` string representation

### LINKED ISSUE #3248: `HTTPMethod.RAW(value: String)` string representation is wrong
### Describe the issue

See title

### Vapor version

4.106.1

macOS 15.0.1

### Swift version

Swift 5.10 using Swift 6 compiler; Swift Package Manager - Swift 6.0.0-dev

### Steps to reproduce

1. create a raw `HTTPMethod` (like  `HTTPMethod.RAW("TEST")`)
2. register a route using the raw method
3. server responds with "not found" json error

### Expected behavior
The method string is `TEST`

### Actual behavior
The method string is `RAW(value: "TEST")`

### Additional notes

Problematic code is located at: https://github.com/vapor/vapor/blob/4d3bc6ce08b72a14c9879810cf0be455ca98f1fb/Sources/Vapor/HTTP/HTTPMethod%2BString.swift#L5

RandomHashTags:
Should fix #3248

### DIFF
--- a/Sources/Vapor/HTTP/HTTPMethod+String.swift
+++ b/Sources/Vapor/HTTP/HTTPMethod+String.swift
@@ -5,6 +5,7 @@ extension HTTPMethod {
     public var string: String {
         switch self {
         case .GET: return "GET"
+        case .RAW(let value): return value
         default: return "\(self)"
         }
     }
--- a/Sources/Vapor/HTTP/HTTPMethod+String.swift
+++ b/Sources/Vapor/HTTP/HTTPMethod+String.swift
@@ -2,7 +2,7 @@ import NIOHTTP1
 
 extension HTTPMethod {
     /// `String` representation of this `HTTPMethod`.
-    @available(*, deprecated, message: "Use 'rawValue' instead")
+    @available(*, deprecated, renamed: "rawValue")
     public var string: String {
         self.rawValue
     }
--- a/Sources/Vapor/HTTP/HTTPMethod+String.swift
+++ b/Sources/Vapor/HTTP/HTTPMethod+String.swift
@@ -3,10 +3,6 @@ import NIOHTTP1
 extension HTTPMethod {
     /// `String` representation of this `HTTPMethod`.
     public var string: String {
-        switch self {
-        case .GET: return "GET"
-        case .RAW(let value): return value
-        default: return "\(self)"
-        }
+        self.rawValue
     }
 }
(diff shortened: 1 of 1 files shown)


Sources/Vapor/HTTP/HTTPMethod+String.swift
```diff
@@ -5,6 +5,7 @@ extension HTTPMethod {
     public var string: String {
         switch self {
         case .GET: return "GET"
+        case .RAW(let value): return value
         default: return "\(self)"
         }
```
gwynne, line 10: The correct fix to this is to use the `rawValue` accessor provided by NIO (ostensibly, this `string` accessor shouldn't exist at all - it's only here because NIO didn't add `rawValue` until a few months after Tanner added this and he never went back to clean it up - but it's public API in Vapor so we can't just remove it, unfortunately).
```suggestion
        self.rawValue
```

gwynne (changes requested):

RandomHashTags:
Maybe remove this file and replace its usages with the `rawValue` in a future patch?

gwynne:
> Maybe remove this file and replace its usages with the `rawValue` in a future patch (or this one 👀 )?

Unfortunately we can't, at least not in Vapor 4 - it's preexisting public API, removing it would be a semver-major break.

Sources/Vapor/HTTP/HTTPMethod+String.swift
```diff
@@ -3,9 +3,6 @@ import NIOHTTP1
 extension HTTPMethod {
     /// `String` representation of this `HTTPMethod`.
     public var string: String {
-        switch self {
-        case .GET: return "GET"
-        default: return "\(self)"
-        }
+        self.rawValue
```
0xTim, line 6: We should probably add a deprecation to the above as well

RandomHashTags: Added

gwynne (approved):

Sources/Vapor/HTTP/HTTPMethod+String.swift
```diff
@@ -2,10 +2,8 @@ import NIOHTTP1
 
 extension HTTPMethod {
     /// `String` representation of this `HTTPMethod`.
+    @available(*, deprecated, message: "Use 'rawValue' instead")
```
0xTim, line 5:
```suggestion
    @available(*, deprecated, renamed: "rawValue")
```

0xTim (approved):

PR #3251, opened 2024-11-07
Don't Duplicate HTTPMediaTypes

0xTim:
A lot of the `HTTPMediaType`s we use in Vapor are already defined - use them

Doesn't need a release as this is only tidying up internal stuff


gwynne (approved):
Nice cleanup!

Issue #3033, opened 2023-06-28
Adopt Swift Tracing


0xTim:
[Swift Tracing](https://github.com/apple/swift-distributed-tracing) is now 1.0 so we should adopt it, probably in middleware to make is easy to add.

We need to be careful as it relies on TLVs so if we cross ELF boundaries we could lose the context, but I think we're ok if we use an async middleware. Will need to test the whole stack

slashmo:
@0xTim Glad to see this being tracked now 👍 Feel free to ping me once you're starting the implementation, then I could update my HTTP client/server example to use Vapor as well as Hummingbird: https://github.com/slashmo/swift-otel/tree/example/http

NeedleInAJayStack:
@0xTim / @gwynne Now that #3253 is merged, I think we can call this closed! Thanks!

NeedleInAJayStack:
Also, @0xTim / @gwynne do you think we should add tracing usage instructions to https://github.com/vapor/docs ? I'd be happy to submit a draft if you think so!

0xTim:
@NeedleInAJayStack Yep some docs would be great!

NeedleInAJayStack:
@0xTim Cool, I've added a pull request: https://github.com/vapor/docs/pull/1039

PR #3253, opened 2024-11-12
Adds swift-distributed-tracing and TracingMiddleware

NeedleInAJayStack:
This adds support for `swift-distributed-tracing` and adds a `TracingMiddleware` type. This type will create a span for each wrapped request, and apply a standard set of span attributes according to OTEL standards.

For background, see: https://github.com/vapor/vapor/issues/3033


NeedleInAJayStack:
As noted in the [issue](https://github.com/vapor/vapor/issues/3033) the `TracingMiddlware` span unfortunately will not cross EventLoopFuture boundaries. One approach could be to add a `rootSpan` stored property onto `Request` so that downstream spans can manually set their parent (as explained [`here`](https://github.com/apple/swift-distributed-tracing/pull/153/files)). I'm definitely open to ideas though!

@slashmo, you asked to be pinged when an implementation was started. I'd love your thoughts!

Sources/Vapor/Middleware/TracingMiddleware.swift
```diff
@@ -0,0 +1,62 @@
...
+                attributes["url.query"] = request.url.query
+                
+                // Recommended
+                attributes["client.address"] = request.peerAddress?.ipAddress
+                attributes["network.peer.address"] = request.remoteAddress?.ipAddress
+                attributes["network.peer.port"] = request.remoteAddress?.port
+                attributes["network.protocol.version"] = "\(request.version.major).\(request.version.minor)"
+                attributes["user_agent.original"] = request.headers[.userAgent].first
+                
+                // Opt-in
```
0xTim, line 44: Let's remove this if we're not going to have any here

Tests/VaporTests/MiddlewareTests.swift
```diff
@@ -144,4 +145,112 @@ final class MiddlewareTests: XCTestCase {
...
+        XCTAssertEqual(span.attributes["network.peer.port"]?.toSpanAttribute(), nil)
+        XCTAssertEqual(span.attributes["network.protocol.version"]?.toSpanAttribute(), "1.1")
+        
+        XCTAssertEqual(span.attributes["user_agent.original"]?.toSpanAttribute(), nil)
+        
+        XCTAssertEqual(span.attributes["http.response.status_code"]?.toSpanAttribute(), 200)
+    }
 }
+
+final class TestTracer: Tracer {
```
0xTim, line 185: Can we pull this out into a separate file?

Tests/VaporTests/MiddlewareTests.swift
```diff
@@ -144,4 +145,112 @@ final class MiddlewareTests: XCTestCase {
...
     }
+    
+    func testTracingMiddleware() async throws {
+        let tracer = TestTracer()
+        InstrumentationSystem.bootstrap(tracer)
+        app.grouped(TracingMiddleware()).get("testTracing") { req -> String in
+            return "done"
+        }
+
+        try await app.testable().test(.GET, "/testTracing") { res in
```
0xTim, line 156: I think if you do `app.testable(method: .running(port: 0))` that will make a real HTTP request to give us more attributes below

Tests/VaporTests/MiddlewareTests.swift
```diff
@@ -144,4 +145,112 @@ final class MiddlewareTests: XCTestCase {
...
+        XCTAssertEqual(span.attributes["server.address"]?.toSpanAttribute(), "127.0.0.1")
+        XCTAssertEqual(span.attributes["server.port"]?.toSpanAttribute(), 8080)
+        XCTAssertEqual(span.attributes["url.query"]?.toSpanAttribute(), nil)
+        
+        XCTAssertEqual(span.attributes["client.address"]?.toSpanAttribute(), nil)
+        XCTAssertEqual(span.attributes["network.peer.address"]?.toSpanAttribute(), nil)
+        XCTAssertEqual(span.attributes["network.peer.port"]?.toSpanAttribute(), nil)
+        XCTAssertEqual(span.attributes["network.protocol.version"]?.toSpanAttribute(), "1.1")
+        
+        XCTAssertEqual(span.attributes["user_agent.original"]?.toSpanAttribute(), nil)
```
0xTim, line 179: We could use the `beforeRequest` closure to set the user agent header to check this

0xTim (changes requested):
Thank you for this! This is great.

Some minor comments but almost ready to go

Sources/Vapor/Middleware/TracingMiddleware.swift
NeedleInAJayStack: Whoops - this was left over from when I included a few of them. I've removed.

Tests/VaporTests/MiddlewareTests.swift
NeedleInAJayStack: Yes, definitely. I've moved it to a new file at `Utilities/TestTracer.swift`

Tests/VaporTests/MiddlewareTests.swift
NeedleInAJayStack: Good idea - done.

NeedleInAJayStack:
One followup question - Given that [AsyncMiddleware uses EventLoopFuture.get](https://github.com/vapor/vapor/blob/60144400f627faa98e31914a1272337a69450bd8/Sources/Vapor/Concurrency/AsyncMiddleware.swift#L23), a span created in the `next` route handler won't be recognized as a child span of the one created in the new `TracingMiddleware`. 

What do you think about providing a `Request.span: Span?` property in which the user can manually propagate the `TracingMiddleware` span (I have a PR in to add documentation on how to do this in a NIO environment [here](https://github.com/apple/swift-distributed-tracing/pull/153/files#diff-873fc97b1dc4983f3375ba329a853b5c918784df2851ec34890134754700d76eR233))

0xTim (approved):
This looks good!

0xTim:
@NeedleInAJayStack yes we either need to do that for the user (so at the point where we bridge between worlds do it in Vapor) or add the property. I'd be in favour of doing it in Vapor to make it easy where possible.

NeedleInAJayStack:
> @NeedleInAJayStack yes we either need to do that for the user (so at the point where we bridge between worlds do it in Vapor) or add the property. I'd be in favour of doing it in Vapor to make it easy where possible.

Sounds good. I actually did both because internally propogating in Vapor needs the property anyway.

I've added the `Request.serviceContext` property in [this commit](https://github.com/vapor/vapor/pull/3253/commits/cbbf68852608161dd276e8a0be28bddaa2e7bd74), and automatically injected the service context into each route & middleware callback in [this commit](https://github.com/vapor/vapor/pull/3253/commits/865efcecf31992b1aa9bf39e80fcf497826d3a22). The pattern of middlewares ["setting and restoring" the `Request.serviceContext`](https://github.com/vapor/vapor/pull/3253/commits/cbbf68852608161dd276e8a0be28bddaa2e7bd74#diff-e86201a80ba1c380d64197003bb66a71212cd8aba3e18fbccdc39d70f4b5da70R17) should probably be copied by any users that write middlewares that create spans in order for their Route closures to nest under the correct span.

Sources/Vapor/HTTP/BasicResponder.swift
```diff
@@ -23,7 +24,9 @@ public struct BasicResponder: Responder {
     /// See `Responder`.
     public func respond(to request: Request) -> EventLoopFuture<Response> {
         do {
-            return try closure(request)
+            return try ServiceContext.withValue(request.serviceContext) {
```
0xTim, line 27: My one concern with this is a potential performance impact and whether we should put it behind a flag or not, given most people won't use it.

Thoughts @gwynne

0xTim (approved):
I think I'm good with this, just want to loop in @gwynne for a second look

Sources/Vapor/HTTP/BasicResponder.swift
gwynne: I agree - this should be behind a flag. I'm uneasy enough with including swift-distributed-tracing as a dependency, on the same principle.

0xTim: I think I'm fine to add the dependency, but I think putting it behind a flag would be the right thing to do

NeedleInAJayStack: This was actually my effort to address:

> I'd be in favour of doing it in Vapor to make it easy where possible.

If there are performance concerns, I'd suggest that we remove all the `ServiceContext.withValue(request.serviceContext) {` calls from this PR, and let the user manage the span/serviceContext instead of introducing a flag. That way, the user can be more specific and only incur the `Request` property and `Task.local` overhead to specifically where they need tracing.

For example, the user could just pass in the `req.serviceContext` into a span manually:

```swift
app.routes.get("hello") { req in
    try await withSpan(name: "hello-span", context: req.serviceContext) { span in
        try await doHello()
    }
}
```

They can always manage the `req.serviceContext` manually as well if downstream methods need it set:

```swift
app.routes.get("hello") { req in
    parentSpan = req.serviceContext
    let hello = try await withSpan(name: "hello-span", context: req.serviceContext) { span in
        req.serviceContext = span.context
        defer {
            req.serviceContext = parentSpan
        }
        return doHello()
    }
    return hello
}
```

Does that make sense to you guys? What do you think?

ktoso: Two things:
- We really need to stop worrying about adding the dependency. This is "the" API package, the same way as logging and metrics, and frameworks _should_ adopt it.
- If you're seeing some performance hit from the no-op tracer I think it's fine to have this be opt-in. A NoOp tracer is optimized to be really fast and "do nothing" but it'll have to read the bootstrapped value so yes I think it's fine to avoid the lookup if you want and just document to folks "if you want to use distributed tracing, enable <<something>>" 👍

Package.swift
```diff
@@ -47,6 +47,12 @@ let package = Package(
 
         // Swift metrics API
         .package(url: "https://github.com/apple/swift-metrics.git", from: "2.5.0"),
+        
+        // Swift tracing API
+        .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "1.0.0"),
```
ktoso, line 52: I'd probably say `1.1.0`, we did some language version specific adjustments in the 1 series.

Sources/Vapor/Middleware/TracingMiddleware.swift
```diff
@@ -0,0 +1,68 @@
+import Tracing
+
+/// Creates a trace and metadata for every request
+///
+/// See https://opentelemetry.io/docs/specs/semconv/http/http-spans/
```
ktoso, line 5: LGTM, and OK to not use the swift-distributed-tracing-extras repo -- as we'll want to change how we have organized all those repos. I think what you're doing here is safer in face of the not-stable-extras repo.

We'll want to make a stable repo that is versioned along with the `semconv`'s version, but we don't have it yet. So this is LGTM as-is

ktoso (approved):
Thank you for the adoption PR, I think this looks good! The attributes look ok as well.

@0xTim it'd be lovely to accept this and we should follow up updating the tracing repo with info that vapor supports tracing (as well as hummingbird) 👍

Package.swift
NeedleInAJayStack: Sounds good, I updated to 1.1.0.

Sources/Vapor/HTTP/BasicResponder.swift
NeedleInAJayStack: Hey @0xTim, just let me know what you think here. I'm happy to either add a tracing flag or remove the `ServiceContext.withValue` calls - whichever you prefer.

0xTim: @NeedleInAJayStack ok I think me and @gwynne are agreed that a flag is the way to go to avoid all the copies as we switch boundaries

NeedleInAJayStack: Sounds good to me! In [this commit](https://github.com/vapor/vapor/pull/3253/commits/428db1ac7f3c4276a25f993d08e09198e716c8cb) I added an `Application.traceAutoPropogation` flag that defaults to false. When enabled, trace spans are propogated automatically, incurring the copying overhead, and when disabled they are not.

I kinda took a guess on where the flag should be. Let me know if you where thinking of it happening somewhere else and I'd be happy to change. Thanks!

0xTim (changes requested):
Thanks! This looks good - just one requested change just to avoid accessing the lock several times

Sources/Vapor/Request/Request.swift
```diff
@@ -343,4 +355,15 @@ public final class Request: CustomStringConvertible, Sendable {
         self._storage = .init(.init())
         self.bodyStorage = .init(bodyStorage)
     }
+    
+    /// Automatically restores tracing serviceContext around the provided closure
+    func propagateTracingIfEnabled<T>(_ closure: () throws -> T) rethrows -> T {
+        if self.application.traceAutoPropagation {
```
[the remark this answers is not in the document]
NeedleInAJayStack: Oh yeah, absolutely. Good call. I've changed it.

0xTim (approved):
Thanks for this! I think we're good to go

gwynne (approved):
LGTM

PR #3254, opened 2024-11-16
sessionID but customizable

bwdmr:
for monitoring purposes coinciding with ID's from
third party services. the capability to determine the sessionID beforehand is crucial.

* extend function member with optional sessionID
* default value provided for add cookie function

a custom session middleware will be necessary to customize sessionID accordingly. default behaviour is not affected.

### DIFF
--- a/Sources/Vapor/Concurrency/AsyncSessionDriver.swift
+++ b/Sources/Vapor/Concurrency/AsyncSessionDriver.swift
@@ -4,17 +4,17 @@ import NIOCore
 ///
 /// This is an async version of `SessionDriver`
 public protocol AsyncSessionDriver: SessionDriver {
-    func createSession(_ data: SessionData, for request: Request) async throws -> SessionID
+    func createSession(_ sessionID: SessionID?, _ data: SessionData, for request: Request) async throws -> SessionID
     func readSession(_ sessionID: SessionID, for request: Request) async throws -> SessionData?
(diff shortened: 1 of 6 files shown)


0xTim:
This isn't usable in its current form as it's a breaking change. We can't just change protocols like that

bwdmr:
in which way is it a breaking change?

0xTim:
You removed `func createSession(_ data: SessionData, for request: Request) async throws -> SessionID` from the protocol. So anyone implementing that will now longer get called when a session is created so it will break all sessions

0xTim:
Also, could you describe the scenario in which the session ID needs to be known up ahead of time and give some examples? Because that's sounds ripe for exploitation

bwdmr:
using distributed tracing with an otel backend including a collector. which is able to generate cross service sessionIDs for service independent tracing and logging. 

to map the user stored in the db to the sessionID generated by the collector, id have to pass the pregenerated sessionID to vapor so sessions behave correctly.
https://opentelemetry.io/docs/specs/semconv/general/session/

i am not aware of a breaking change.
the function createSession has been extended by an argument, sessionID with a default value of nil,
by default nothing changes as the value remains to be nil and createCookie function will be either above the
customized driver or the default function passing nil and generate a new id.

bwdmr:
that is correct. so far i cant argue against it. 

i am ok to close this.
thanks for elaborating on your end.

PR #3257, opened 2024-11-20
Better compatibility with Swift Testing

MahdiBM:
Introduces a new `VaporTesting` module for testing with better compatibility with Swift Testing that removes usages of `XCTest`.

Use the new `testing()` extension on `Application` to write your tests. E.g.:

```swift
try await app.testing().test(.POST, "/decode") { req in
  try req.content.encode(MyContent())
} afterResponse: { res in
  #expect(res.status == .ok)
  expectContains(res.body.string, "Hello World!")
}
```


gwynne:
> The API breakage failure is incorrect as far as I can tell.

Correct - the API breakage checker, along with all its many other flaws, can't see through `@_exported` to tell that you moved the names between modules and reexported them from the original.

Sources/VaporTesting/+TestingApplicationTester.swift
```diff
@@ -0,0 +1,148 @@
...
+        try await beforeRequest(&request)
+        do {
+            return try await self.performTest(request: request)
+        } catch {
+            Issue.record("\(String(reflecting: error))", sourceLocation: sourceLocation)
+            throw error
+        }
+    }
+
+    @available(*, noasync, message: "Use the async method instead.")
```
MahdiBM, line 76: readded these 3 noasync functions as well

Sources/VaporTesting/+TestingApplicationTester.swift
```diff
@@ -0,0 +1,197 @@
...
+                filePath: filePath,
+                line: line,
+                column: column
+            )
+            Issue.record("\(String(reflecting: error))", sourceLocation: sourceLocation)
+            throw error
+        }
+    }
+
+    @available(*, noasync, message: "Use the async method instead.")
```
0xTim, line 101: Do we actually need these? Given this is a new API, why not just offer the async variants only?

Sources/XCTVapor/XCTVaporTests.swift
```diff
@@ -1,19 +1,5 @@
 import Vapor
 import XCTest
-import NIOConcurrencyHelpers
-
-/**
- We recommend configuring this in your test’s `class func setUp()`
- */
-public var app: (@Sendable () throws -> Application)! {
```
0xTim, line 8: I think we should keep this here and not offer it as part of the test utils. Anyone using Swift testing shouldn't be using this global

0xTim (changes requested):
Looks good overall, just some changes since this is a new API and no point adding on tech debt where not needed

Sources/VaporTesting/+TestingApplicationTester.swift
0xTim: Yeah the compiler gets really confused by these forcing the closure to be marked as `async` so this should solve it

Tests/VaporTests/VaporTesting.swift
```diff
@@ -0,0 +1,62 @@
...
+    func contentContainerDecode() throws {
+        struct FooContent: Content, Equatable {
+            var message: String = "hi"
+        }
+        struct FooDecodable: Decodable, Equatable {
+            var message: String = "hi"
+        }
+
+        let app = Application(.testing)
+        defer { app.shutdown() }
```
0xTim, line 20: We should make this async and use the with syntax given that's what we want to push people towards (and test it etc)

Tests/VaporTests/VaporTesting.swift
```diff
@@ -0,0 +1,62 @@
...
+
+        app.routes.post("decode") { req async throws -> String in
+            #expect(try req.content.decode(FooContent.self) == FooContent())
+            #expect(try req.content.decode(FooDecodable.self, as: .json) == FooDecodable())
+            return "decoded!"
+        }
+
+        try app.testable().test(.POST, "/decode") { req in
```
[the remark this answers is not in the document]
MahdiBM: hmm interesting how it works anyway although i haven't imported XCTVapor.

MahdiBM:
@0xTim interesting find: Swift struggles to find the correct function even when i haven't imported XCTVapor at all.
I had to do something (add a parameter name) to help Swift disambiguate between the functions: https://github.com/vapor/vapor/pull/3257/commits/da428df9fcc06b10c7e75451c718a7724f8eb850

0xTim:
I'm not completely happy with the added parameter. If you don't have XCTVapor imported it shouldn't be able to see those functions so I'm not sure why it would fail. What was the actual error message?

MahdiBM:
@0xTim 
> If you don't have XCTVapor imported it shouldn't be able to see those functions so I'm not sure why it would fail. What was the actual error message?

On another project it works fine, but here in the VaporTesting test file, I could confirm that it picks the XCT function, although I haven't imported XCTVapor.

0xTim:
> On another project it works fine, but here in the VaporTesting test file, I could confirm that it picks the XCT function, although I haven't imported XCTVapor.

That's probably just SwiftPM being weird because it's the same module. Feel free to namespace the calls in the tests if needed to make it work

Sources/VaporTesting/TestingApplicationTester.swift
```diff
@@ -0,0 +1,140 @@
+import NIOHTTP1
+import NIOCore
+#if compiler(>=6.0)
+import Testing
+#endif
+
+public protocol TestingApplicationTester: Sendable {
+    @available(*, noasync, message: "Use the async method instead.")
```
0xTim, line 8: This is a new module right? So we can removed the deprecation and only have the async variant

Sources/VaporTesting/TestingApplicationTester.swift
```diff
@@ -0,0 +1,140 @@
...
+        try self.boot()
+        switch method {
+        case .inMemory:
+            return try InMemory(app: self)
+        case let .running(hostname, port):
+            return try Live(app: self, hostname: hostname, port: port)
+        }
+    }
+
+    @available(*, noasync, message: "Use the async method instead.")
```
0xTim, line 27: See above

0xTim (changes requested):
Last couple of changes

Sources/VaporTesting/TestingApplicationTester.swift
MahdiBM: Ah right

MahdiBM: actually i had this change, just forgot to commit

0xTim (approved):

PR #3265, opened 2024-12-13
Fix Concurrency Warnings and Deprecations

0xTim:
* Fix some deprecation warnings caused by `HTTPMethod.name` being deprecated (not sure why this wasn't caught during the deprecation)
* Fix sendable warning related to `HTTPClient.Response` not being marked as `Sendable` - there's no point moving over to the async APIs for now
* Fix more Sendable warnings related to `.get()` and friends requiring values to be `Sendable` - see https://github.com/apple/swift-nio/releases/tag/2.76.0
* As a result, `View` and `Content` now have sendability requirements

### DIFF
--- a/Sources/XCTVapor/XCTApplication.swift
+++ b/Sources/XCTVapor/XCTApplication.swift
@@ -1,4 +1,4 @@
-@preconcurrency import AsyncHTTPClient
+import AsyncHTTPClient
 import NIOCore
 import NIOHTTP1
 import XCTest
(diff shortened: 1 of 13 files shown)


Sources/XCTVapor/XCTApplication.swift
```diff
@@ -1,4 +1,4 @@
-import AsyncHTTPClient
+@preconcurrency import AsyncHTTPClient
```
gwynne, line 1: It should be possible to resolve Sendable warnings coming from AHC rather than silencing them.

0xTim: This is caused by the tightening of restrictions in https://github.com/apple/swift-nio/releases/tag/2.76.0

It would work apart from `.get()` which can't be wrapped in a box. I'm going to submit a PR to AHC

0xTim: Turns out just an outdated AHC, all resolved

gwynne (approved):

PR #3274, opened 2024-12-26
Minor swift-testing related improvements

MahdiBM:


Sources/VaporTesting/VaporTestingContext.swift
```diff
@@ -12,7 +12,7 @@ public enum VaporTestingContext {
         column: Int
     ) {
         let shouldWarn = VaporTestingContext.emitWarningIfCurrentTestInfoIsUnavailable ?? true
-        let isNotInSwiftTesting = Test.current == nil
+        var isNotInSwiftTesting: Bool { Test.current == nil }
```
MahdiBM, line 15: Shouldn't matter much but still an improvement.

gwynne (approved):
Definite improvements here!

Sources/VaporTesting/VaporTestingContext.swift
gwynne: Can't hurt, if nothing else.

PR #3275, opened 2024-12-28
In withApp(), always shut down app even if error is thrown

gwynne:
This method is documented as "while handling the lifecycle of the application", so let's actually do so.


gwynne:
Note: Waits to return `result` until after the `catch` block so that errors thrown by `asyncShutdown()` don't cause it to be called twice.

MahdiBM (approved):

PR #3278, opened 2025-01-05
Pass session to cookie session factory allowing to modify cookie based on session data

rafalwojcik:
### Motivation

I was struggling to implement a “Remember me” feature for user login. Even after adding middleware like the snippet below, the SessionMiddleware kept overriding my cookie values.

```swift
struct ExtendSessionMiddleware<A: Authenticatable>: AsyncMiddleware {
    let authenticableType: A.Type

    init(authenticableType: A.Type) {
        self.authenticableType = authenticableType
    }

    func respond(to req: Request, chainingTo next: AsyncResponder) async throws -> Response {
        guard
            req.auth.has(authenticableType),
            let rememberMe = Bool(req.session.data["remeberMe"] ?? "")
        else { return try await next.respond(to: req) }

        let response = try await next.respond(to: req)

        let extendTime: TimeInterval = rememberMe ? 7 * 24 * 60 * 60 : 24 * 60 * 60
        let expiresAt = Date().addingTimeInterval(extendTime)

        let cookieName = req.application.sessions.configuration.cookieName
        if var cookie = req.cookies[cookieName] {
            cookie.expires = expiresAt
            cookie.maxAge = Int(extendTime)
            response.cookies[cookieName] = cookie
        }

        return response
    }
}
```

### Idea

My proposed solution is to pass the session itself to the cookie factory. This approach allows reading session values set during user login and overriding the default session behavior.

### Usage

```swift
var config = SessionsConfiguration.default()
config.cookieFactory = { sessionID, session in
    let rememberMe = Bool(session.data["remeberMe"] ?? "")
    let time: TimeInterval = (rememberMe ?? false) ? 7 * 24 * 60 * 60 : 24 * 60 * 60
    let expiresAt = Date().addingTimeInterval(time)
    return .init(string: sessionID.string, expires: expiresAt)
}
application.middleware.use(SessionsMiddleware(session: application.sessions.driver, configuration: config))
```

With this code, the session expires after one day if the user doesn’t select “Remember me,” and after seven days if they do.

### Question
Is there another way to solve this issue without requiring any changes from my PR?

### DIFF
--- a/Sources/Vapor/Sessions/SessionsConfiguration.swift
+++ b/Sources/Vapor/Sessions/SessionsConfiguration.swift
@@ -3,7 +3,7 @@ import Foundation
 /// Configuration options for sessions.
 public struct SessionsConfiguration: Sendable {
     /// Creates a new `HTTPCookieValue` for the supplied value `String`.
-    public var cookieFactory: @Sendable (SessionID) -> HTTPCookies.Value
+    public var cookieFactory: @Sendable (SessionID, Session) -> HTTPCookies.Value
 
     /// Name of HTTP cookie, used as a key for the cookie value.
     public var cookieName: String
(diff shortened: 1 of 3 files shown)


MurdoqZodoq (approved):

rafalwojcik:
I'm writing tests now for this functionality

0xTim:
This is a no go at the moment as it changes the public API. Here's an example of doing it via middleware https://github.com/brokenhandsio/SteamPress/blob/vapor4/Sources/SteamPress/Middleware/BlogRememberMeMiddleware.swift

rafalwojcik:
@0xTim, unfortunately, the above solution doesn't work with Vapor 4. I can understand that breaking public API won't go to the Vapor 4 release. Could you consider including it in Vapor 5?

rafalwojcik:
@0xTim  I've rebased changes over `vapor-5` branch and updated PR base

0xTim:
We can definitely look at improving the usability and API for Vapor 5. But that solution should work fine for Vapor 4 since I built it for that

rafalwojcik:
Hmm, I've tried almost all things, and `SessionsMiddleware` is always called at the end of the middleware chain, so it overrides cookie value.

0xTim:
Register the remember me middleware before the sessions middleware and it should work

rafalwojcik:
When I registered it, before I couldn't access session because it not initalized and fatalError is triggered.

MurdoqZodoq (approved):

PR #3280, opened 2025-01-11
Adds Support for Tracing Header Extraction

NeedleInAJayStack:
This adds header extraction support for tracing backends as described in [this document](https://swiftpackageindex.com/apple/swift-distributed-tracing/main/documentation/tracing/instrumentyourlibrary#Handling-inbound-requests). In the context of OTEL, this is used to correlate trace data across the HTTP boundary by passing the `traceparent` and `tracestate` header, as [documented here](https://www.w3.org/TR/trace-context/#header-name) and implemented in the [swift-otel package here](https://github.com/swift-otel/swift-otel/blob/main/Sources/OTel/Tracing/Propagating/OTelW3CPropagator.swift).

It also adds a closure to `TracingMiddleware` that allows users to set custom span metadata on requests.

Finally, it fixes a few warnings caused by the MiddlewareTest function introduced in https://github.com/vapor/vapor/pull/3253

### DIFF
--- a/Sources/Vapor/Middleware/TracingMiddleware.swift
+++ b/Sources/Vapor/Middleware/TracingMiddleware.swift
@@ -16,7 +16,9 @@ public final class TracingMiddleware: AsyncMiddleware {
     }
     
     public func respond(to request: Request, chainingTo next: any AsyncResponder) async throws -> Response {
-        let parentContext = request.serviceContext
+        var parentContext = request.serviceContext
+        InstrumentationSystem.instrument.extract(request.headers, into: &parentContext, using: HTTPHeadersExtractor())
+        
         return try await withSpan(
             // Name: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name
             request.route?.description ?? "vapor_route_undefined",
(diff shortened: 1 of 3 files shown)


Sources/Vapor/Middleware/TracingMiddleware.swift
```diff
@@ -66,3 +80,16 @@ public final class TracingMiddleware: AsyncMiddleware {
         }
     }
 }
+
+// Allows backends to extract information from the request headers. For example, in OTel W3C, this allows frontend/backend
+// correlation using the `traceparent` and `tracestate` headers. For more information, see
```
gwynne, line 85: See what?

NeedleInAJayStack: Oh my gosh, great catch 🤦. I added the missing link here.

Sources/Vapor/Middleware/TracingMiddleware.swift
```diff
@@ -4,10 +4,21 @@ import Tracing
...
 public final class TracingMiddleware: AsyncMiddleware {
-    public init() {}
+    private let setCustomAttributes: @Sendable (inout SpanAttributes, Request) -> Void
+    
+    /// Create a TracingMiddleware
+    /// - Parameter setCustomAttributes: Closure that allows setting custom span attributes for a particular request. A custom span attribute could be extracted from a request
+    /// header, for example. This closure is called during span creation on every request, so should be lightweight.
+    public init(
+        setCustomAttributes: @escaping @Sendable (inout SpanAttributes, Request) -> Void = { _, _ in }
+    ) {
```
gwynne, line 14: This will get rid of the API breakage complaint (which is for once technically correct).

```suggestion
    public init() {
        self.setCustomAttributes = { _, _ in }
    }
    
    public init(
        setCustomAttributes: @escaping @Sendable (inout SpanAttributes, Request) -> Void
    ) {
```

NeedleInAJayStack: Oh huh, I guess I thought adding a parameter with a default value was non-breaking. In any case, I've adjusted to your suggestion.

0xTim: Technically it's a break but normally we don't treat it as such in Vapor. It does however keep the API checker quiet

0xTim (approved):
Thanks!

PR #3285, opened 2025-01-15
Fix warnings from NIO 2.79.0

0xTim:
Fixes a number of `Sendable` warnings introduced by https://github.com/apple/swift-nio/releases/tag/2.79.0

This also deprecates the main `Application.init()` API that was blocking on an event loop that has been marked as `noasync` for a while. Vapor users should migrate to the async APIs.

### DIFF
--- a/Sources/Vapor/Environment/Environment+Secret.swift
+++ b/Sources/Vapor/Environment/Environment+Secret.swift
@@ -1,6 +1,7 @@
 import NIOCore
 import NIOPosix
 import AsyncKit
+import _NIOFileSystem
 
 extension Environment {
     /// Reads a file's content for a secret. The secret key is the name of the environment variable that is expected to
@@ -27,6 +28,7 @@ extension Environment {
     ///
     /// - Important: Do _not_ use `.wait()` if loading a secret at any time after the app has booted, such as while
     ///   handling a `Request`. Chain the result as you would any other future instead.
+    @available(*, deprecated, message: "Use an async version of load instead")
     public static func secret(key: String, fileIO: NonBlockingFileIO, on eventLoop: EventLoop) -> EventLoopFuture<String?> {
         guard let filePath = self.get(key) else {
             return eventLoop.future(nil)
@@ -45,6 +47,7 @@ extension Environment {
     /// - Returns:
     ///   - On success, a succeeded future with the loaded content of the file.
     ///   - On any kind of error, a succeeded future with a value of `nil`. It is not currently possible to get error details.
+    @available(*, deprecated, message: "Use an async version of load instead")
     public static func secret(path: String, fileIO: NonBlockingFileIO, on eventLoop: EventLoop) -> EventLoopFuture<String?> {
         return fileIO
             .openFile(path: path, eventLoop: eventLoop)
@@ -62,5 +65,26 @@ extension Environment {
             .recover { _ -> String? in
                 nil
             }
+    }
+
+    /// Load the content of a file at a given path as a secret.
+    ///
+    /// - Parameters:
+    ///   - path: Path to the file containing the secret
+    ///
+    /// - Returns:
+    ///   - On success, the loaded content of the file.
+    ///   - On any kind of error `nil`. It is not currently possible to get error details.
+    public static func secret(path: String) async throws -> String? {
+        do {
+            return try await FileSystem.shared.withFileHandle(forReadingAt: .init(path)) { handle in
+                let buffer = try await handle.readToEnd(maximumSizeAllowed: .megabytes(32))
+                return buffer
+                    .getString(at: buffer.readerIndex, length: buffer.readableBytes)!
+                    .trimmingCharacters(in: .whitespacesAndNewlines)
+            }
+        } catch {
+            return nil
+        }
     }
 }
--- a/Sources/Vapor/Environment/Environment+Secret.swift
+++ b/Sources/Vapor/Environment/Environment+Secret.swift
@@ -28,7 +28,7 @@ extension Environment {
     ///
     /// - Important: Do _not_ use `.wait()` if loading a secret at any time after the app has booted, such as while
     ///   handling a `Request`. Chain the result as you would any other future instead.
-    @available(*, deprecated, message: "Use an async version of load instead")
+    @available(*, deprecated, message: "Use an async version of secret instead")
     public static func secret(key: String, fileIO: NonBlockingFileIO, on eventLoop: EventLoop) -> EventLoopFuture<String?> {
         guard let filePath = self.get(key) else {
             return eventLoop.future(nil)
@@ -47,7 +47,7 @@ extension Environment {
     /// - Returns:
     ///   - On success, a succeeded future with the loaded content of the file.
     ///   - On any kind of error, a succeeded future with a value of `nil`. It is not currently possible to get error details.
-    @available(*, deprecated, message: "Use an async version of load instead")
+    @available(*, deprecated, message: "Use an async version of secret instead")
     public static func secret(path: String, fileIO: NonBlockingFileIO, on eventLoop: EventLoop) -> EventLoopFuture<String?> {
         return fileIO
             .openFile(path: path, eventLoop: eventLoop)
@@ -86,5 +86,32 @@ extension Environment {
         } catch {
             return nil
         }
+    }
+
+    /// Reads a file's content for a secret. The secret key is the name of the environment variable that is expected to
+    /// specify the path of the file containing the secret.
+    ///
+    /// - Parameters:
+    ///   - key: The environment variable name
+    ///   - fileIO: `NonBlockingFileIO` handler provided by NIO
+    ///   - eventLoop: `EventLoop` for NIO to use for working with the file
+    ///
+    /// Example usage:
+    ///
+    /// ````
+    /// func configure(_ app: Application) async throws {
+    ///     // ...
+    ///
+    ///     let databasePassword = try await Environment.secret(key: "DATABASE_PASSWORD_FILE")
+    ///
+    /// ````
+    ///
+    /// - Important: Do _not_ use `.wait()` if loading a secret at any time after the app has booted, such as while
+    ///   handling a `Request`. Chain the result as you would any other future instead.
+    public static func secret(key: String) async throws -> String? {
+        guard let filePath = self.get(key) else {
+            return nil
+        }
+        return try await self.secret(path: filePath)
     }
 }
--- a/Sources/Vapor/Environment/Environment+Secret.swift
+++ b/Sources/Vapor/Environment/Environment+Secret.swift
@@ -93,8 +93,6 @@ extension Environment {
     ///
     /// - Parameters:
     ///   - key: The environment variable name
-    ///   - fileIO: `NonBlockingFileIO` handler provided by NIO
-    ///   - eventLoop: `EventLoop` for NIO to use for working with the file
     ///
     /// Example usage:
     ///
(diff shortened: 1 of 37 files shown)


gwynne (approved):

0xTim:
@gwynne lots of changes since the last review including some significant deprecations

Sources/Vapor/Environment/Environment+Secret.swift
```diff
@@ -63,4 +66,52 @@ extension Environment {
...
+    /// ````
+    /// func configure(_ app: Application) async throws {
+    ///     // ...
+    ///
+    ///     let databasePassword = try await Environment.secret(key: "DATABASE_PASSWORD_FILE")
+    ///
+    /// ````
+    ///
+    /// - Important: Do _not_ use `.wait()` if loading a secret at any time after the app has booted, such as while
+    ///   handling a `Request`. Chain the result as you would any other future instead.
```
gwynne, line 110: This comment makes no sense here.

Sources/Vapor/Environment/Environment+Secret.swift
```diff
@@ -63,4 +66,52 @@ extension Environment {
...
+        }
+    }
+
+    /// Reads a file's content for a secret. The secret key is the name of the environment variable that is expected to
+    /// specify the path of the file containing the secret.
+    ///
+    /// - Parameters:
+    ///   - key: The environment variable name
+    ///   - fileIO: `NonBlockingFileIO` handler provided by NIO
+    ///   - eventLoop: `EventLoop` for NIO to use for working with the file
```
gwynne, line 97: These parameters do not exist.

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -330,16 +333,17 @@ public struct FileIO: Sendable {
     ///   - path: The file's path.
     ///   - lastModified: When the file was last modified.
     /// - Returns: An `EventLoopFuture<String>` which holds the ETag.
```
gwynne, line 335: No longer returns a future.

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -407,6 +411,8 @@ public struct FileIO: Sendable {
     ///        print("chunk: \(data)")
     ///    }
     ///
+    ///    **Warning** it's the caller's responsibility to close the file handle provided in ``FileChunks`` when finished.
+    ///
```
gwynne, line 415: Use a warning callout instead:
```suggestion
    ///  > Warning: It is the caller's responsibility to close the file handle provided in ``FileChunks`` when finished.
    ///
```

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -441,18 +447,19 @@ public struct FileIO: Sendable {
     ///     let data = ByteBuffer(string: "ByteBuffer")
     ///     try await req.fileio.writeFile(data, at: "/path/to/file.txt")
     ///
+    /// **WARNING**: this will overwrite the file if it already exists.
```
gwynne, line 450:
```suggestion
    /// > Warning: This method will overwrite the file if it already exists.
```

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -441,18 +447,19 @@ public struct FileIO: Sendable {
...
+    ///
     /// - parameters:
     ///     - path: Path to file on the disk.
     ///     - buffer: The `ByteBuffer` to write.
     /// - returns: `Void` when the file write is finished.
```
gwynne, line 455: This line is unnecessary.
```suggestion
```

Tests/VaporTests/AsyncAuthTests.swift
```diff
@@ -244,4 +280,48 @@ final class AsyncAuthenticationTests: XCTestCase {
         var config = Middlewares()
         config.use(Test.authenticator())
     }
+
+    func testAsyncAuthenticator() throws {
```
gwynne, line 284: If this is an async authenticator test, why is it using `BasicAuthenticator` instead of `AsyncBasicAuthenticator`?

Tests/VaporTests/AsyncClientTests.swift
```diff
@@ -6,8 +6,8 @@ import NIOCore
 import Logging
 import NIOEmbedded
 
-final class AsyncClientTests: XCTestCase {
-    
+final class AsyncClientTests: XCTestCase, @unchecked Sendable {
```
gwynne, line 9: Can this be replaced with usage of `@MainActor` instead?

Tests/VaporTests/AsyncClientTests.swift
```diff
@@ -159,6 +183,31 @@ final class AsyncClientTests: XCTestCase {
...
+        let startingPistol = DispatchGroup()
+        startingPistol.enter()
+        startingPistol.enter()
+
+        let finishLine = DispatchGroup()
+        finishLine.enter()
+        Thread.async {
+            startingPistol.leave()
+            startingPistol.wait()
+            XCTAssert(type(of: self.app.http.client.shared) == HTTPClient.self)
```
gwynne, line 196: What does this actually verify?

Tests/VaporTests/DotEnvTests.swift
```diff
@@ -35,6 +36,36 @@ final class DotEnvTests: XCTestCase {
         try elg.syncShutdownGracefully()
     }
 
+    func testReadFile() async throws {
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        let pool = NIOThreadPool(numberOfThreads: 1)
+        pool.start()
```
gwynne, line 42: What is the ELG for? It doesn't seem to be used at all. Also, why not use the singletons?

gwynne (changes requested):
Several nits

Tests/VaporTests/AsyncClientTests.swift
0xTim: Nope because some of the async functions have a different isolation where self is passed in so you still get warnings. I've half a mind to just convert this to VaporTesting but I'll leave that for now

Tests/VaporTests/DotEnvTests.swift
0xTim: Legacy copy/paste - fixed!

Tests/VaporTests/AsyncClientTests.swift
0xTim: This looks like a Tanner special but I think it's testing Vapor's storage and retrieving the client from 2 different threads

gwynne: It's hardly a reliable test for a race condition... I'm of a mind to say do't bother with it at all.

0xTim: Removed! We still have it in the other client tests and it's no longer a thing with Vapor 5 anyway

Tests/VaporTests/AsyncClientTests.swift
```diff
@@ -126,6 +130,26 @@ final class AsyncClientTests: XCTestCase {
...
+        let data = try res.content.decode(AnythingResponse.self)
+        XCTAssertEqual(data.json, ["hello": "world"])
+        XCTAssertEqual(data.headers["content-type"], "application/json; charset=utf-8")
+    }
+
+    func testClientTimeout() async throws {
+        try await app.asyncBoot()
+
+        XCTAssertNoThrow(try app.client.get("http://localhost:\(remoteAppPort!)/json") { $0.timeout = .seconds(1) }.wait())
+        XCTAssertThrowsError(try app.client.get("http://localhost:\(remoteAppPort!)/stalling") { $0.timeout = .milliseconds(200) }.wait()) {
```
gwynne, line 147: `.wait()` in an `async` context 🙁

Tests/VaporTests/AsyncAuthTests.swift
gwynne: The way I read it, "someone" who shall remain unnecessarily named thought "hop to a thread pool and then back onto the even loop" counted as testing whether returning a non-trivial future worked. This seems thoroughly meaningless to me, I suggest rewriting it as a test of `AsyncBasicAuthenticator` (assuming there isn't already such a test, in which case I say delete this one completely).

Tests/VaporTests/DotEnvTests.swift
gwynne: Did I miss a commit? This doesn't look fixed yet.

Tests/VaporTests/MiddlewareTests.swift
```diff
@@ -177,19 +176,20 @@ final class MiddlewareTests: XCTestCase {
...
-                XCTAssertEqual(response.body.string, "done")
-            }
-        )
-        
+        try await app.server.start(address: .hostname("127.0.0.1", port: 0))
+
+        guard let port = app.http.server.shared.localAddress?.port else {
+            XCTFail("Failed to get port for app")
+            return
+        }
```
gwynne, line 184:
```suggestion
        let port = try XCTUnwrap(app.http.server.shared.localAddress?.port, "Failed to get port for app")
```

Sources/Vapor/Environment/Environment+Secret.swift
```diff
@@ -63,4 +66,47 @@ extension Environment {
...
     }
+
+    /// Load the content of a file at a given path as a secret.
+    ///
+    /// - Parameters:
+    ///   - path: Path to the file containing the secret
+    ///
+    /// - Returns:
+    ///   - On success, the loaded content of the file.
+    ///   - On any kind of error `nil`. It is not currently possible to get error details.
```
gwynne, line 77: Why _don't_ we let the actual errors get thrown with this API? A `nil` return should (ideally) be reserved for indicating "no such secret" IMO.

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -407,6 +411,8 @@ public struct FileIO: Sendable {
     ///        print("chunk: \(data)")
     ///    }
     ///
+    ///    > Warning: It's the caller's responsibility to close the file handle provided in ``FileChunks`` when finished.
```
gwynne, line 414: Still incorrectly indented.
```suggestion
    /// > Warning: It's the caller's responsibility to close the file handle provided in ``FileChunks`` when finished.
```

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -441,18 +447,18 @@ public struct FileIO: Sendable {
     ///     let data = ByteBuffer(string: "ByteBuffer")
     ///     try await req.fileio.writeFile(data, at: "/path/to/file.txt")
     ///
+    /// > Warning: This method will overwrite the file if it already exists.
+    ///
     /// - parameters:
     ///     - path: Path to file on the disk.
     ///     - buffer: The `ByteBuffer` to write.
```
gwynne, line 454: While I'm being pedantic, these are in the wrong order.

Tests/VaporTests/AsyncClientTests.swift
gwynne: It's the use of `@autoclosure` - the resulting inner closure can't be inferred as `async` because `reasync` isn't a thing, so the compiler sees a non-async context. Technically this is an issue with _all_ XCTest assertions (which is why I had https://gist.github.com/gwynne/2fbbc57eaad942368ea87dd0d2e6628e before swift-testing came along).

0xTim: Ok problem solved, I just migrated to Swift testing instead

Sources/Vapor/Environment/Environment+Secret.swift
0xTim: My guess is so we don't throw an error when a `.env` file is missing - this function is for loading an entire file, not just a single secret. It's a bizarre API though, I do admit

gwynne (approved):
Good enough for me 🙂

Pezo808 (commented):
gut

PR #3289, opened 2025-01-23
Fix `request.body.data` becomes nil instead of empty when the request body is empty

### LINKED ISSUE #3288: Unable to Decode Empty Body
### Describe the issue

`request.content.decode(Foo.self)` consistently fails when the body is empty, which is not the desired behavior

### Vapor version

4.111.1

macOS 15.2

### Swift version

Apple Swift version 6.0.3 (swiftlang-6.0.3.1.10 clang-1600.0.30.1) Target: arm64-apple-macosx15.0

### Steps to reproduce

- Server

```swift
import Vapor

let app = try await Application.make()

struct PostBody: Decodable {
    var value: String?
}

app.routes.post("foo") { request in
    let body = try request.content.decode(PostBody.self)
    return body.value ?? "empty"
}

try await app.execute()
```

- Client

```sh
curl 'http://localhost:8080/foo' \
-X 'POST' \
-H 'Content-Type: application/x-www-form-urlencoded'
```

- Client output

```
{"error":true,"reason":"Unprocessable Entity"}
```

output should be `empty`.

### Outcome

When using `application/x-www-form-urlencoded`, even if the body is empty, decoding should succeed as long as all properties of the type being decoded are optional.
However, this currently results in an error due to the following lines:
https://github.com/vapor/vapor/blob/8589cb562feab069f2563bdcdeb8f9608a07a2c7/Sources/Vapor/Request/Request.swift#L141-L144

I believe this issue arises because when Vapor reads the request, it leaves `request.bodyStorage` as `.none` if the body is empty. It should instead set `.collected(ByteBuffer())` to explicitly denote that the body is empty.

### Additional notes

_No response_

sidepelican:
When parsing a request, if it is determined that the body is empty, the `request.bodyStorage` is set to `.collected(ByteBuffer())`, resulting in `request.body.data` being empty data.
This ensures that even with an empty body, such as with `x-www-form-urlencoded`, decoding can proceed correctly.

- Fixes: https://github.com/vapor/vapor/issues/3288

### DIFF
--- a/Sources/Vapor/HTTP/Server/HTTPServerRequestDecoder.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServerRequestDecoder.swift
@@ -105,6 +105,7 @@ final class HTTPServerRequestDecoder: ChannelDuplexHandler, RemovableChannelHand
             switch self.requestState {
             case .ready: assertionFailure("Unexpected state: \(self.requestState)")
             case .awaitingBody(let request):
+                request.bodyStorage.withLockedValue { $0 = .collected(ByteBuffer()) }
                 context.fireChannelRead(self.wrapInboundOut(request))
             case .awaitingEnd(let request, let buffer):
                 request.bodyStorage.withLockedValue { $0 = .collected(buffer) }
--- a/Tests/VaporTests/ContentTests.swift
+++ b/Tests/VaporTests/ContentTests.swift
@@ -154,6 +154,32 @@ final class ContentTests: XCTestCase {
         }
     }
 
+    func testContentContainerDecodeEmptyBody() throws {
+        struct FooContent: Content, Equatable {
+            var message: String?
+        }
+        struct FooDecodable: Decodable, Equatable {
+            var message: String?
+        }
+
+        let app = Application(.testing)
+        defer { app.shutdown() }
+
+        app.routes.post("decode") { req async throws -> String in
+            XCTAssertEqual(try req.content.decode(FooContent.self), FooContent())
+            XCTAssertEqual(try req.content.decode(FooDecodable.self, as: .urlEncodedForm), FooDecodable())
+            return "decoded!"
+        }
+
+        // INFO: The .inMemory tester cannot make requests with an empty body
+        try app.testable(method: .running(port: 0)).test(.POST, "/decode") { req in
+            req.headers.contentType = .urlEncodedForm
+        } afterResponse: { res in
+            XCTAssertEqual(res.status, .ok)
+            XCTAssertContains(res.body.string, "decoded!")
+        }
+    }
+
     func testMultipartDecode() throws {
         let data = """
         --123\r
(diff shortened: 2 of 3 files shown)


0xTim:
I don't actually think this is the right approach as it could have unintended consequences for other decoders. Especially JSON - that should fail without a body. I think this should be fixed inside the URL form decoder

sidepelican:
I think it would be the same if the JSON decoder handled it internally.
The crux of the problem is that it's impossible to distinguish whether the request body was empty or cannot read.

sidepelican:
This issue is not limited to the URL form decoder; it is a broader problem that affects all decoders that can accept an empty body.
I think many people are not aware of the problem because they only use JSON.

sidepelican:
@0xTim Could you review this?

sidepelican:
@0xTim  Sorry, I might have caused some confusion with my explanation.

No matter how much we tweak `FormDataDecoder`, it won’t resolve this issue. That’s because if `body.data` becomes `nil`, the process doesn’t even reach `FormDataDecoder`.

The error occurs here:

https://github.com/vapor/vapor/blob/1fdd1c2ff090427476f534fcc9b9eb45a195c390/Sources/Vapor/Request/Request.swift#L141-L144

This handling works fine for JSONDecoder, but it may not be correct for other decoders.
That said, I don’t think this handling itself is the problem. The real issue, in my view, is that there are cases where the body is being read correctly, yet `request.body.data` still ends up as `nil`. That’s why I went with the approach in this PR.

gwynne:
This makes sense to me; if the lack of a body never reaches the decoders, the decoders can't make their respective decisions about how to handle it. I'm not sure what else a change at this level might affect, though.

0xTim:
This would be a major break, removing automatic error handling and unprocessible responses in existing code. What we probably need to do is have a look up one decoders to see what can have an empty body and use the content type header to determine which to use. I'm pretty sure this is only applicable to url encoded forms

0xTim:
@sidepelican Where did you hit this bug? Because I'm trying to dig through the URI/HTTP spec to ensure that no body is actually valid (as opposed to an empty body)

sidepelican:
@0xTim  
The server-side source code is described in: https://github.com/vapor/vapor/issues/3288

The browser-side HTML is in the following format:

```html
<form action="/list" method="POST">
<input type="hidden" name="filter" value="some filtering text">
</form>
```

The following line is added only when performing an advanced search:
`<input type="hidden" name="filter" value="some filtering text">`

On a certain list page, the form is used without any parameters under normal circumstances, but when the user performs an advanced search, a form with filtering parameters is used instead.
I’d like to handle this on the server side with a single endpoint.

```swift
struct PostBody: Decodable {
    var filter: String?
}
```

I want to receive the data in this format, but currently, it cannot handle in vapor.

0xTim:
Ok I see what you mean. I think right now, especially Vapor 4, this probably just falls underneath unsupported behaviour. The spec isn't very clear about whether no body is the same as an empty body and it's probably up to implementors to decide.

For your use case you have a couple of options - check the length of the body before decoding, check headers or just wrap the decode in a `do`/`catch`

PR #3295, opened 2025-01-29
Add VaporTesting to api-docs.yml

fpseverino:

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -10,5 +10,5 @@ jobs:
      secrets: inherit
      with:
        package_name: vapor
-       modules: Vapor,XCTVapor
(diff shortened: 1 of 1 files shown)


.github/workflows/api-docs.yml
```diff
@@ -10,5 +10,5 @@ jobs:
      secrets: inherit
      with:
        package_name: vapor
-       modules: Vapor,XCTVapor
+       modules: Vapor,XCTVapor,VaporTesting
        pathsToInvalidate: /vapor/* /xctvapor/*
```
gwynne, line 14: This needs to be updated too.

gwynne (changes requested):

gwynne (approved):

PR #3298, opened 2025-02-07
Use singleton NIOThreadPool by default

gwynne:
All Vapor applications currently end up with at least two running thread pools - the global singleton and the one created by Vapor by default. This change configures the global singleton (whose thread count defaults to `System.coreCount` anyway) as the default, eliminating the additional thread pool. This is especially nice for backtraces, as it cuts down noticeably on the number of excess threads in a Vapor process.

We also now ignore `UnsupportedOperation` errors during thread pool shutdown, which is necessary for using the singleton pool, and was previously preventing users from explicitly configuring use of the singleton thread pool, due to the (inconsistent) use of `try!`.

### DIFF
--- a/Sources/Vapor/Core/Core.swift
+++ b/Sources/Vapor/Core/Core.swift
@@ -43,6 +43,7 @@ extension Application {
                     try $0.syncShutdownGracefully()
                 } catch is NIOThreadPoolError.UnsupportedOperation {
                     // ignore, singleton thread pool throws this error on shutdown attempts
+                    // see https://github.com/apple/swift-nio/blob/c51907a839e63ebf0ba2076bba73dd96436bd1b9/Sources/NIOPosix/NIOThreadPool.swift#L142-L147
                 } catch {
                     fatalError("Unexpected error shutting down old thread pool")
                 }
@@ -105,6 +106,7 @@ extension Application {
                     try application.threadPool.syncShutdownGracefully()
                 } catch is NIOThreadPoolError.UnsupportedOperation {
                     // ignore, singleton thread pool throws this error on shutdown attempts
+                    // see https://github.com/apple/swift-nio/blob/c51907a839e63ebf0ba2076bba73dd96436bd1b9/Sources/NIOPosix/NIOThreadPool.swift#L142-L147
                 } catch {
                     application.logger.debug("Failed to shutdown thread pool", metadata: ["error": "\(error)"])
                 }
@@ -117,6 +119,7 @@ extension Application {
                     try await application.threadPool.shutdownGracefully()
                 } catch is NIOThreadPoolError.UnsupportedOperation {
                     // ignore, singleton thread pool throws this error on shutdown attempts
+                    // see https://github.com/apple/swift-nio/blob/c51907a839e63ebf0ba2076bba73dd96436bd1b9/Sources/NIOPosix/NIOThreadPool.swift#L142-L147
                 } catch {
                     application.logger.debug("Failed to shutdown thread pool", metadata: ["error": "\(error)"])
                 }
--- a/Sources/Vapor/Core/Core.swift
+++ b/Sources/Vapor/Core/Core.swift
@@ -39,7 +39,13 @@ extension Application {
             }
 
             self.core.storage.threadPool.withLockedValue({
-                try! $0.syncShutdownGracefully()
+                do {
+                    try $0.syncShutdownGracefully()
+                } catch is NIOThreadPoolError.UnsupportedOperation {
+                    // ignore, singleton thread pool throws this error on shutdown attempts
+                } catch {
+                    fatalError("Unexpected error shutting down old thread pool")
+                }
                 $0 = newValue
                 $0.start()
             })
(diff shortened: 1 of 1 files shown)


Sources/Vapor/Core/Core.swift
```diff
@@ -95,16 +95,24 @@ extension Application {
 
         struct LifecycleHandler: Vapor.LifecycleHandler {
             func shutdown(_ application: Application) {
-                try! application.threadPool.syncShutdownGracefully()
+                do {
+                    try application.threadPool.syncShutdownGracefully()
+                } catch is NIOThreadPoolError.UnsupportedOperation {
+                    // ignore, singleton thread pool throws this error on shutdown attempts
```
MahdiBM, line 101: a trace or debug log perhaps?

Sources/Vapor/Core/Core.swift
```diff
@@ -95,16 +95,24 @@ extension Application {
...
+                }
             }
         }
         
         struct AsyncLifecycleHandler: Vapor.LifecycleHandler {
             func shutdownAsync(_ application: Application) async {
                 do {
                     try await application.threadPool.shutdownGracefully()
+                } catch is NIOThreadPoolError.UnsupportedOperation {
+                    // ignore, singleton thread pool throws this error on shutdown attempts
```
MahdiBM, line 113: a trace or debug log perhaps?

Sources/Vapor/Core/Core.swift
gwynne: There's no point - the user can't do anything about it and an "unsupported operation" error is harmless here. It'll just confuse users without adding any value.

Sources/Vapor/Core/Core.swift
```diff
@@ -95,16 +95,24 @@ extension Application {
...
+                    application.logger.debug("Failed to shutdown thread pool", metadata: ["error": "\(error)"])
+                }
             }
         }
         
         struct AsyncLifecycleHandler: Vapor.LifecycleHandler {
             func shutdownAsync(_ application: Application) async {
                 do {
                     try await application.threadPool.shutdownGracefully()
+                } catch is NIOThreadPoolError.UnsupportedOperation {
```
0xTim, line 112: Is this documented anywhere?

gwynne: Not that I know of, other than the source code. But logging this error (which would appear every time an app is shut down) would just unnecessarily confuse/worry users IMO.

gwynne: It's also worth noting that this is the _only_ error that shutting down a thread pool can throw, and even more than that, shutting down a singleton thread pool is the only time it's ever thrown.

Sources/Vapor/Core/Core.swift
MahdiBM: Maybe it becomes harmful in the future 🤷‍♂️ worth a trace IMO, can't hurt

gwynne: What do you imagine as a harmful future for "shutting down the thread pool is an unsupported operation"? 🤨 I mean, at the point of this call we're already shutting down the whole server anyway, I don't particularly think it even _matters_ whether it fails in the first place, much less being worthy of note.

gwynne: I really think logging here, even at trace level, is just going to confuse users to no purpose. It's an expected failure mode (and not even a real failure), why treat it as anything else?

MahdiBM: It's not a hill I would die on but IMO logging it better just in case something goes wrong, even if it's in NIO.

https://www.swift.org/documentation/server/guides/libraries/log-levels.html

> trace is the finest log level, and end-users of a library will not usually use it unless debugging very specific issues. You should consider it as a way for library developers to “log everything we could possibly need to diagnose a hard to reproduce bug.”

0xTim (approved):
I'd prefer to link to the code that explains what the error we're swallowing is but otherwise LGTM

Sources/Vapor/Core/Core.swift
```diff
@@ -95,16 +101,24 @@ extension Application {
 
         struct LifecycleHandler: Vapor.LifecycleHandler {
             func shutdown(_ application: Application) {
-                try! application.threadPool.syncShutdownGracefully()
+                do {
+                    try application.threadPool.syncShutdownGracefully()
```
0xTim, line 105: In fact, do we even need to shut down the thread pool?

gwynne: Non-singleton thread pools assert on deinit if not shut down. Even with the default being the singleton, we still need to shut down in case the user has configured their own non-singleton thread pool.

gwynne:
> I'd prefer to link to the code that explains what the error we're swallowing is but otherwise LGTM

Linked 👍

Sources/Vapor/Core/Core.swift
0xTim: Right but if they pass in their own thread pool then they're responsible for it right?

gwynne: In theory yes, but they haven't actually been so up to this point, so changing the behavior now would break existing code.

PR #3304, opened 2025-02-14
Remove throws requirement from RouteCollection.boot method

bwdmr:
The commit removes the unnecessary throws requirement from the RouteCollection.boot method since route registration operations typically don't throw errors in practice.


gwynne:
This would be source-breaking because it changes the signature of existing public API.

bwdmr:
The proposed PR removes the throws keyword from the boot(routes:) method in the RouteCollection protocol and its implementation. This change is technically sound and improves the codebase.

Based on the provided tests, the implementation of the boot(routes:) method doesn't actually throw errors in practice. The test cases demonstrate that the method functions correctly. By removing throws from a method that doesn't actually throw errors, we eliminate this overhead in what could be a frequently called method during application setup.

Please reconsider.

0xTim:
We cannot reconsider. It is part of the public API. Just because you removed it from the code, doesn't mean that anyone else implementing a `RouteCollection` doesn't throw an error in `boot()`. There are plenty of people who throw errors in their route collections, myself included.

bwdmr:
thank you

PR #3306, opened 2025-02-24
Draft: Add ServiceIdentifier for application tracking

bwdmr:
This commit introduces a new ServiceIdentifier system that allows applications to be uniquely identified and tracked. The implementation includes:

- ServiceIdentifier actor that stores app type, label, and version
- Application extension to store and retrieve identifiers
- Storage mechanism using the Vapor application storage system
- Custom string representation for easy debugging
- Comprehensive test suite validating functionality

The system prevents duplicate registrations of the same app type and provides async accessor methods to maintain thread safety.

### DIFF
--- a/Sources/Vapor/Services/ServiceIdentifier+Service.swift
+++ b/Sources/Vapor/Services/ServiceIdentifier+Service.swift
@@ -6,11 +6,11 @@ extension ObjectIdentifier: @retroactive CustomStringConvertible {
 }
 
 extension Application {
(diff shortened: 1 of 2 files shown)


Naveen-C-Ramachandrappa (approved):

0xTim:
I don't think this makes sense to add to Vapor. I don't see the benefit it would give for wider Vapor uses and can be added as a separate package if needed

PR #3310, opened 2025-02-25
Decode percent-encoded path segments to fix Unicode route matching (#3309)

### LINKED ISSUE #3309: Unable to route requests containing Unicode characters in path segments (returns Not Found)
### Describe the issue

Using Unicode characters (e.g., emojis or Japanese text) in the path segment causes the route to return NotFound instead of the expected response.

### Vapor version

4.113.2

macOS 15.3.1

### Swift version

Swift Package Manager - Swift 6.0.3

### Steps to reproduce

1. Create a new Vapor project (or open an existing one).  
2. Define routes that include Unicode characters (for example, emojis or Japanese text) in the path segments. For instance:  
   ```swift
   app.get("Good👍") { req in
       return "👍"
   }
   app.get("ようこそ世界へ") { req in
       return "おめでとう"
   }
   ```  
3. Run the application in the usual manner (e.g., `swift run`, or using Xcode).  
4. Send a request to these routes (e.g., `GET /Good👍` or `GET /ようこそ世界へ`).  
5. Observe that the response status is `NotFound` instead of returning the expected body (e.g., `"👍"` or `"おめでとう"`).

### Outcome

**Observed (Incorrect) Behavior**  
When sending a request to any route that includes Unicode characters (such as `/Good👍` or `/ようこそ世界へ`), Vapor returns a `NotFound` status code. This indicates that the router fails to match these paths to the defined routes.

**Expected Behavior**  
Vapor should correctly match these routes containing Unicode path segments and return the intended response. For example, a request to `GET /Good👍` should return a `200 OK` with `"👍"` in the response body, and a request to `GET /ようこそ世界へ` should return `200 OK` with `"おめでとう"`.

### Additional notes

_No response_

lemo-nade-room:
**Summary**

This pull request addresses an issue where routes containing Unicode characters (such as emojis or Japanese text) always return a `NotFound` response. The root cause was that incoming path segments are percent-encoded, causing mismatches with the stored routes. By decoding any segments that contain `%` in `DefaultResponder.getRoute()`, we ensure that the router can correctly match Unicode paths.

**Changes**

- Modified `DefaultResponder.getRoute(for:)` to decode percent-encoded segments before routing.
- Added a new `testUnicodePath` test case in `AsyncRouteTests` to verify that routes with emojis and other multi-byte characters can now be matched correctly.

**Reasoning**

- Attempting to register already-encoded routes was problematic, as different HTTP clients may use varying cases (uppercase vs lowercase) for percent-encoded characters.
- Decoding on-the-fly in the responder is more robust and ensures consistent matching regardless of the client’s encoding style.

**Testing**

- A new test `testUnicodePath` has been added. It defines routes with emojis and Japanese text and verifies that they return the expected response. All tests pass locally.

**Issue Reference**

- Resolves #3309

Thank you for reviewing this PR! If you have any questions or suggestions, please let me know.

### DIFF
--- a/Sources/Vapor/Responder/DefaultResponder.swift
+++ b/Sources/Vapor/Responder/DefaultResponder.swift
@@ -79,7 +79,13 @@ internal struct DefaultResponder: Responder {
         let pathComponents = request.url.path
             .split(separator: "/")
             .map(String.init)
-        
+            .map { path in
+                guard path.contains("%") else {
+                    return path
+                }
+                return path.removingPercentEncoding ?? path
+            }
+
         // If it's a HEAD request and a HEAD route exists, return that route...
         if request.method == .HEAD, let route = self.router.route(
             path: [HTTPMethod.HEAD.rawValue] + pathComponents,
--- a/Tests/VaporTests/AsyncRouteTests.swift
+++ b/Tests/VaporTests/AsyncRouteTests.swift
@@ -401,6 +401,29 @@ final class AsyncRouteTests: XCTestCase {
             XCTAssertEqual(res.body.string, "foopbarp")
         }
     }
+
+    func testUnicodePath() throws {
+        app.get("Good👍") { req in
+            "👍"
+        }
+        app.get("ようこそ世界へ") { req in
+            "おめでとう"
+        }
+        app.get("ascii", "🙆‍♂️") { req in
+            "🙅‍♂️"
+        }
+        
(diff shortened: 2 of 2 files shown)


0xTim (approved):
Thanks. I think this makes sense for international users

0xTim:
@lemo-nade-room looks like we might need to dig in more to fix this on 5.9/5.10

PR #3312, opened 2025-03-04
Fix thread pool size documentation comment

MrMage:
Let me know if I've misread the code and the value of 64 is actually correct.


gwynne (approved):
Nope, you have it right! I failed to update the comment when I switched to using the singleton thread pool as the default - nice catch!

gwynne:
Come to think of it, the value `64` has been wrong for a much longer time than that, actually! It's been `System.coreCount` ever since #3092, and the comment didn't get updated then either 😅

PR #3314, opened 2025-03-09
Added a cache policy for file middleware

dimitribouniol:
Currently, if a vapor app uses `FileMiddleware` to deliver static resources, it would have no control over the cache policy used by the browser for those files. This adds a configurable cache policy interface with reasonable defaults to better control when and how often a browser re-requests files (for instance, disabling caches during development).

### DIFF
--- a/Sources/Vapor/Middleware/FileMiddleware.swift
+++ b/Sources/Vapor/Middleware/FileMiddleware.swift
@@ -11,6 +11,7 @@ public final class FileMiddleware: AsyncMiddleware {
     private let defaultFile: String?
     private let directoryAction: DirectoryAction
     private let advancedETagComparison: Bool
+    private let cachePolicy: CachePolicy
 
     public struct BundleSetupError: Equatable, Error {
         
@@ -41,11 +42,19 @@ public final class FileMiddleware: AsyncMiddleware {
     ///     an absolute path from the public directory root. If `nil`, no default files are served.
     ///     - directoryAction: Determines the action to take when the request doesn't have a trailing slash but matches a directory.
     ///     - advancedETagComparison: The method used when ETags are generated. If true, a byte-by-byte hash is created (and cached), otherwise a simple comparison based on the file's last modified date and size.
-    public init(publicDirectory: String, defaultFile: String? = nil, directoryAction: DirectoryAction = .none, advancedETagComparison: Bool = false) {
+    ///     - cacheControl: Specifies the browser's cache policy that should be used for all files. Defaults to the browser's default behavior ``CachePolicy/browserDefault``.
+    public init(
+        publicDirectory: String,
+        defaultFile: String? = nil,
+        directoryAction: DirectoryAction = .none,
+        advancedETagComparison: Bool = false,
+        cachePolicy: CachePolicy = .browserDefault
+    ) {
         self.publicDirectory = publicDirectory.addTrailingSlash()
         self.defaultFile = defaultFile
         self.directoryAction = directoryAction
         self.advancedETagComparison = advancedETagComparison
+        self.cachePolicy = cachePolicy
     }
     
     public func respond(to request: Request, chainingTo next: any AsyncResponder) async throws -> Response {
@@ -80,7 +89,10 @@ public final class FileMiddleware: AsyncMiddleware {
                         
                         if try await FileSystem.shared.info(forFileAt: .init(absPath)) != nil {
                             // If the default file exists, stream it
-                            return try await request.fileio.asyncStreamFile(at: absPath, advancedETagComparison: advancedETagComparison)
+                            return try await request
+                                .fileio
+                                .asyncStreamFile(at: absPath, advancedETagComparison: advancedETagComparison)
+                                .cachePolicy(cachePolicy)
                         }
                     }
                 } else {
@@ -92,7 +104,10 @@ public final class FileMiddleware: AsyncMiddleware {
                 }
             } else {
                 // file exists, stream it
-                return try await request.fileio.asyncStreamFile(at: absPath, advancedETagComparison: advancedETagComparison)
+                return try await request
+                    .fileio
+                    .asyncStreamFile(at: absPath, advancedETagComparison: advancedETagComparison)
+                    .cachePolicy(cachePolicy)
             }
         }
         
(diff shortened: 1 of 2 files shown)


Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -41,11 +42,19 @@ public final class FileMiddleware: AsyncMiddleware {
...
     ///     - directoryAction: Determines the action to take when the request doesn't have a trailing slash but matches a directory.
     ///     - advancedETagComparison: The method used when ETags are generated. If true, a byte-by-byte hash is created (and cached), otherwise a simple comparison based on the file's last modified date and size.
-    public init(publicDirectory: String, defaultFile: String? = nil, directoryAction: DirectoryAction = .none, advancedETagComparison: Bool = false) {
+    ///     - cacheControl: Specifies the browser's cache policy that should be used for all files. Defaults to the browser's default behavior ``CachePolicy/browserDefault``.
+    public init(
+        publicDirectory: String,
+        defaultFile: String? = nil,
+        directoryAction: DirectoryAction = .none,
+        advancedETagComparison: Bool = false,
+        cachePolicy: CachePolicy = .browserDefault
```
dimitribouniol, line 51: Went with the current behavior (browser default), but we may want to consider a better default, such as `.cacheUpToDuration(.minutes(5))` to make it explicit for new projects.

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -171,3 +193,53 @@ fileprivate extension String {
...
+        }
+    }
+}
+
+extension Response {
+    /// Update the response with a cache policy.
+    /// - Parameter policy: The cache policy to use.
+    /// - Returns: The same response as the receiver.
+    @discardableResult
+    public func cachePolicy(_ policy: FileMiddleware.CachePolicy) -> Response {
```
dimitribouniol, line 236: I opted to let this be public, but we can mark it internal if we aren't ready to explore this pattern.

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -41,11 +42,19 @@ public final class FileMiddleware: AsyncMiddleware {
...
     ///     - advancedETagComparison: The method used when ETags are generated. If true, a byte-by-byte hash is created (and cached), otherwise a simple comparison based on the file's last modified date and size.
-    public init(publicDirectory: String, defaultFile: String? = nil, directoryAction: DirectoryAction = .none, advancedETagComparison: Bool = false) {
+    ///     - cacheControl: Specifies the browser's cache policy that should be used for all files. Defaults to the browser's default behavior ``CachePolicy/browserDefault``.
+    public init(
+        publicDirectory: String,
+        defaultFile: String? = nil,
+        directoryAction: DirectoryAction = .none,
+        advancedETagComparison: Bool = false,
+        cachePolicy: CachePolicy = .browserDefault
+    ) {
```
dimitribouniol, line 52: This is technically API breaking, but only if people use the init as a key path. I can duplicate the methods to work around it, but I've also seen in past PRs we've chosen not to do this, so let me know if we even should.

Sources/Vapor/Middleware/FileMiddleware.swift
0xTim: I think leaving it as is is probably the best idea in case anyone is relying on it

Sources/Vapor/Middleware/FileMiddleware.swift
0xTim: Yeah I think this is fine as it is

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -171,3 +193,53 @@ fileprivate extension String {
...
+        /// The browser will always ask before requesting the full file.
+        ///
+        /// This can be used if the files served change very often, or in development so any change to a file is immediately reflected.
+        public static let noCache = CachePolicy(cacheControlHeader: .init(noCache: true))
+        
+        /// The browser will cache the file for the specified duration.
+        ///
+        /// A typical cache duration may be 5 minutes, for instance: `.cacheUpToDuration(.minutes(5))`
+        public static func cacheUpToDuration(_ duration: TimeAmount) -> CachePolicy {
+            CachePolicy(cacheControlHeader: .init(maxAge: Int(duration.nanoseconds/1_000_000_000)), ageHeader: 0)
```
0xTim, line 217: Why not just use seconds here instead of nanoseconds and then dividing them

Sources/Vapor/Middleware/FileMiddleware.swift
```diff
@@ -171,3 +193,53 @@ fileprivate extension String {
...
+        }
+        
+        /// A custom cache control policy that should be used for all files.
+        /// - Parameters:
+        ///   - cacheControlHeader: The `Cache-Control` header to use. If none is specified, any previous cache control header will be cleared.
+        ///   - ageHeader: The `Age` header to use. If none is specified, any previous age header will be cleared.
```
0xTim, line 223: Would be good to clarify what the number here means - seconds, minutes etc

Sources/Vapor/Middleware/FileMiddleware.swift
0xTim: I think keep this internal for now

0xTim (changes requested):

Sources/Vapor/Middleware/FileMiddleware.swift
dimitribouniol: If only 😅 
image

Neither Swift's Duration (which is unavailable on the lowest supported vapor deploy for appleOSs), nor NIOs TimeAmount (which provides sane initializers for minutes and hours) seem to support accessing seconds directly (leaving aside `.components.seconds` which feels like an implementation detail)… unless there's an API I'm missing?

Sources/Vapor/Middleware/FileMiddleware.swift
dimitribouniol: Made it internal

Sources/Vapor/Middleware/FileMiddleware.swift
dimitribouniol: Fixed!

Sources/Vapor/Middleware/FileMiddleware.swift
0xTim: Ohhh we're going that way, yeah fair enough

0xTim (approved):

PR #3318, opened 2025-03-30
Add `Partitioned` field to Cookie struct

tikiram:
Cookie struct now has the `Partitioned` attribute

### DIFF
--- a/Sources/Vapor/HTTP/Headers/HTTPCookies.swift
+++ b/Sources/Vapor/HTTP/Headers/HTTPCookies.swift
@@ -88,6 +88,8 @@ struct HTTPSetCookie {
                     return nil
                 }
                 self.value.sameSite = HTTPCookies.SameSitePolicy(rawValue: .init(parameter))
+            case "partitioned":
+              self.value.isPartitioned = true
             default:
                 return nil
             }
@@ -143,6 +145,9 @@ public struct HTTPCookies: ExpressibleByDictionaryLiteral, Sendable {
         ///
         /// This restriction mitigates attacks such as cross-site request forgery (XSRF).
         public var sameSite: SameSitePolicy?
+      
+        /// The cookie can only be read within the context of the top-level site it was set on.
+        public var isPartitioned: Bool
         
         // MARK: Init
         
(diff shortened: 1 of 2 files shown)


ProSeFraudCatcher (approved):

Sources/Vapor/HTTP/Headers/HTTPCookies.swift
```diff
@@ -180,6 +187,7 @@ public struct HTTPCookies: ExpressibleByDictionaryLiteral, Sendable {
             self.isSecure = isSecure || forceSecure
             self.isHTTPOnly = isHTTPOnly
             self.sameSite = sameSite
+            self.isPartitioned = isPartitioned
```
0xTim, line 190: We should add an assertion here to ensure that `secure` is set as it's a requirement [in the RFC](https://datatracker.ietf.org/doc/html/draft-cutler-httpbis-partitioned-cookies#name-partitioned-requires-secure)

0xTim (changes requested):
We need to assert on the secure state

Do we have a link anywhere to where the spec has been approved?

Sources/Vapor/HTTP/Headers/HTTPCookies.swift
tikiram: Sounds good, I can add an `assert` statement, the `isPartitioned` field is a public `var`, so I guess I have to place the `assert` into a `willSet` block for the `isPartitioned` field

```swift
let cookie = ...
cookie.isPartitioned = true // <--- assert 
```

By the way, I didn't find any use of `willSet` in this project. Is it okay to use it in this case?

0xTim: Sorry for the delay! Yeah that's fine to use it in the `willset` for now

tikiram: No worries, I added an 'assert' statement to the `isPartititoned` property, I wanted to add a test for this case, but I didn't find an easy way to check this `assert` statement with XCTest (`XCTAssertThrowsError` only works with throwables), do you know how this is handle in this project?

PR #3323, opened 2025-04-14
Range eof

keniwhat:
Content range requests past EOF will fail witth the current version of vapor.   It is common practice to just return the remaining bytes rather than throw a bad request error. This helps  existing imperfect clients, still out there in the wild to continue to function.

This change captures an attempt to read past EOF (eg  a range request of 10-200, when there are only 180 bytes in the file), and returns a response of "partial content" and the remaining bytes.  It replaces code that returned a "bad request" error.

The existing tests have been updated to reflect this change and test for a response of type "partial content". 
In addition two new asserts have been added to a test:
- The response header content range meets expectation
- The response body byte count meets expectation

As well as passing all tests, the code has been tested with a real life test case.

NOTE: The test 'AsyncRouteTests.testGH2716()' does not not pass. However, it does not pass on the vapor/main branch from which this has been forked.

### DIFF
--- a/Sources/Vapor/HTTP/Headers/HTTPHeaders+ContentRange.swift
+++ b/Sources/Vapor/HTTP/Headers/HTTPHeaders+ContentRange.swift
@@ -263,11 +263,12 @@ extension HTTPHeaders.Range.Value {
             }
             return .withinWithLimit(start: (limit - end), end: limit - 1, limit: limit)
         case .within(let start, let end):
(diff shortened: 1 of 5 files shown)


Naveen-C-Ramachandrappa (approved):

0xTim:
Thanks for this! If you want to point the base branch to `vapor4` we can look at adding it there. There's a lot of duplicated logic in the range checking which we can merge together (see #3554) if you want to pick this up again. Sorry for the delay!

PR #3324, opened 2025-04-15
Adds OTel metrics middleware

NeedleInAJayStack:
This adds a middleware that implements the [OTel standard http metrics](https://opentelemetry.io/docs/specs/semconv/http/http-metrics).

There is significant overlap with the metrics in [DefaultResponder](https://github.com/vapor/vapor/blob/main/Sources/Vapor/Responder/DefaultResponder.swift#L101), and I'm not sure what you guys prefer, whether it's deprecation of the DefaultResponder metrics, just sticking with those and not accepting these changes, or supporting both independently. I'm interested to hear your thoughts!

Thanks!

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -146,6 +146,7 @@ let package = Package(
                 .target(name: "XCTVapor"),
                 .target(name: "VaporTesting"),
                 .target(name: "Vapor"),
+                .product(name: "MetricsTestKit", package: "swift-metrics"),
             ],
             resources: [
                 .copy("Utilities/foo.txt"),
--- a/Sources/Vapor/Middleware/MetricsMiddleware.swift
+++ b/Sources/Vapor/Middleware/MetricsMiddleware.swift
@@ -1,0 +1,125 @@
+import Metrics
+
+/// Creates metrics for every request
+///
+/// See https://opentelemetry.io/docs/specs/semconv/http/http-metrics/
+public final class MetricsMiddleware: AsyncMiddleware {
+    
+    public init() {}
+    
+    public func respond(to request: Request, chainingTo next: any AsyncResponder) async throws -> Response {
+        let start = DispatchTime.now().uptimeNanoseconds
+        
+        // Attributes: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-semantic-conventions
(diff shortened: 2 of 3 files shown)


0xTim:
I think at this point we'll probably stick with the built in stuff in the responder otherwise we risk breaking apps that rely on it and we don't want to duplicate all the metrics. Are there any metrics that we're missing with the inbuilt stuff for Otel? I thought we had them all.

_However_, this would be a good addition for Vapor 5! You could update the PR to point to the `vapor5` branch if you fancy it

NeedleInAJayStack:
> I think at this point we'll probably stick with the built in stuff in the responder otherwise we risk breaking apps that rely on it and we don't want to duplicate all the metrics.

That's totally reasonable.

> Are there any metrics that we're missing with the inbuilt stuff for Otel? I thought we had them all.

Nope, the only required one is `http.server.request.duration`, but we've already got that covered with `http_request_duration_seconds`. The other ones I added are marked optional in OTel.

> However, this would be a good addition for Vapor 5! You could update the PR to point to the vapor5 branch if you fancy it

Oh awesome! Yeah, I'll do that. Thanks!

0xTim:
@NeedleInAJayStack want to take a look at the conflicts? We can also remove the metrics from the default responder

NeedleInAJayStack:
Sure, I rebased on the tip of `vapor-5`, resolved the compilation errors, and removed the metrics in the default resolver. The failing tests appear to be unrelated to this PR's changes.

Maybe worth noting since I had forgotten - The default resolver's `http_requests_total` metric is effectively replaced by OTel's `http.server.request.duration_count`.

Let me know what you think!

0xTim:
@NeedleInAJayStack sorry for the churn, could you update again? I've updated the test files to use a proper task local injected metrics system that will allow the tests to pass in parallel

0xTim:
As for changing the metrics, I think that's fine. We can mark it in the release notes and people can duplicate the middleware and use their own labels if they want

NeedleInAJayStack:
No worries. I've updated it. I also added a few missing methods to the `TaskLocalMetricsSystemWrapper` to support `Meter` and `FloatingPointCounter` types.

PR #3332, opened 2025-05-14
Improve `VaporTesting`'s `withApp` method

fpseverino:
**These changes are now available in [4.115.0](https://github.com/vapor/vapor/releases/tag/4.115.0)**

This PR makes the `withApp` helper method work as advertised, allowing an optional configuration method for the `Application` to be passed.

Also, marks it with `@discardableResult` to avoid unnecessary warnings.

### DIFF
--- a/Sources/VaporTesting/withApp.swift
+++ b/Sources/VaporTesting/withApp.swift
@@ -17,7 +17,7 @@ import Vapor
 /// ```
 ///
 /// - Parameters:
-///   - configure: The method where you should register services like routes, databases, providers, and more.
+///   - configure: A closure where you can register routes, databases, providers, and more.
 ///   - test: The method where you can perform your tests with the configured application.
 @discardableResult
 public func withApp<T>(
--- a/Sources/VaporTesting/withApp.swift
+++ b/Sources/VaporTesting/withApp.swift
@@ -27,9 +27,7 @@ public func withApp<T>(
     let app = try await Application.make(.testing)
     let result: T
     do {
-        if let configure {
-            try await configure(app)
-        }
+        try await configure?(app)
         result = try await test(app)
     } catch {
         try? await app.asyncShutdown()
--- a/Sources/VaporTesting/withApp.swift
+++ b/Sources/VaporTesting/withApp.swift
@@ -7,7 +7,7 @@ import Vapor
 /// ```swift
 /// @Test
 /// func helloWorld() async throws {
-///     try await withApp(/* whatever set up needed*/) { app in
+///     try await withApp(configure: configure) { app in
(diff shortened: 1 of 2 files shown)


Sources/VaporTesting/withApp.swift
```diff
@@ -7,19 +7,29 @@ import Vapor
...
 ///         })
 ///     }
 /// }
 /// ```
-public func withApp<T>(_ block: (Application) async throws -> T) async throws -> T {
+///
+/// - Parameters:
+///   - configure: The method where you should register services like routes, databases, providers, and more.
+///   - test: The method where you can perform your tests with the configured application.
+@discardableResult public func withApp<T>(
```
gwynne, line 22:
```suggestion
@discardableResult
public func withApp<T>(
```

Sources/VaporTesting/withApp.swift
```diff
@@ -7,19 +7,29 @@ import Vapor
...
+    configure: ((Application) async throws -> Void)? = nil,
+    _ test: (Application) async throws -> T
+) async throws -> T {
     let app = try await Application.make(.testing)
     let result: T
     do {
-        result = try await block(app)
+        if let configure {
+            try await configure(app)
+        }
```
gwynne, line 31:
```suggestion
        try await configure?(app)
```

Sources/VaporTesting/withApp.swift
```diff
@@ -7,19 +7,29 @@ import Vapor
...
 ///             #expect(res.status == .ok)
 ///             #expect(res.body.string == "Hello, world!")
 ///         })
 ///     }
 /// }
 /// ```
-public func withApp<T>(_ block: (Application) async throws -> T) async throws -> T {
+///
+/// - Parameters:
+///   - configure: The method where you should register services like routes, databases, providers, and more.
```
gwynne, line 20:
```suggestion
///   - configure: A closure where you can register routes, databases, providers, and more.
```

Sources/VaporTesting/withApp.swift
```diff
@@ -7,19 +7,29 @@ import Vapor
...
+///   - test: The method where you can perform your tests with the configured application.
```
gwynne, line 21:
```suggestion
///   - test: A closure which performs your actual test with the configured application.
```

gwynne (changes requested):
Multiple little nits.

Sources/VaporTesting/withApp.swift
fpseverino: I didn't even know this was possible

gwynne: TYL.

PR #3338, opened 2025-06-03
test: fix `configuration` argument

MojtabaHs:
Fixed all occurrences of `configuration` throughout the code. (This was only present in test files.)

Although I initially added a soft-deprecation shim to preserve backward compatibility, I ultimately removed it because:
1. The affected code only exists in the v5 branch, which has not yet been released.
2. ~~The typo appears exclusively in test files, not in the public API of the library itself.~~
3. Even if it is used somehow by the outside world, with a new major version, breaking changes are expected and acceptable.

Given these points, the shim was unnecessary and has been omitted.

Please let me know if this needs anymore work from me.
Thanks

### DIFF
--- a/Sources/VaporTesting/TestingHTTPRequest.swift
+++ b/Sources/VaporTesting/TestingHTTPRequest.swift
@@ -10,12 +10,17 @@ public struct TestingHTTPRequest: Sendable {
     public var body: ByteBuffer
     public var contentConfiguration: ContentConfiguration
 
-    public init(method: HTTPRequest.Method, url: URI, headers: HTTPFields, body: ByteBuffer, contentConfigurtion: ContentConfiguration) {
+    public init(method: HTTPRequest.Method, url: URI, headers: HTTPFields, body: ByteBuffer, contentConfiguration: ContentConfiguration) {
         self.method = method
         self.url = url
         self.headers = headers
         self.body = body
-        self.contentConfiguration = contentConfigurtion
+        self.contentConfiguration = contentConfiguration
+    }
+
+    @available(*, deprecated, renamed: "init(method:url:headers:body:contentConfiguration:)")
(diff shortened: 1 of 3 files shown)


0xTim:
Unfortunately VaporTesting is a public API module so we'll need the old function deprecated so as not to break the API

MojtabaHs:
No problem @0xTim . I’ve reverted the last commit to restore the deprecated initializer and ensure the change doesn’t introduce a breaking update.

MahdiBM:
(repeating the other comment: https://github.com/vapor/vapor/pull/3337#discussion_r2128320376)
I think @0xTim and I missed that this PR is targeting the `vapor-5` branch. I think this is fine to be merged into `main` for Vapor 4 users to receive this patch, considering the deprecation is reasonable and is easy to fix.

Sources/VaporTesting/TestingHTTPRequest.swift
```diff
@@ -10,12 +10,17 @@ public struct TestingHTTPRequest: Sendable {
...
         self.url = url
         self.headers = headers
         self.body = body
-        self.contentConfiguration = contentConfigurtion
+        self.contentConfiguration = contentConfiguration
+    }
+
+    @available(*, deprecated, renamed: "init(method:url:headers:body:contentConfiguration:)")
+    public init(method: HTTPRequest.Method, url: URI, headers: HTTPFields, body: ByteBuffer, contentConfigurtion: ContentConfiguration) {
+        self.init(method: method, url: url, headers: headers, body: body, contentConfiguration: contentConfigurtion)
```
0xTim, line 23: Since this is targeting the Vapor 5 branch we don't need this here

0xTim (changes requested):

0xTim (approved):
Thanks!

PR #3340, opened 2025-06-12
Add dynamic origin strategy & correct Vary: Origin handling in CORS middleware

### LINKED ISSUE #2988: Allow wildcard origins in CORS middleware
**Is your feature request related to a problem? Please describe.**

We have one server that deals with several data silos, one per customer, and this is done via subdomain, e.g. ACME.my.app and PEAK.my.app, etc. As and when new customers sign up, we create new DNS entries, but have to recompile our server code and add yet another origin into the CORS configuration `.any([...])`. 

In addition to `case any([String])` I would like `case anyWildcard([NSRegularExpression])` so that I can code what amounts to "https://*.my.app" via `NSRegularExpression("https://.*\\.my\\.app")`
The caller creates the regular expressions once only during configuration and they are cached inside the configuration object. 

Matching an incoming origin header to a short list of pre compiled regular expressions at runtime should be just a little more expensive than matching that origin in long list of strings. Regardless of the runtime performance, this enhancement will decouple the creation of a new customer from having to recompile the server code, even though we have automated this.

**Describe alternatives you've considered**

Scanning all the strings in the `case .any(...)` for wildcard characters "*" and "?" then converting those strings to NSRegularExpressions on the fly would be too much extra work at runtime, on every incoming request to the server. Plus, this may break someone's existing code in the wild.

**Additional context**

Please see https://github.com/shelton-dev/vapor/pull/1 for an initial implementation.

ajevans99:
This PR makes two changes to CORSMiddleware:

1. Adds `.dynamic`, a new enum case that accepts`@Sendable (Request) -> String)` letting callers compute `Access-Control-Allow-Origin` per request (resolves #2988). The closure can use any matching logic (regex, prefix, etc.); the middleware itself stays agnostic.
2. Sets `Vary: Origin` for `.dynamic`, `.originBased`, and `.any`. Previously, only `.originBased` included this header. I believe `any` (and the new `dynamic` case) should also have this logic based on [RFC 7231 § 4.1](https://datatracker.ietf.org/doc/html/rfc7234#section-4.1) and [RFC 7231 § 7.1.4](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.4)
3. Adds tests for both.

### DIFF
--- a/Tests/VaporTests/MiddlewareTests.swift
+++ b/Tests/VaporTests/MiddlewareTests.swift
@@ -137,7 +137,7 @@ struct MiddlewareTests {
             try await app.testing().test(.get, "/order", headers: [.origin: "http://example-123.com"]) { res in
                 #expect(res.status == .ok)
                 #expect(res.body.string == "done")
-                #expect(res.headers[values: .vary] == [])
+                #expect(res.headers[values: .vary] == ["origin"])
                 #expect(res.headers[values: .accessControlAllowOrigin] == ["http://example-123.com"])
                 #expect(res.headers[values: .accessControlAllowHeaders] == [""])
             }


0xTim:
Unfortunately we can't add this to Vapor 4 as this would be a breaking change (adding a new case to a public enum). We can add it to Vapor 5 though

ajevans99:
> Unfortunately we can't add this to Vapor 4 as this would be a breaking change (adding a new case to a public enum). We can add it to Vapor 5 though

Rebased into Vapor 5 branch :)

ajevans99:
Gently ping @0xTim. I think this would be a valuable addition to Vapor 5.

PR #3344, opened 2025-07-13
 feat: complete Vapor 5 core implementation with async/await and HTTPTypes

1amageek:
## Summary
  - Complete safe request body handling without force unwraps
  - Implement ContentContainer async subscript operators for keypath-based access
  - Restore HTTP/HTTPS scheme detection in server configuration
  - Fix HTTP method case compatibility with HTTPTypes library
  - Add graceful fallback for dual body system during migration

  ## Changes
  - **Request.swift**: Safe newBody property with graceful fallback for testing/legacy compatibility
  - **ContentContainer.swift**: Async subscript operators enabling `req.content[at: "user", "name"]`
  syntax
  - **Application.swift**: Port-based HTTP/HTTPS scheme detection (443, 8443 → HTTPS)
  - **VaporTesting.swift**: HTTP method case fixes (.GET → .get) for HTTPTypes compatibility

### DIFF
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -193,9 +193,17 @@ public final class Application: Sendable {
 
         /// A human-readable description of the configured address. Used in log messages when starting server.
         var addressDescription: String {
-            #warning("Bring back")
-//            let scheme = tlsConfiguration == nil ? "http" : "https"
-            let scheme = "https"
+            // Use HTTPS for standard secure ports, HTTP otherwise
+            let scheme: String
+            switch address {
+            case .hostname(_, let port):
+                // Port 443 (HTTPS), 8443 (dev HTTPS) use https, others use http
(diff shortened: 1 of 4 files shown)


Sources/Vapor/Application.swift
```diff
@@ -193,9 +193,17 @@ public final class Application: Sendable {
 
         /// A human-readable description of the configured address. Used in log messages when starting server.
         var addressDescription: String {
-            #warning("Bring back")
-//            let scheme = tlsConfiguration == nil ? "http" : "https"
-            let scheme = "https"
+            // Use HTTPS for standard secure ports, HTTP otherwise
```
0xTim, line 196: We should rely on whether we have a `tlsConfiguration` set or not, rather than the port

0xTim (changes requested):
Thanks for this! Can we split this into separate PRs, only for the compile fixes and one for the migrations

PR #3345, opened 2025-07-14
Add Sendable annotations to the OTP types

0xTim:
Add some missing Sendable annontations to the OTP types so `@preconcurrency import` is not required when using the types across isolation boundaries


gwynne (approved):

vanvoorden:
https://forums.swift.org/t/build-failure-from-hotp-and-totp-checking-sendable/83027

BTW… I am seeing some build failures now that could be related to this change.

PR #3352, opened 2025-07-28
Expose peer certificate in request handlers

josephnoir:
Additional certificate information can be relevant in mTLS deployments. Exposing the certificate makes this information available per request. The certificate will be nil if the client does not provide one.

Example usage:
```swift
app.get { req async in
    if let cert = req.peerCertificate {
        return "Your certificate is: \(cert.description)."
    } else {
        return "You provided no certificate."
    }
}
```

### DIFF
--- a/Sources/Vapor/Request/Request.swift
+++ b/Sources/Vapor/Request/Request.swift
@@ -1,6 +1,7 @@
 import Foundation
 import NIOCore
 import NIOHTTP1
+import NIOSSL
 import Logging
 import RoutingKit
 import NIOConcurrencyHelpers
@@ -95,6 +96,11 @@ public final class Request: CustomStringConvertible, Sendable {
         }
 
         return self.remoteAddress
+    }
+
+    /// The certificate of the peer. This returns nil if the client did not authenticate with a certificate.
+    public var peerCertificate: NIOSSLCertificate? {
+        return self.requestBox.withLockedValue { $0.peerCertificate }
     }
 
     // MARK: Content
@@ -277,6 +283,7 @@ public final class Request: CustomStringConvertible, Sendable {
         var route: Route?
         var parameters: Parameters
         var byteBufferAllocator: ByteBufferAllocator
+        var peerCertificate: NIOSSLCertificate?
     }
     
     let requestBox: NIOLockedValueBox<RequestBox>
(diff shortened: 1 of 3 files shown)


Tests/VaporTests/ServerTests.swift
```diff
@@ -1273,7 +1273,66 @@ final class ServerTests: XCTestCase, @unchecked Sendable {
...
+        clientConfig.certificateVerification = .none
+        clientConfig.certificateChain = [.certificate(cert)]
+        clientConfig.privateKey = .privateKey(key)
+        app.http.client.configuration.tlsConfiguration = clientConfig
+
+        app.environment.arguments = ["serve"]
+
+        app.get("hello") { req in
+            // The peer certificate should be available ...
+            XCTAssertNotNil(req.peerCertificate)
```
0xTim, line 1312: An `XCTUnwrap` would be better here so it can be used in the next line

Tests/VaporTests/ServerTests.swift
```diff
@@ -1273,7 +1273,66 @@ final class ServerTests: XCTestCase, @unchecked Sendable {
         let a = try app.http.client.shared.execute(request: request).wait()
         XCTAssertEqual(a.body, ByteBuffer(string: "world"))
     }
-    
+
+    func testCanAccessPeerCertificate() throws {
```
0xTim, line 1277: Can we make the test async and rely on the async functions? The sync functions can be problematic because the setup is asynchronous (I know the rest of the tests need migrating but that's another time)

0xTim (changes requested):
Thanks for this! A couple of suggestions for the tests and then we can merge

Tests/VaporTests/ServerTests.swift
josephnoir: Good point!

Tests/VaporTests/ServerTests.swift
0xTim: Same file is fine ta

Sources/Vapor/Request/Request.swift
```diff
@@ -97,6 +98,11 @@ public final class Request: CustomStringConvertible, Sendable {
         return self.remoteAddress
     }
 
+    /// The certificate of the peer. This returns nil if the client did not authenticate with a certificate.
+    public var peerCertificate: NIOSSLCertificate? {
```
FranzBusch, line 102: Is there a reason why we are using `NIOSSLCertificate` here instead of `X509.Certifacte`? The latter seems like the better type for users nowadays or rather `[X509.Certificate]` for the entire chain?

PR #3359, opened 2025-09-04
Allow initializing Vapor.Application with a Logger

czechboy0:
**These changes are now available in [4.116.0](https://github.com/vapor/vapor/releases/tag/4.116.0)**

When using a single Logger in the whole app, it's useful to be able to initialize the Vapor app with one as well, rather than letting it create one internally.

This doesn't change behavior for existing users, just adds a new variant of the `Vapor.Application.make` method that takes a Logger explicitly.

Added unit tests to cover both the default and the new option.


czechboy0:
I suspect the JWT test failure is unrelated?

0xTim:
> I suspect the JWT test failure is unrelated?

Yeah correct, it's a SIWA token that's proving problematic. Not a blocker

PR #3362, opened 2025-09-17
Expose peer certificates in request handlers

josephnoir:
Additional certificate information can be relevant in mTLS deployments. This PR exposes the certificate chain of the peer per request. In contrast to #3352, this includes not just the leaf but the validated peer certificate chain, i.e., the certificates that establish trust of the peer identity from the leaf to (and including) the root certificate.

The validated certificate chain is only available when setting a custom verification callback with metadata. Configuration of this callback is made available in this PR. It also adds a dependency on `swift-certificates` to use `X509.ValidatedCertificateChain` for the certificate chain.

Example usage:

```swift
app.get { req async in
  if let chain = req.peerCertificateChain {
    return "I am trusting you because I trust: \(chain.root.description)"
  } else {
    return "I am not trusting you."
  }
}
```

I suggest close #3352 in favor of this PR.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -35,7 +35,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.81.0"),
 
         // Bindings to OpenSSL-compatible libraries for TLS support in SwiftNIO
-        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.8.0"),
+        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.34.0"),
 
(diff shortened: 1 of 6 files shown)


gwynne:
> @gwynne any thoughts before we merge

Mostly "_Dang_, those are some verbose type/variable names." But that's not a blocker 😅.

josephnoir:
Great 😁 I'll add a commit to address the api breakage.

0xTim:
@josephnoir looks like the tests might need a dependency on ASN1 and an import

josephnoir:
Thanks for checking the problem! I added both.

PR #3364, opened 2025-09-26
Empty responses should not be compressed

sergiocampama:
When returning empty responses, such as `Response(status: .ok)` and the browser has sent `Accept-Encoding: gzip, deflate, br, zstd` headers, Vapor should not compress responses.

When compression is enabled on empty responses, some compression encodings, such as `deflate`, end up returning 8 bytes of data, so the final response that the browser receives has `Content-Length == 8`, but with no body. This ends up erroring with `net::ERR_CONTENT_LENGTH_MISMATCH` on the browser side.


gwynne:
This behavior is correct according to spec. Empty responses should be returned using response code `204 No Content` rather than `200 OK`.

I just went through this same thought process about two weeks ago—that's where I ended up 😅.

PR #3372, opened 2025-10-13
Do not create a session if no cookie was provided

czechboy0:
**These changes are now available in [4.121.1](https://github.com/vapor/vapor/releases/tag/4.121.1)**

## Summary

When using `SessionsMiddleware` and a custom `AsyncSessionAuthenticator` together, we were seeing that even in cases when no cookie is provided in the request, and no error was thrown by e.g. a guard middleware or the handler, we were still seeing a `set-cookie` header in the response, which isn't desired.

This lead to many sessions created in the sessions driver for calls that didn't require authentication, nor did it provide any.

This seemed to be a side effect of the way `AsyncSessionAuthenticator` checked if the session already contained an authenticated value - by using the getter `request.session`, a session was created lazily.

The fix is to prepend a check to `request.hasSession` and only using `request.session...` if one indeed exists.

## Test Plan

A regression test was added which exercises this scenario - and it fails without the fix.


gwynne:
I think this is partially intended behavior, in that `request.session` ought to always exist, but also incorrect behavior, in that a cookie should _not_ be set if 1) no cookie was provided in the request, and 2) the session is still empty when it's time to send the response. As such, I believe the check belongs in `SessionsMiddleware.addCookies(to:for:)`, which does not currently check whether a new session (`session.id` is `nil`)  is empty.

0xTim:
Yep agree with @gwynne here, if you have a global sessions middleware you should see a `set-cookie` on every request. We should add the check there since this is the part doing the actual logic to see whether to set a session

czechboy0:
> but also incorrect behavior, in that a cookie should _not_ be set if 1) no cookie was provided in the request, and 2) the session is still empty when it's time to send the response. As such, I believe the check belongs in `SessionsMiddleware.addCookies(to:for:)`, which does not currently check whether a new session (`session.id` is `nil`) is empty.

Gotcha, so the check in `addCookies` should be expanded with:
1. A check for `session.id`? (This is already used to decide whether to call `updateSession` or `createSession`)
2. A check for `session.data` being empty? (Is that right? Or could that be considered a breaking change, since up until now, it was possible to use an empty session to keep a user signed in, in theory)

Happy to update the logic, but what is the desired behavior?

gwynne:
@czechboy0 My take is that when creating a new session (`session.id` is `nil`), there must be at least one piece of data stored in the session for a session to actually be created. Once the session already exists, it doesn't matter if it's empty or not. @0xTim Does that make sense to you?

0xTim:
Yeah I think if we have a session cookie, then we should always set the session. If there's no session cookie in the request, then we're in a fresh state to decide whether to set the `set-cookie` header. So yeah I think session ID is the right way to go

czechboy0:
The problem is that when you have an `AsyncSessionAuthenticator` in your middleware stack after a `SessionsMiddleware`, even if no `cookie` was provided in the request, the `AsyncSessionAuthenticator` creates an empty session (no explicit ID, no data entries) here (because the `request.session` is lazy and creates an empty session on access): https://github.com/vapor/vapor/blob/25faea95826f7b274129c02286f0483c416d2339/Sources/Vapor/Concurrency/Authentication%2BConcurrency.swift#L103

And then, since the session already exists, the `SessionsMiddleware` asks the session driver to create the session (because no ID is there, but there is a session). At that point, an ID will be created, but still no data is in there (you can see the empty `{}` in e.g. the Fluent-driven database). And since there's a session, `SessionsMiddleware` will set `set-cookie`.

I'm trying to avoid the session being created in the driver at all.

What specifically do you think should change?

czechboy0:
@0xTim @gwynne gentle ping, I'm still hoping for a fix here, as it's polluting the session store

gwynne:
Ummm. Mentally reloading the context here... IIRC what I was thinking needs to happen is that SessionsMiddleware needs to avoid setting the cookie if 1) there is no session ID at all (no existing cookie and no session was ever asked after), or 2) there is no previously set cookie, and there is no session data (no existing cookie and a session was asked after but nothing was done with it).

That being said, arguably the behavior of `[Async]SessionAuthenticator` is also wrong here. It should be checking `request.hasSession` first to avoid the lazy-creation behavior of `request.session`. It's doubly annoying, because it actually _does_ do this in one place, but not in the other (https://github.com/vapor/vapor/blob/3014d7de87d86939f31c5e88b3aa261a46e1d460/Sources/Vapor/Authentication/SessionAuthenticatable.swift#L20 versus https://github.com/vapor/vapor/blob/3014d7de87d86939f31c5e88b3aa261a46e1d460/Sources/Vapor/Authentication/SessionAuthenticatable.swift#L35). I would go so far as to call that an outright bug, in fact. If that alone were fixed, it seems like it would resolve the immediate issue.

At the same time, I also agree that we shouldn't be creating empty sessions. We should respect them if they _already_ exist, but not create them if they don't.

czechboy0:
Thanks @gwynne - completely agreed, that's what this PR tries to fix. Did I miss the mark? 🙂

gwynne:
Wow, I... probably should have actually looked at the diff while I was reloading my mental context there 😅. Looks good to me, unless @0xTim has any objections.

gwynne:
(I do think SessionsMiddleware should also be smarter about this, but fixing the authenticator covers the most critical ground.)

gwynne (changes requested):
Oh, wow, I almost missed this— @czechboy0 Can you also fix this for the non-async `SessionAuthenticator`?

czechboy0:
> Oh, wow, I almost missed this— @czechboy0 Can you also fix this for the non-async `SessionAuthenticator`?

Yes, good catch - fixed in a7ffee1858852a089b5387adbcfb58fe4d9e4598

PR #3390, opened 2025-11-06
Fix a couple of import issues

0xTim:
**These changes are now available in [4.119.1](https://github.com/vapor/vapor/releases/tag/4.119.1)**

Fixes a couple of import issues seen recently:

* Sendable error with TOTP types on old macOS SDKs - CryptoKit types have Sendable annotations from Xcode 16.3 (Swift 6.1). If trying to compile on older versions of Xcode, which we support you get an error in Swift 6 mode
* Some UBI9 based images complain about a missing import for CNIOLinux

PR #3392, opened 2025-11-07
Explicitly import CNIOLinux module

colindignazio:
Fix "error: initializer 'init()' is not available due to missing import of defining module 'CNIOLinux'" errors when compiling on Linux.


0xTim:
@colindignazio can you test #3390 - we don't have a Linux system in our CI that fails

0xTim:
Closing in favour of #3390

PR #3403, opened 2025-11-22
Remove ConsoleKit and prepare for Swift Configuration

fpseverino:
- Update ConsoleKit to v5. Now uses only ConsoleLogger without the rest of ConsoleKit
- Prepare the ground for integrating Swift Configuration as a first class citizen (see #3421)
- Remove old ConsoleKit v4 commands
    - `serve` is now the public function `startup`, which will take a `ConfigReader` for reading hostname, port, etc.
    - `routes` is now a public function named `routesASCIITable` that returns the table as a `String`

### DIFF
--- a/Sources/Vapor/Application+AddressConfiguration.swift
+++ b/Sources/Vapor/Application+AddressConfiguration.swift
@@ -28,7 +28,6 @@ extension Application {
     struct SendableBox: Sendable {
         var didShutdown: Bool
         var running: Application.Running?
-        var signalSources: [any DispatchSourceSignal]
         var server: (any Server)?
     }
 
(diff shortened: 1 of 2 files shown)


Sources/Vapor/Logging/LoggingSystem+Environment.swift
fpseverino: Do we even need this file, since we don't bootstrap from the Environment anymore and we have [this](https://github.com/vapor/console-kit/blob/consolekit-5/Sources/ConsoleLogger/LoggingSystem%2Bbootstrap.swift) in ConsoleKit that takes a `ConfigReader`?

0xTim: No I think we can remove all this

0xTim:
Just as a reminder - we should split this into 2, possibly 3 PRs - one to add Swift Configuration, one to fix/migrate the existing commands, and one to remove ConsoleKit

Sources/Vapor/Routing/Application+routesASCIITable.swift
fpseverino: This is essentially the old `RoutesCommand.swift` file, now transformed into a function. I imagine GitHub detects too many changes to flag it as a renamed file

0xTim:
@fpseverino looks like there are some build errors on Linux at least. We can get rid of signal sources since that will be handled by service lifecycle

PR #3416, opened 2026-01-06
fix: stream file hashing for ETags (#3415)

### LINKED ISSUE #3415: FileMiddleware Hashing should be streaming

Don't load the whole file into memory

For calculating the hashes of files for etags, we currently load the whole file into memory. This is obviously inefficient (see #3405) and Crypto actually offers `update()` functions to do streaming hashing. We should use that

_No response_

elskow:
fixes #3415


elskow:
don't think this is on me @0xTim. Looks unrelated. Mind re-running it?

0xTim:
@elskow thanks for this! There were a few issues with ETags and hashing which should now be fixed. This PR was pointed against `vapor4` before that branched off, so feel free to point it there and update with a test if you'd like!

PR #3418, opened 2026-01-12
Update tests to behave correctly without AHC

ptoffy:
When the `HTTPClient` trait is disabled, tests using the client don't behave correctly since the client used is the `BlackholeClient` which fails unconditionally. This fixes it by
- hiding the `ClientTests` and the `EndpointCacheTests` behind the same trait so they won't be run, since they mainly test client functionality which doesn't make sense to test without an actual client
- creates a copy of the `VaporHTTPClient` using an actual `AsyncHTTPClient` for the `ApplicationTests` which do need application logic tested with an actual client. If just imported in tests, the AHC dependency shouldn't be propagated to the sources and therefore won't be bundled when Vapor is used as a dependency. If and when https://github.com/swiftlang/swift-package-manager/issues/9523 lands and we will have the ability to conditionally enable traits for testing purposes, we can remove the copy of the client from the tests and just use the one in the sources

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,4 +55,4 @@ jobs:
         uses: actions/checkout@v6
       - name: Run tests with traits disabled
         # Since we disable all traits here, we also disable HTTPClient, which is used by a lot of tests to send real requests. So just run a subset of tests that don't require it, since we're really just making sure it works.
-        run: swift test --disable-default-traits --filter "AuthenticationTests"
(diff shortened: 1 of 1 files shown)


0xTim:
Actually one last thing - in CI we just filter on the authentication tests - we can remove that filter now I think?

ptoffy:
The tracing middleware tests are failing (locally too) but I doubt this PR is related

PR #3421, opened 2026-01-22
Adopt Swift Configuration

### LINKED ISSUE #3389: Swift Configuration Support

Integrate it as a first class citizen

Better ecosystem integration

_No response_

fpseverino:
- Integrate Swift Configuration as a first class citizen
- Remove `arguments` from `Environment`

Resolves #3389 

This PR completes the work of #3403 

### `Environment.detect(from:)`
- `vapor.env`: (string, optional, default: `.development`): The name of the environment to use.

### `Application.AddressConfiguration.init(from:)`
- `hostname`: (string, optional): The hostname the server will run on.
- `port`: (int, optional): The port the server will run on.
- `bind`: (string, optional): The hostname and port together in the format `"hostname:port"`.
- `unix-socket`: (string, optional): The path for the unix domain socket file the server will bind to.

### DIFF
--- a/Sources/Vapor/Application.swift
+++ b/Sources/Vapor/Application.swift
@@ -289,7 +289,7 @@ public final class Application: Sendable, Service {
     /// (manually) shut down before returning.
     public func run() async throws {
         do {
-            try await self.startup(from: self.configReader)
+            try await self.startup()
             try await self.running?.onStop.get()
         } catch {
             self.logger.report(error: error)
@@ -305,10 +305,10 @@ public final class Application: Sendable, Service {
     /// If you start Vapor through this method, you'll need to prevent your Swift Executable from closing yourself.
     /// If you want to run your ``Application`` indefinitely, or until your code shuts the application down,
     /// use ``execute()`` instead.
-    public func startup(from config: ConfigReader) async throws {
+    public func startup() async throws {
         try await self.boot()
 
-        try await self._startup(addressConfiguration: AddressConfiguration(from: config))
+        try await self._startup(addressConfiguration: AddressConfiguration(from: self.configReader))
     }
 
     /// Called when the applications starts up, will trigger the lifecycle handlers. The asynchronous version of ``boot()``
--- a/Tests/VaporTests/ApplicationTests.swift
+++ b/Tests/VaporTests/ApplicationTests.swift
@@ -12,10 +12,10 @@ import RoutingKit
 struct ApplicationTests {
     @Test("Test stopping the application", .disabled())
     func testApplicationStop() async throws {
-        try await withApp { app in
+        try await withApp(configReader: testConfigReader) { app in
             //app.environment.arguments = ["serve"]
             app.serverConfiguration.address = .hostname("127.0.0.1", port: 0)
-            try await app.startup(from: testConfigReader)
+            try await app.startup()
             guard let running = app.running else {
                 Issue.record("app started without setting 'running'")
                 return
@@ -152,7 +152,7 @@ struct ApplicationTests {
 
     @Test("Test configuration address details reflected after being set")
     func testConfigurationAddressDetailsReflectedAfterBeingSet() async throws {
-        try await withApp { app in
+        try await withApp(configReader: testConfigReader) { app in
             app.serverConfiguration.address = .hostname("0.0.0.0", port: 0)
             let portPromise = Promise<Int>()
             app.serverConfiguration.onServerRunning = { channel in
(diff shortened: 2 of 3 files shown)


Sources/VaporTesting/withApp.swift
```diff
@@ -23,11 +24,12 @@ import NIOCore
 @discardableResult
 public func withApp<T>(
     address: BindAddress? = nil,
+    configReader: ConfigReader = ConfigReader(provider: EnvironmentVariablesProvider()),
```
fpseverino: Should we provide a default? And if so, is this one acceptable?

Sources/Development/entrypoint.swift
```diff
@@ -1,14 +1,21 @@
...
 import Vapor
 import Logging
 
 @main
 struct Entrypoint {
     static func main() async throws {
-        let env = try Environment.detect()
-        ConsoleLogger.bootstrap()
+        let config = ConfigReader(providers: [
+                // The `CommandLineArgumentsProvider` requires the `CommandLineArgumentsSupport` package trait
```
0xTim: Why can't we enable this for the Development target?

Sources/VaporTesting/withApp.swift
0xTim: Yes I think this is fine, at least for now

Tests/VaporTests/ClientTests.swift
```diff
@@ -110,8 +110,8 @@ struct ClientTests {
     }
 
     // MARK: - Helpers
-    func withRemoteApp<T>(_ block: (Application, Int) async throws -> T) async throws -> T {
-        let remoteApp = try await Application(.testing)
+    func withRemoteApp<T>(_ block: @Sendable (Application, Int) async throws -> T) async throws -> T {
```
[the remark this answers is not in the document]
0xTim: It's because you're passing a global instance into it, which is essentially non isolated so the closure needs to be Sendable for it to be safe. Let's not create a top level variable and wrap it in an enum that can be a computed property which should satisfy the compiler. Making the Sendable would be difficult for users

Sources/Vapor/Application.swift
```diff
@@ -79,6 +80,8 @@ public final class Application: Sendable, Service {
         }
     }
 
+    private let configReader: ConfigReader
```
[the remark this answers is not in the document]
0xTim: Yes I think it should be public so they can use a default one

Sources/Vapor/Application.swift
```diff
@@ -198,14 +201,15 @@ public final class Application: Sendable, Service {
     public convenience init(
         _ environment: Environment = .development,
         configuration: ServerConfiguration = .init(address: .hostname("127.0.0.1", port: 8080), onServerRunning: { _ in }),
+        configReader: ConfigReader,
```
0xTim: We should provide something default here to avoid people having to create it

Sources/Vapor/Application.swift
```diff
@@ -290,17 +295,15 @@ public final class Application: Sendable, Service {
...
     
-    /// When called, this will asynchronously execute the startup command provided through an argument. If no startup
-    /// command is provided, the default is used. Under normal circumstances, this will start running Vapor's webserver.
+    /// When called, this will start running Vapor's webserver.
     ///
     /// If you start Vapor through this method, you'll need to prevent your Swift Executable from closing yourself.
     /// If you want to run your ``Application`` indefinitely, or until your code shuts the application down,
     /// use ``execute()`` instead.
-    public func startup() async throws {
+    public func startup(from config: ConfigReader) async throws {
```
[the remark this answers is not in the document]
0xTim: Yes use the one passed in for now

Sources/Development/entrypoint.swift
0xTim: Or at all, what does it pull in?

ProSeFraudCatcher:
Got it.


Sean Bruce HenryFounder, Creator, Owner, Sole MBR, Mortgage Underwriter, Property Manager, Family Caretaker, Human Rights Activist, Private Investigator 
#CyberFraudBountyHunters+1 (651) ***@***.*** 8th Ave SSouth Saint Paul MN 55075Dakota CountyUnited States Of America
***@***.******@***.******@***.******@***.******@***.******@***.******@***.******@***.***


On Friday, January 23, 2026, 7:45 AM, Francesco Paolo Severino ***@***.***> wrote:





@fpseverino commented on this pull request.

On Sources/Vapor/Application+AddressConfiguration.swift:

You mean like instead of having --hostname, --port, --bind and --unix-socket, we would have just one --address flag and then we disambiguate ourselves what the flag represents? (I used CLI arguments just as an example)

fpseverino:
I don't know why, but CI on Linux never completes, it freezes

Sources/Vapor/Application.swift
0xTim: Any reason we can't do that here?

0xTim (changes requested):
One last change then I think we're there!

PR #3430, opened 2026-02-16
Fix empty response bodies being compressed

### LINKED ISSUE #3414: Empty response bodies are compressed by Vapor
### Describe the issue

Requests with an Accept-Encoding header supported by Vapor cause empty body responses to be "compressed" from 0 to 8 bytes with deflate

### Vapor version

4.120.0

macOS 15.5

### Swift version

Swift Package Manager - Swift 6.1.1

### Steps to reproduce

1. Clone [the demo repo](https://github.com/semicoleon/vapor-response-compression-bug).
2. Run the example test case, which will fail.

### Outcome

An empty response body (and maybe even any sufficiently small body) should not be compressed automatically by Vapor, but it currently is.

### Additional notes

As I was digging into this issue (which was found by a vendor calling our APIs and failing to decompress the 8 bytes the empty body was encoded into) I noticed that Vapor is using a compression predicate that checks a blacklist of `Content-Type`s to decide whether to compress, but this response doesn't have a `Content-Type` (because it's empty) so that code just passes.

gomminjae:
Fixes compression of empty response bodies (fixes #3414).

  Responses without a `Content-Type` header are now excluded from compression.

  > Note: Previously, the compression predicate only checked content types against a blacklist. Responses without a `Content-Type` header bypassed this check entirely, resulting in empty bodies being compressed from 0 to 8 bytes.

  ## Changes
  - `Sources/Vapor/HTTP/Server/HTTPServer.swift`: Added early return when `Content-Type` header is missing
  - `Tests/VaporTests/ConditionalResponseCompressionTests.swift`: Added `testGH3414` regression test

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,42 @@
+# AI & Automated Agent Policy
+
+This document outlines our policy on AI-assisted contributions and automated agents interacting with this project.
+
+## Issues
+
+### Good First Issues
+
+Issues tagged with `good-first-issue` are there to allow new contributors to learn the processs of contributing to open source and Vapor with relatively easy tasks, to help grow open source. These issues are written by human maintainers and are intended to be solved by humans learning the codebase. **Do not use automated agents to claim, triage, or submit solutions to `good-first-issue` tickets.**
+
+### Security Issues
+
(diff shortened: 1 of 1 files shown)


gwynne:
Duplicate of #3364

PR #3426, opened 2026-02-13
Add `contains` method to URLQueryContainer

william-laverty:
## Summary

Closes #3343

Adds a `contains(_:)` method to `URLQueryContainer` that checks whether a query parameter exists at a given keypath, without needing to know or care about its type.

## Motivation

As described in #3343, sometimes you just want to know if a query parameter exists — not its value. Currently you have to attempt a decode and catch the error, which is awkward.

## Changes

**`Sources/Vapor/Content/URLQueryContainer.swift`:**
- Added `contains(_ path: CodingKeyRepresentable...)` — variadic convenience
- Added `contains(at path: [CodingKeyRepresentable])` — array-based implementation

**`Tests/VaporTests/QueryTests.swift`:**
- Added `testContainsQueryParameter()` test

## Usage

```swift
// Check if a parameter exists
if req.query.contains("page") {
    // paginate
}

// Nested keypath
if req.query.contains("filter", "status") {
    // apply filter
}
```

PR #3432, opened 2026-02-18
Add `contains` method to `URLQueryContainer`

### LINKED ISSUE #3343: `URLQueryContainer` should have a `has`/`contains` method
Sometimes you don't care about the value of a query parameter, you just want to know if it exists or not.

william-laverty:
## Description

Adds `contains(_:)` and `contains(at:)` methods to `URLQueryContainer` to check whether a query parameter exists at a given keypath without needing to decode its value.

This addresses a common pattern where you only care about presence, not value:

```swift
// Before: awkward workaround
let hasPage = (try? req.query.get(String.self, at: "page")) != nil

// After: clean and readable
if req.query.contains("page") { ... }
```

Both variadic and array-based keypaths are supported, consistent with the existing `get` API.

## Changes
- Added `contains(_: CodingKeyRepresentable...)` — variadic keypath check
- Added `contains(at: [CodingKeyRepresentable])` — array keypath check
- Added test in `QueryTests.swift`

Closes #3343

### DIFF
--- a/Sources/Vapor/Content/URLQueryContainer.swift
+++ b/Sources/Vapor/Content/URLQueryContainer.swift
@@ -79,6 +79,22 @@ extension URLQueryContainer {
         try self.get(D.self, path: path.map(\.codingKey))
     }
 
+    // MARK: - Existence check
+
+    /// Check whether a query parameter exists at the supplied keypath.
+    ///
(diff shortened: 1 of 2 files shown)


william-laverty:
Closing in favor of #3426 which addresses the same issue. Apologies for the duplicate!

PR #3433, opened 2026-02-18
Use streaming hash for FileMiddleware ETag generation

### LINKED ISSUE #3415: FileMiddleware Hashing should be streaming

william-laverty:
## Summary

Uses SHA256's incremental `update` API with `readChunks` instead of `readToEnd` when generating ETag hashes in `FileMiddleware`. This avoids loading entire files into memory for hashing.

### Changes

- `generateETagHash` now reads the file in 2 MiB chunks using `handle.readChunks(in: ..., chunkLength: .mebibytes(2))`
- Each chunk is fed into `SHA256.update(data:)` incrementally
- The final digest is produced with `hasher.finalize()`
- The resulting hash is identical to the previous implementation — only memory usage changes

### Motivation

For large static files, the previous implementation loaded the entire file into a single `ByteBuffer` before hashing. This caused unnecessary memory spikes proportional to file size. The streaming approach keeps memory usage constant regardless of file size.

Fixes #3415

### DIFF
--- a/Sources/Vapor/Utilities/FileIO.swift
+++ b/Sources/Vapor/Utilities/FileIO.swift
@@ -338,8 +338,11 @@ public struct FileIO: Sendable {
             return hash.digestHex
         } else {
             return try await FileSystem.shared.withFileHandle(forReadingAt: .init(path)) { handle in
-                let buffer = try await handle.readToEnd(maximumSizeAllowed: .bytes(.max))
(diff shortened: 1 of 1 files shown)


gwynne:
Duplicate of #3416

PR #3434, opened 2026-02-20
Add contains method to URLQueryContainer

### LINKED ISSUE #3343: `URLQueryContainer` should have a `has`/`contains` method

corvid-agent:
## Summary
- Adds a non-throwing `contains(_:)` method to `URLQueryContainer` that checks whether a key exists in the query string without needing to decode its value
- Supports both simple keys (`req.query.contains("foo")`) and nested keypaths (`req.query.contains("user", "name")`)
- Works with valueless/flag-style query parameters (e.g. `?verbose`)

Resolves #3343

## Implementation
- Introduces `ContainerHasKeyExecutor`, a lightweight `Decodable` struct that uses `KeyedDecodingContainer.contains(_:)` to check for key presence without decoding values
- Follows the same keypath traversal pattern used by the existing `ContainerGetPathExecutor`
- The method is non-throwing and returns `false` when the decoder itself fails (e.g. no query string present)

## Test plan
- [x] `testContainsKey` - verifies basic key presence/absence with `?foo=a&bar=b`
- [x] `testContainsValuelessKey` - verifies valueless/flag params like `?flag`
- [x] `testContainsNoQuery` - verifies `false` when URL has no query string
- [x] `testContainsNestedKey` - verifies nested keys like `?user[name]=Vapor`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/Vapor/Content/ContainerGetPathExecutor.swift
+++ b/Sources/Vapor/Content/ContainerGetPathExecutor.swift
@@ -29,6 +29,39 @@ internal struct ContainerGetPathExecutor<D: Decodable>: Decodable {
     }
 }
 
+/// Checks whether a key exists at a key path without decoding a value.
+internal struct ContainerHasKeyExecutor: Decodable {
+    let result: Bool
+
+    static func userInfo(for keyPath: [CodingKey]) -> [CodingUserInfoKey: Sendable] {
+        [.containerGetKeypath: keyPath]
+    }
+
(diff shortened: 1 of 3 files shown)


0xTim:
@corvid-agent stop submitting pull requests to all repositories under the Vapor organisation

PR #3446, opened 2026-03-28
Pass through source to swift-log from Logger+Report.swift

### LINKED ISSUE #2877: Handled a DebuggableError
### Describe the bug

When an error is handled as a DebuggableError in the Logger+report file, the log function is called, ignoring source. Because of this, there is no way to log #file, #function, #line. To fix this, you need to call another function from swift-log:
public func log(level: Logger.Level,
                     _message: @autoclosure() -> Logger.Message,
                     metadata: @autoclosure() -> Logger.Metadata? = nil,
                     source: @autoclosure() -> String? = nil,
                     file: String = #fileID, function: String = #function, line: UInt = #line)

* Vapor Framework version: 4.65.1
* Vapor Toolbox version:
* OS version: macOS Monterey

### Additional context
Maybe I don't know how to cook properly?

bplattenburg:
Allow passthrough of the `source` of errors to swift-log when using `Logger+Report` - without this swift-log can eat some of the original context.

Closes #2877

### DIFF
--- a/Sources/Vapor/Logging/Logger+Report.swift
+++ b/Sources/Vapor/Logging/Logger+Report.swift
@@ -9,11 +9,12 @@ extension Logger {
     public func report(
         error: Error,
         metadata: @autoclosure () -> Logger.Metadata? = nil,
+        source: @autoclosure() -> String? = nil,
         file: String = #fileID,
         function: String = #function,
         line: UInt = #line
     ) {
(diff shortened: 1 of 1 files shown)


bplattenburg:
I'm not 100% sure I've solved the original requester's issue, but this feels like a "free" improvement to allow people pass things through if needed

PR #3455, opened 2026-04-19
fix(validation): prevent partial range in validator hang

### LINKED ISSUE #3048: Using the `.in()` validator with a PartialRange never completes
### Describe the bug

I have a test that validates a JSON object query. One of the object's properties is an integer, and it's corresponding validation is as follows:

```swift
    validations.add(.offset, as: Int.self, is: .in(0...))
  }
```

When the test runs, because the `Sequence` passed to `.in()` is unbounded, the test continues to run forever, racking up more and more memory:
Screenshot showing extreme memory usage of test

### To Reproduce

1. See above description.

### Expected behavior

The `.in()` validation should not accept an unbounded or partial range.

### Environment

* Vapor Framework version: 4.77.2
* Vapor Toolbox version: 18.7.1
* OS version: 13.4.1

dioKaratzas:
Updates the `.in` validator overload to accept finite collections instead of arbitrary sequences.

Previously, `.in(0...)` compiled because `PartialRangeFrom<Int>` conforms to `Sequence`. During validation, Vapor materialized the supplied sequence into an array for `ValidatorResults.In`, which meant `Array(0...)` never completed and could consume unbounded memory.

This change narrows the generic `.in` overload from `Sequence` to `Collection`, preserving supported finite inputs such as arrays and closed ranges while preventing unbounded partial ranges from being passed to `.in`.

The correct API for lower/upper-bound validation remains `.range(...)`. For example:

```swift
validations.add("offset", as: Int.self, is: .range(0...))
```

Fixes #3048.

Validation:
- Not run, per request.


0xTim:
Feel free to reopen targeting the `main` branch!

dioKaratzas:
Thanks @0xTim! Reopened against `main` (rebased onto latest) here: #3477 — the original branch targeted `vapor-5`, so this is a fresh PR to `main`.

PR #3456, opened 2026-04-19
fix(routing): enforce configurable body size

### LINKED ISSUE #3388: Support configurable body size

Make it work again

Currently commented out - bring it back

_No response_

dioKaratzas:
Updates route body collection so configurable max body size is enforced for requests handled by the Vapor 5 server path.

Previously, the route collection logic only collected `request.body`, but requests coming through `HTTPServerNew` store their body in `newBodyStorage`. That meant routes using `.collect` could skip body-size enforcement and continue to the handler, returning `200 OK` instead of `413 Content Too Large`.

This change collects from `newBodyStorage` when present, stores the collected buffer back into `bodyStorage` for the existing Vapor request APIs, and preserves the existing `request.body.collect(max:)` fallback for old body storage.

Also re-enables the configurable max body size test for running server requests.

Fixes #3388.

### DIFF
--- a/Sources/Vapor/Routing/RoutesBuilder+Method.swift
+++ b/Sources/Vapor/Routing/RoutesBuilder+Method.swift
@@ -2,6 +2,7 @@ import RoutingKit
 import HTTPTypes
 import NIOPosix
 import NIOCore
+import NIOConcurrencyHelpers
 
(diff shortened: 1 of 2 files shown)


0xTim:
Feel free to reopen targeting the `main` branch - this may need some more work to ensure we use the new HTTP server correctly

PR #3457, opened 2026-04-19
fix(auth): add basic auth challenge header

### LINKED ISSUE #2337: Should BasicAuthenticator add WWW-Authenticate header?
### Summary
Routes protected by `BasicAuthenticator`s do not properly trigger browser or `URLSession` authentication prompts due to the lack of a `WWW-Authenticate` header.

### Steps to reproduce
1. Create a simple `BasicAuthenticator`, like this one:

```swift
struct BasicPathAuthenticator: BasicAuthenticator {
    enum Error: Swift.Error { case invalidRequest, invalidCredentials }
    
    func authenticate(basic: BasicAuthorization, for request: Request) -> EventLoopFuture<Void> {
        guard let username = request.parameters["user", as: String.self],
            let password = request.parameters["passwd", as: String.self] else {
                return request.eventLoop.makeFailedFuture(Error.invalidRequest)
        }
        
        guard basic.username == username, basic.password == password else {
            return request.eventLoop.makeFailedFuture(Error.invalidCredentials)
        }
        
        request.auth.login(request)
    }
}
```

using it like this:

```swift
let basicAuth = app.grouped(BasicPathAuthenticator())
basicAuth.on([.GET, .POST, .PUT, .PATCH, .DELETE], "basic-auth", ":user", ":passwd") { request -> EventLoopFuture<Response> in
    guard request.isAuthenticated else {
        return request.eventLoop.makeSucceededFuture(Response(status: .unauthorized))
    }
    
    return try reply(to: request).encodeResponse(for: request)
}
```

2. Make a request, notice no automatic prompting from browser, or `URLSession` automatically applying `URLCredential`s.

### Expected behavior

Responses returned using the `BasicAuthorization` should have the `WWW-Authenticate` header applied automatically so that clients perform automatic behaviors.

### Actual behavior

Header has to be added manually in the route's response

### Environment

* Vapor Framework version: 4.4.0
* Vapor Toolbox version: ?
* OS version: 10.15.4

### Discussion

Perhaps I'm misunderstanding the role of `Middleware` here, or the role of `Authorization` middleware, but this seems like a desirable behavior. I was unaware of the `WWW-Authenticate` header and how it triggered the automatic behavior, which is my own ignorance of backend development. I was rather surprised to see a difference in behavior between the Basic auth offered by vapor and the basic auth offered by, say, a web server, like Apache, which adds the header automatically.

dioKaratzas:
Adds a typed `WWW-Authenticate` header helper and updates Basic authentication middleware to include a Basic challenge on unauthorized responses.

Previously, routes protected by `BasicAuthenticator` could return `401 Unauthorized` without a `WWW-Authenticate` header, which prevents clients such as browsers and `URLSession` from recognizing the response as an authentication challenge.

This change adds a default Basic realm of `Vapor`, while allowing authenticators to customize the realm by overriding `realm`:

```swift
struct AdminAuthenticator: BasicAuthenticator {
    var realm: String { "Admin" }

    func authenticate(basic: BasicAuthorization, for request: Request) async throws {
        // Authenticate credentials.
    }
}
```

Unauthorized Basic auth responses now include:

```http
WWW-Authenticate: Basic realm="Vapor"
```

If a response or thrown abort already includes `WWW-Authenticate`, Vapor preserves the existing header.

Fixes #2337.

### DIFF
--- a/Sources/Vapor/Authentication/Authenticator.swift
+++ b/Sources/Vapor/Authentication/Authenticator.swift
@@ -1,4 +1,5 @@
 import NIOCore
+import HTTPTypes
 
 /// Capable of being authenticated.
 public protocol Authenticatable: Sendable { }
@@ -27,10 +28,38 @@ extension RequestAuthenticator {
 
 /// Helper for creating authentication middleware using the Basic authorization header.
 public protocol BasicAuthenticator: RequestAuthenticator {
+    /// Realm to advertise in the `WWW-Authenticate` challenge.
+    var realm: String { get }
+
     func authenticate(basic: BasicAuthorization, for request: Request) async throws
 }
 
 extension BasicAuthenticator {
+    public var realm: String {
+        "Vapor"
+    }
+
+    private var authenticateHeaderValue: HTTPFields.WWWAuthenticate {
+        .basic(realm: self.realm)
+    }
+
+    public func respond(to request: Request, chainingTo next: any Responder) async throws -> Response {
+        do {
+            try await self.authenticate(request: request)
+            let response = try await next.respond(to: request)
+            if response.status == .unauthorized, response.headers[.wwwAuthenticate] == nil {
+                response.headers.wwwAuthenticate = self.authenticateHeaderValue
+            }
+            return response
+        } catch let error as any AbortError where error.status == .unauthorized {
+            var headers = error.headers
+            if headers[.wwwAuthenticate] == nil {
+                headers.wwwAuthenticate = self.authenticateHeaderValue
+            }
+            throw Abort(.unauthorized, headers: headers, reason: error.reason)
+        }
+    }
+
     public func authenticate(request: Request) async throws {
         guard let basicAuthorization = request.headers.basicAuthorization else {
             return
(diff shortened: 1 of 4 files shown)


Sources/Vapor/Authentication/Authenticator.swift
```diff
@@ -27,10 +28,38 @@ extension RequestAuthenticator {
...
+
+    public func respond(to request: Request, chainingTo next: any Responder) async throws -> Response {
+        do {
+            try await self.authenticate(request: request)
+            let response = try await next.respond(to: request)
+            if response.status == .unauthorized && response.headers[.wwwAuthenticate] == nil {
+                response.headers.wwwAuthenticate = self.authenticateHeaderValue
+            }
+            return response
+        } catch let error as any AbortError where error.status == .unauthorized {
```
dioKaratzas, line 54: The `catch` covers the common `BasicAuthenticator` + `guardMiddleware()` path. `guardMiddleware()` throws `Abort(.unauthorized)` instead of returning a response, so without catching the `AbortError` here, `BasicAuthenticator` never gets a response to attach `WWW-Authenticate` to. We only handle unauthorized aborts, preserve any existing header, and rethrow with the Basic challenge for `ErrorMiddleware` to render normally.

0xTim:
@dioKaratzas feel free to reopen this targeting the main branch now it's merged in

dioKaratzas:
Thanks @0xTim! Reopened against `main` (rebased onto latest) here: #3478 — the original branch targeted `vapor-5`, so this is a fresh PR to `main`.

PR #3462, opened 2026-05-21
Fix request query percent encoding

### LINKED ISSUE #3348: Some query strings are double percent-encoded
### Describe the issue

`GET /hello?yolo=<\` (string ends with a literal backslash) is decoded as `["yolo": "%3C\\"]` (string also ends with one literal backslash)

### Vapor version

4.115.1

macOS 15.5 (24F74)

### Swift version

Swift Package Manager - Swift 6.1.0

### Steps to reproduce

Write a vapor server with an endpoint that just does this:
```swift
print(req.url.query)
try? print(req.query.decode([String: String].self))
return "whatever"
```

Then run the endpoint with the given query string.

### Outcome

The query string is improperly decoded.

### Additional notes

The source of the problem lies in `URLComponents` whose behavior, as is well-known, is completely inconsistent with itself.
In particular in our case, the browser sends the following exact query string to the server: `yolo=%3C\`. `URLComponents` will see a character which is supposed to be escaped in a query string (the backslash) _not_ escaped, so it will escape the whole query string. The same happens if the backslash is replaced with a percent sign.
Indeed, if backslash is removed, the query string is properly parsed.

I understand this is not Vapor’s fault, but I guess it’s Vapor’s problem anyways… I’m not even sure as to how properly fix this issue tbh, but it is a fairly important issue IMHO…

puneetdixit200:
Fixes #3348.

This preserves already-percent-encoded query escapes when Vapor parses an HTTP request target through `URI(path:)`, while still percent-encoding raw query characters that `URLComponents` rejects in a `percentEncodedQuery` value. That prevents valid escapes such as `%3C` from being encoded again when the same query also contains a raw character like `\`.

I added a regression test that sends `/hello?yolo=%3C\` through the test client and asserts that query decoding produces `<\`.

Validation:
- `swift build --target Vapor`
- `swift test --filter QueryTests/testGH3348` could not run on this machine because only Command Line Tools are installed and the local Swift toolchain cannot import `XCTest`; GitHub Actions should run the test target in the normal CI environment.

### DIFF
--- a/Sources/Vapor/Utilities/URI.swift
+++ b/Sources/Vapor/Utilities/URI.swift
@@ -109,7 +109,11 @@ public struct URI: CustomStringConvertible, ExpressibleByStringInterpolation, Ha
             // systematic misuse of both the URI type and concept"), we must collapse any non-zero number of
             // leading `/` characters into a single character (thus breaking the ability to parse what is otherwise a
             // valid URI format according to spec) to avoid weird routing misbehaviors.
-            components = URL(string: "/\(path.drop(while: { $0 == "/" }))").flatMap { .init(url: $0, resolvingAgainstBaseURL: true) }
+            let path = "/\(path.drop(while: { $0 == "/" }))"
+            components = URL(string: path).flatMap { .init(url: $0, resolvingAgainstBaseURL: true) }
+            if let rawQuery = path.rawQueryForPathURI {
+                components?.percentEncodedQuery = rawQuery.withQueryPercentEncodingPreservingEscapes
+            }
         } else {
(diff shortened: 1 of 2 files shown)


0xTim:
@puneetdixit200 if you want to update the base branch to `vapor4` we can look at merging this

gwynne:
Technically, Vapor's current behavior here is correct, in that `\` is not a valid character anywhere in a URL according to [RFC 3986 § Appendix A](https://www.ietf.org/rfc/rfc3986.html#appendix-A). Clients sending this as a URL are in violation of spec.

PR #3473, opened 2026-06-17
HTTPServer: add idle-connection timeout option

hassang2:
Adds a knob to specify an `idleTimeout`. if a connection has no activity (read AND write) for this amount of time, we close out the connection.

### DIFF
--- a/Sources/Vapor/HTTP/Server/CloseOnIdleHandler.swift
+++ b/Sources/Vapor/HTTP/Server/CloseOnIdleHandler.swift
@@ -1,0 +1,24 @@
+import Logging
+import NIOCore
+
+/// Closes the channel on receipt of an `IdleStateHandler.IdleStateEvent`.
+final class CloseOnIdleHandler: ChannelInboundHandler {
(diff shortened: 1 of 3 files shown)


hassang2:
Thank you for taking a look! I have addressed your suggestions.

gwynne (approved):
This seems alright to me. @0xTim Want to take a look?

hassang2:
Unfortunate a new initializer is required for every new configuration knob to avoid API breakage.

gwynne:
> Unfortunate a new initializer is required for every new configuration knob to avoid API breakage.

Truth is, we tend to not enforce that particular concern, since it quickly results in a huge proliferation of methods and is only source-breaking in very unusual cases. 😅

hassang2:
ah makes sense. Do let me know if you'd like me to revert that change before merge.

0xTim (approved):
Thanks @hassang2 - if you want to remove the extra init, that would be great. A default parameter is fine

hassang2:
done

PR #3475, opened 2026-06-22
Add typed Retry-After HTTP header accessor

somaz94:
The `.retryAfter` header name constant already exists in Vapor, but there was no
typed accessor for it. This adds one, following the existing `Expires` and
`Last-Modified` helpers in the same directory.

`HTTPHeaders.retryAfter` exposes a `RetryAfter` enum with two cases: `.seconds(Int)`
for a delay and `.date(Date)` for an HTTP-date. The getter parses the raw header
value (negative delay-seconds and unparseable dates return nil), the setter
serializes it, and setting nil removes the header. Date parsing reuses the same
three formats as the `Expires` helper (IMF-fixdate plus the two obsolete formats).

This is commonly paired with 429 and 503 responses.

Tests cover the seconds case, the date case, invalid/negative input, and removal.
Ran the header suite on Linux (swift:6.0-jammy): 45 passed, 0 failed.

### DIFF
--- a/Sources/Vapor/HTTP/Headers/HTTPHeaderRetryAfter.swift
+++ b/Sources/Vapor/HTTP/Headers/HTTPHeaderRetryAfter.swift
@@ -1,0 +1,85 @@
+import Foundation
+import NIOHTTP1
+
+extension HTTPHeaders {
+    /// Represents the HTTP `Retry-After` header.
+    ///
+    /// The value is either a number of seconds to wait before retrying, or a specific
+    /// date after which the client may retry. This is typically sent with `429 Too Many
+    /// Requests` or `503 Service Unavailable` responses.
+    /// - See Also:
+    /// [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After)
+    public enum RetryAfter {
+        /// The number of seconds to wait before retrying.
+        case seconds(Int)
+        /// The date after which the client may retry.
+        case date(Date)
+
+        internal static func parse(_ value: String) -> RetryAfter? {
+            let trimmed = value.trimmingCharacters(in: .whitespaces)
+
+            // `delay-seconds` is a non-negative integer (RFC 9110 § 10.2.3).
+            if let seconds = Int(trimmed) {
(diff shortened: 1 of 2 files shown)


somaz94:
@0xTim @gwynne — gentle ping. This adds a typed accessor for the `Retry-After` header.

You were both added as reviewers back on 8 Jul, but I don't think there's anything to review against yet: the `test` workflow on the current head sha is sitting at `action_required` with zero check runs, so CI has never actually executed. My force-push on 24 Jul is what re-armed it.

Could one of you approve the run? Happy to address anything it turns up.

Sources/Vapor/HTTP/Headers/HTTPHeaderRetryAfter.swift
```diff
@@ -0,0 +1,85 @@
...
+
+extension HTTPHeaders {
+    /// Represents the HTTP `Retry-After` header.
+    ///
+    /// The value is either a number of seconds to wait before retrying, or a specific
+    /// date after which the client may retry. This is typically sent with `429 Too Many
+    /// Requests` or `503 Service Unavailable` responses.
+    /// - See Also:
+    /// [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After)
+    public enum RetryAfter {
```
0xTim: This probably should be a struct with static properties, public enums can be very problematic

0xTim: Oh and should also be `Sendable`

0xTim (changes requested):
Thanks, this looks good! Couple of nits

somaz94:
Thanks! Both done.

`RetryAfter` is now a struct with a private storage enum, conforming to `Sendable` and `Equatable`.

Two decisions worth flagging:

- Both cases carry a payload, so they are static *methods* rather than static properties — `.seconds(120)` and `.date(someDate)` read the same at the call site.
- A struct cannot be switched over exhaustively, so I added `seconds: Int?` and `date: Date?` to read the value back. The tests use those now.

CI still has not run here — the workflow is sitting at `action_required`. Could you approve it when you get a chance?

PR #3477, opened 2026-07-02
fix(validation): prevent partial range in validator hang

### LINKED ISSUE #3048: Using the `.in()` validator with a PartialRange never completes

dioKaratzas:
Reopens #3455 against `main` as requested by @0xTim (the original PR targeted `vapor-5`, which has since been merged into `main`). Rebased onto the latest `main`.

---

Updates the `.in` validator overload to accept finite collections instead of arbitrary sequences.

Previously, `.in(0...)` compiled because `PartialRangeFrom<Int>` conforms to `Sequence`. During validation, Vapor materialized the supplied sequence into an array for `ValidatorResults.In`, which meant `Array(0...)` never completed and could consume unbounded memory.

This change narrows the generic `.in` overload from `Sequence` to `Collection`, preserving supported finite inputs such as arrays and closed ranges while preventing unbounded partial ranges from being passed to `.in`. Because `PartialRangeFrom` does not conform to `Collection`, `.in(0...)` now fails to compile rather than hanging at runtime.

The correct API for lower/upper-bound validation remains `.range(...)`. For example:

```swift
validations.add("offset", as: Int.self, is: .range(0...))
```

Fixes #3048.


0xTim (approved):
Thanks!

PR #3467, opened 2026-06-15
Prefer FoundationEssentials

ptoffy:
We should use FoundationEssentials where possible to avoid the Foundation dependency bloat on Linux

As part of this, Swift `Regex` doesn't support the old `regexInternationalEmail` pattern we had which used code-point ranges (e.g `\u0080-\u00FF\u0100-\u017...`), so I had to replace it to use unicode property classes

PR #3478, opened 2026-07-02
fix(auth): add basic auth challenge header

### LINKED ISSUE #2337: Should BasicAuthenticator add WWW-Authenticate header?

dioKaratzas:
Reopens #3457 against `main` as requested by @0xTim (the original PR targeted `vapor-5`, which has since been merged into `main`). Rebased onto the latest `main`.

---

Adds a typed `WWW-Authenticate` header helper and updates Basic authentication middleware to include a Basic challenge on unauthorized responses.

Previously, routes protected by `BasicAuthenticator` could return `401 Unauthorized` without a `WWW-Authenticate` header, which prevents clients such as browsers and `URLSession` from recognizing the response as an authentication challenge.

This change adds a default Basic realm of `Vapor`, while allowing authenticators to customize the realm by overriding `realm`:

```swift
struct AdminAuthenticator: BasicAuthenticator {
    var realm: String { "Admin" }

    func authenticate(basic: BasicAuthorization, for request: Request) async throws {
        // Authenticate credentials.
    }
}
```

Unauthorized Basic auth responses now include:

```http
WWW-Authenticate: Basic realm="Vapor"
```

If a response or thrown abort already includes `WWW-Authenticate`, Vapor preserves the existing header.

Fixes #2337.


Sources/Vapor/Authentication/Authenticator.swift
```diff
@@ -27,10 +28,38 @@ extension RequestAuthenticator {
...
+            if response.status == .unauthorized, response.headers[.wwwAuthenticate] == nil {
+                response.headers.wwwAuthenticate = self.authenticateHeaderValue
+            }
+            return response
+        } catch let error as any AbortError where error.status == .unauthorized {
+            var headers = error.headers
+            if headers[.wwwAuthenticate] == nil {
+                headers.wwwAuthenticate = self.authenticateHeaderValue
+            }
+            throw Abort(.unauthorized, headers: headers, reason: error.reason)
```
0xTim: We probably want to try and keep the source/details etc from the original error as well (might need to do a check to see if it conforms to `DebuggableError`)

Sources/Vapor/Authentication/Authenticator.swift
```diff
@@ -27,10 +28,38 @@ extension RequestAuthenticator {
...
+            let response = try await next.respond(to: request)
+            if response.status == .unauthorized, response.headers[.wwwAuthenticate] == nil {
+                response.headers.wwwAuthenticate = self.authenticateHeaderValue
+            }
+            return response
+        } catch let error as any AbortError where error.status == .unauthorized {
+            var headers = error.headers
+            if headers[.wwwAuthenticate] == nil {
+                headers.wwwAuthenticate = self.authenticateHeaderValue
+            }
```
0xTim: Could we do the same change in the bearer middleware as well?

Sources/Vapor/HTTP/Headers/HTTPFields+WWWAuthenticate.swift
```diff
@@ -0,0 +1,46 @@
+import Foundation
```
0xTim: See the recent work for `FoundationEssentials` - we don't want to pull in all of Foundation on Linux

Tests/VaporTests/AuthenticationTests.swift
```diff
@@ -96,6 +98,36 @@ struct AuthenticationTests {
         }
     }
 
+    @Test("Test Basic Authenticator WWW Authenticate Header")
```
0xTim, line 108: This is a good start but I think we need to expand the test suite:
* what happens if the user has manually set the `www-authenticate` header? We don't want to overwrite that
* test for bearer `www-authenticate` as well
* test throwing custom errors with reasons etc
* test a throwing authenticator

Tests/VaporTests/HTTPHeaderTests.swift
```diff
@@ -111,6 +111,20 @@ struct HTTPHeaderTests {
         #expect(serializer.serialize() == "foo; bar=\"baz\", qux=\"quuz\"")
     }
 
+    @Test("Test WWW-Authenticate Header")
+    func testWWWAuthenticate() throws {
```
0xTim, line 119: Can we also test an empty realm and test a realm with a backslash in?

0xTim (changes requested):
This looks good! Just a few edge cases to sort out

dioKaratzas:
Thanks @0xTim, think that's all of them.

The catch block now pulls  `identifier`  and  `source`  off the error when it's a  `DebuggableError`. And if the error already has a  `WWW-Authenticate`  set we don't catch it at all now, so it reaches  `ErrorMiddleware`  exactly as thrown.

Bearer got the same treatment,  `realm`  and the challenge handling, sharing a helper with Basic.

Dropped the Foundation import entirely rather than going the FoundationEssentials route  `replacingOccurrences`  was the only thing needing it, so the escaping is plain stdlib now.

Tests cover bearer, throwing authenticators, custom reasons, challenges set manually (returned and thrown), non-401 and non-`AbortError`, and empty/backslash realms.

Couple of things I wasn't sure on:

I added  `charset="UTF-8"`  to the Basic challenge, which is a bit beyond what you asked. Came out of  [#2337](https://github.com/dioKaratzas/vapor/issues/2337) , we decode credentials as UTF-8 but never advertised it, so clients fall back to ISO-8859-1 and non-ASCII passwords fail even when they're correct. Both workarounds in that issue set it. Say the word and I'll drop it.

Chaining is a bit odd. Middleware is applied outermost-first, so the innermost authenticator stamps the challenge and everything above sees a header already there, a route with Basic + Bearer only ever advertises one. RFC 7235 allows multiple so appending is probably better, but I wrote a test for the current behaviour instead of changing it. Separate issue?

Last thing, you mentioned in  [#2337](https://github.com/dioKaratzas/vapor/issues/2337)  that ErrorMiddleware should own this. I put it on the authenticators because ErrorMiddleware only sees thrown errors, and the original report returns a 401 straight from the handler, so that case never reaches it. It also can't know which scheme a route uses. Can move it if you still prefer that.

Sources/Vapor/HTTP/Headers/HTTPFields+WWWAuthenticate.swift
```diff
@@ -0,0 +1,61 @@
...
+    public struct WWWAuthenticate: ExpressibleByStringLiteral, Equatable, Sendable {
+        /// The serialized header value.
+        public let value: String
+
+        /// Creates a `WWW-Authenticate` header from a serialized header value.
+        public init(value: String) {
+            self.value = value
+        }
+
+        public init(stringLiteral value: String) {
```
0xTim, line 17: Missing DocC comment

0xTim (changes requested):
Thanks for this, this looks good! One minor nit and we're good. I'm going to hold off merging until #3467 is in as that should fix CI so we can make sure the tests pass

Sources/Vapor/HTTP/Headers/HTTPFields+WWWAuthenticate.swift
dioKaratzas: Ready!

PR #3485, opened 2026-07-27
Add a task local for Vapor.Request, and a matching middleware

czechboy0:
## Summary

> Note: This is primarily driven by the need of swift-openapi-vapor, but technically other adopters could benefit from it, so I'm starting with a PR here. Let me know if you'd like me to move it to swift-openapi-vapor instead.

This PR adds a task local for Vapor.Request, and a middleware that injects the current request into that task local automatically.

In projects like Swift OpenAPI Vapor, adopters don't get passed Vapor.Request directly, yet they sometimes need to inspect it (for auth, logger, etc). Putting the request into the task local is the easiest way to address that need.

If we land this, it'll remove the need for adopters of Swift OpenAPI Vapor to manually implement a similar middleware in each of their projects, plus it'll make it easier to inspect e.g. auth information in structured contexts.

## Test Plan

Added unit tests.

### DIFF
--- a/Sources/Vapor/Middleware/RequestTaskLocalMiddleware.swift
+++ b/Sources/Vapor/Middleware/RequestTaskLocalMiddleware.swift
@@ -1,0 +1,43 @@
+
+extension Request {
+    /// The request associated with this task.
+    ///
+    /// The recommended way to set this task-local value is by adding ``RequestTaskLocalMiddleware`` to your app's middleware stack.
+    @TaskLocal public static var current: Request?
+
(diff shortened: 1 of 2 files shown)


0xTim (changes requested):
I'm inclined to say this should live in OpenAPI Vapor, since most other use cases just pass the `Request` manually.

Also we should add new tests as Swift Testing tests

czechboy0:
Gotcha - will move it and update the test

PR #3486, opened 2026-07-28
Migrate to Task Local Logger

0xTim:
Migrate to SwiftLog's new task local logger. Means we don't need to pass the Logger everywhere


Sources/VaporTesting/withApp.swift
```diff
@@ -29,26 +30,29 @@ import ServiceLifecycle
 public func withApp<T>(
     address: BindAddress? = nil,
     configReader: ConfigReader = ConfigReader(providers: [CommandLineArgumentsProvider(), EnvironmentVariablesProvider()]),
+    logger: Logger? = nil,
```
fpseverino: Why not
```suggestion
    logger: Logger = Logger.current,
```
?

It removes an unnecessary `nil` check and also shows up in DocC making the behavior clearer

Package.swift
```diff
@@ -86,9 +86,6 @@ let package = Package(
         // Service Lifecycle Management
         .package(url: "https://github.com/swift-server/swift-service-lifecycle.git", from: "2.6.3"),
 
-        // Network IO on Apple platforms
-        .package(url: "https://github.com/apple/swift-nio-transport-services.git", from: "1.20.0"),
```
fpseverino, line 90: Unrelated, out of curiosity: is it being removed because `swift-http-server` handles it now?

fpseverino (approved):
So the plan is for final users to use `Logger.current` directly in route handlers? I love it! 🚀

I don’t think the task-local logger is used in any SwiftNIO code on an `EventLoop` (although I’m not entirely familiar with the codebase), so everything should be fine

Just a quick note on the public API

Package.swift
0xTim: Yeah we don't need to worry about this anymore - at least not for now until we hook up configuration etc

Sources/VaporTesting/withApp.swift
0xTim: Yeah good point, will use that

0xTim:
> So the plan is for final users to use `Logger.current` directly in route handlers? I love it! 🚀
> 
> I don’t think the task-local logger is used in any SwiftNIO code on an `EventLoop` (although I’m not entirely familiar with the codebase), so everything should be fine

@fpseverino yep `Logger.current` direct. We don't have any futures in Vapor anymore so all good!

PR #3487, opened 2026-07-28
Changes for MultipartKit Beta

0xTim:
Update for latest MultipartKit to get CI passing again

### DIFF
--- a/Sources/Vapor/Multipart/File+Multipart.swift
+++ b/Sources/Vapor/Multipart/File+Multipart.swift
@@ -4,16 +4,18 @@ import HTTPTypes
 import NIOCore
 
 extension File: MultipartPartConvertible {
-    public var multipart: MultipartPart<ByteBufferView>? {
+    public typealias Body = ByteBufferView
+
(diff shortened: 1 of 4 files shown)


Sources/Vapor/Multipart/File+Multipart.swift
```diff
@@ -4,16 +4,18 @@ import HTTPTypes
...
         part.contentType = self.contentType?.serialize()
         part.filename = self.filename
         return part
     }
     
-    public init?(multipart: MultipartPart<some MultipartPartBodyElement>) {
+    public init(multipart: MultipartPart<some MultipartPartBodyElement>) throws {
         guard let filename = multipart.filename else {
-            return nil
+            throw Abort(.badRequest)
```
ptoffy: Maybe this should be a custom error?

0xTim: I think in this instance it's right because we say we expect it to be provided

ptoffy (approved):

PR #3490, opened 2026-07-29
Fix the FileIO code examples

### LINKED ISSUE #2974: Wrong api docs
### Describe the bug

Found wrong api docs [here](https://docs.vapor.codes/advanced/files/#read)，there is no async version of FileIO.readFile method.

### Environment

* Vapor Framework version: 4.7.0
* Vapor Toolbox version:
* OS version:

dioKaratzas:
Fixes #2974.

#2973 fixed this in 2023, but the file got folded into `Utilities/FileIO.swift` for Vapor 5 and the comments drifted again. None of the examples compile.

Also reworded the warning on `readFile`: it says you have to close the file handle yourself, but it's wrapped in `withFileHandle` so that already happens.

Comments only.

### DIFF
--- a/Sources/Vapor/Utilities/FileIO.swift
+++ b/Sources/Vapor/Utilities/FileIO.swift
@@ -23,19 +23,19 @@ extension Request {
 ///
 /// It can read files, both in their entirety and chunked.
 ///
-///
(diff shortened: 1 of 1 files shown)


0xTim (approved):
Thanks!

PR #3494, opened 2026-07-31
Modernise Authentication

0xTim:
Update Authentication to move away from the old Storage based system.

To be honest, this is actually pretty similar, just should be a bit more performant. I did try the task local value based approach which would be a bit 'better' but it made the API for authenticating worse and ever more importantly would scope authentication to the next closure only, making logout and session based authentication impossible


Sources/Vapor/Authentication/Authentication.swift
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /// Returns `true` if the type has been authenticated.
+    public func has<A: Authenticatable>(_ type: A.Type = A.self) -> Bool {
+        self.storage.withLock { $0.contains(A.self) }
+    }
+}
+
+/// The actual storage for the request's authentication. Uses an array instead of a dictionary as
+/// this was more performant with a small number of entries
```
gwynne, line 63: Define "more performant"—is the difference noticeable in practical terms?

Sources/Vapor/Authentication/Authentication.swift
```diff
@@ -0,0 +1,85 @@
...
+    /// Unauthenticates an authenticatable type.
+    public func logout<A: Authenticatable>(_ type: A.Type = A.self) {
+        self.storage.withLock { $0.remove(A.self) }
+    }
+
+    /// Returns an instance of the supplied type. Throws if no
+    /// instance of that type has been authenticated or if there
+    /// was a problem.
+    @discardableResult
+    public func require<A: Authenticatable>(_ type: A.Type = A.self) throws -> A {
```
gwynne:
```suggestion
    /// Look up an instance of the supplied type, treating it as an error if there is no such instance.
    ///
    /// - Parameter type: The authenticatable type to look up.
    /// - Returns: An instance of the requested type, if one exists.
    /// - Throws: `Abort(.unauthorized)` if no instance of the requested type is available.
    @discardableResult
    public func require<A: Authenticatable>(_ type: A.Type = A.self) throws(Abort) -> A {
```

Sources/Vapor/Authentication/Authentication.swift
0xTim: Currently no, in the future maybe. Dictionary has 6x instruction count compared to Array for lookups, but we're talking 23 vs 136 instructions. It might matter more once all the other low hanging fruit has been fixed. Happy to defer if you think it's better to wait and see.

(I have a branch with benchmarks which I'll PR in a bit)

Sources/Vapor/Authentication/Authentication.swift
```diff
@@ -0,0 +1,87 @@
...
+    public func require<A: Authenticatable>(_ type: A.Type = A.self) throws(Abort) -> A {
+        guard let a = self.get(A.self) else {
+            throw Abort(.unauthorized)
+        }
+        return a
+    }
+
+    /// Returns the authenticated instance of the supplied type.
+    ///
+    /// > Note: `nil` if no type has been authenticated.
```
gwynne:
```suggestion
    /// Look up an instance of the supplied type, if one has been authenticated.
    ///
    /// - Parameter type: The authenticatable type to look up
    /// - Returns: An instance of the requested type, if one exists, `nil` otherwise
```

Sources/Vapor/Authentication/Authentication.swift
```diff
@@ -0,0 +1,87 @@
...
+    public func get<A: Authenticatable>(_ type: A.Type = A.self) -> A? {
+        self.storage.withLock { $0.value(A.self) }
+    }
+
+    /// Returns `true` if the type has been authenticated.
```
gwynne:
```suggestion
    /// Look up whether or not an instance of the supplied type has been authenticated.
    ///
    /// - Parameter type: The authenticatable type to look up
    /// - Returns: `true` if an instance of the given type is available, `false` otherwise
```

Sources/Vapor/Authentication/Authentication.swift
```diff
@@ -0,0 +1,87 @@
...
+    /// Unauthenticates an authenticatable type.
+    public func logout<A: Authenticatable>(_ type: A.Type = A.self) {
+        self.storage.withLock { $0.remove(A.self) }
+    }
+
+    /// Look up an instance of the supplied type, treating it as an error if there is no such instance.
+    ///
+    /// - Parameter type: The authenticatable type to look up.
+    /// - Returns: An instance of the requested type, if one exists.
+    /// - Throws: `Abort(.unauthorized)` if no instance of the requested type is available.
```
gwynne:
```suggestion
    /// - Parameter type: The authenticatable type to look up
    /// - Returns: An instance of the requested type, if one exists
    /// - Throws: `Abort(.unauthorized)` if no instance of the requested type is available
```

Sources/Vapor/Authentication/Authentication.swift
gwynne: Doesn't seem like it matters that much to me, I was just curious. 🙂

Sources/Vapor/Authentication/Authenticator.swift
```diff
@@ -1,23 +1,23 @@
 import NIOCore
 import HTTPTypes
 
-/// Capable of being authenticated.
+/// Protocol for types to conform to that can be authenticated
```
gwynne:
```suggestion
/// A protocol to which a type may conform to enable use of that type for authentication.
```

gwynne (approved):
Just some more grammatical and convention nits.

PR #3498, opened 2026-08-04
Wire up TLS support for the server

mob-connection:
TLS configuration for the HTTP server was commented out on Vapor 5's main. This wires it back up on the new swift-http-server based server.

- Add `ServerConfiguration.TLSConfiguration` (`.pemFile` / `.inMemory`) built on swift-certificates types, without exposing any `NIOHTTPServer` types in Vapor's public API.
- Add `ServerConfiguration.tlsConfiguration` and `isTLSEnabled`; fix `addressDescription` to use the correct http/https scheme.
- Map the configuration to `NIOHTTPServer`'s `TransportSecurity` inside `NIOHTTPServerAdapter`.
- Add `ServerTLSTests` covering the pemFile, inMemory and plaintext paths plus the config / addressDescription behaviour.

mTLS and HTTP/2 are intentionally left out of scope as separate follow-ups.


0xTim (changes requested):
Thanks! This is good, just need to add a couple more tests.

Can you add a test for trying to connect to `http` with TLS enabled - that should fail because it can't decrypt the handshake and then in the happy path tests can we verify the cert is the one we're presenting to it?

mob-connection:
> Thanks! This is good, just need to add a couple more tests.
> 
> Can you add a test for trying to connect to `http` with TLS enabled - that should fail because it can't decrypt the handshake and then in the happy path tests can we verify the cert is the one we're presenting to it?

Done! Added both:
   1. Plaintext-to-TLS test (testPlaintextRequestToTLSServerFails): sends a plain http:// request to a TLS-enabled server and asserts it throws — the server can't parse the plaintext as a TLS handshake, so the connection is dropped.
   2. Certificate verification in the happy path (testServerPresentsConfiguredCertificate): the client pins the server's certificate as its only trust root, so the handshake succeeds only if the server presents exactly that certificate — verifying we serve the cert we configured. I embedded a long-lived self-signed cert/key (TLSData) for this, since the existing expired.crt can't pass trust validation.

0xTim (approved):
Ok we can add more tests but I think fix forward rather than holding this up, especially as it's needed of H/2

PR #3501, opened 2026-08-06
Stream response bodies through the server handler

### LINKED ISSUE #3514: Support Streaming Response Bodies

Allow response bodies to be streamed

Currently we load the entire response into memory before writing it out to the response, we should offer a streaming API as well

_No response_

mob-connection:
Streams response bodies straight to the connection instead of buffering them, with backpressure from the transport to the producer. No NIOHTTPServer types leak into the public API.

• Response.Body.init(stream:) now takes `@Sendable` (any ResponseBodyWriter) async throws -> (); the closure writes chunks with await and must not end the stream (the server finishes it, throwing aborts it). ResponseBodyWriter is now AnyObject; the old AsyncBodyStreamWriter / asyncStream: / managedAsyncStream: APIs are removed.
• req.fileio.streamFile and Response.Body.collect() go through the same writer. A throwing closure aborts the stream (swift-http-server#115).
• New StreamingBodyTests (real server) covering buffered/chunked/empty/large/backpressure/bad-length/mid-stream error/client-abort/collect(); FileTests re-enabled on a real server (multi-request cases use one app.test(method: .running) { runner in … }, per #3521); streaming examples added to [routes.swift](file:///Users/oleksandrzhurba/Desktop/vapor/Sources/Development/routes.swift).

Follow-ups (separate PRs): .collect(on:) migration and typed stream reset.


Sources/Vapor/Response/Response+Body.swift
```diff
@@ -292,4 +293,91 @@ extension Response.Body {
...
+        case .error(let error):
+            self.continuation.finish(throwing: error)
+        }
+    }
+}
+
+/// Bridges Vapor's EventLoop/promise-based `BodyStreamWriter` to a pull-based
+/// `AsyncThrowingStream`. Each write is forwarded into the stream and its promise is
+/// fulfilled immediately.
+private final class EventLoopStreamBodyWriter: BodyStreamWriter, Sendable {
```
0xTim: I think we can bypass this entirely - can we write directly to the response writer on the response on the HTTP Server?

0xTim (changes requested):
One big architectural question, some conflicts to fix.

Mainly, we're moving away from any `EventLoop` related, we shouldn't really need anything referencing an event loop (for the most part) in Vapor so I think we should be able to bypass it completely. May require some bigger fundamental changes to how we write responses

Tests/VaporTests/FileTests.swift
```diff
@@ -253,12 +253,12 @@ struct FileTests {
 
             var headerRequest = HTTPFields()
             headerRequest.range = .init(unit: .bytes, ranges: [.within(start: -20, end: 25)])
-            try await app.testing(method: .running).test(.get, "/file-stream", headers: headerRequest) { res async in
```
0xTim, line 256: These need to run against a real server otherwise the streaming is only simulated. Note that there's a bug #3521 so you'll need to do 

```swift
try await app.test(method: .running) { runner in
  try await runner.sendRequest...
  try await runner.sendRequest...
}
```

Rather than multiple `.test` calls

Sources/Vapor/HTTP/Server/ResponseBodyWriter.swift
```diff
@@ -1,24 +1,37 @@
...
-/// and allow Vapor's response body types to write their data to the underlying connection.
-public protocol ResponseBodyWriter {
+/// Implementations of this protocol are provided by the HTTP server layer and allow Vapor's
+/// response body types to write their data straight to the underlying connection. Writes are
+/// backpressured: `write` suspends while the transport can't accept more data, so a streaming
+/// body naturally throttles to the speed of the client.
+///
+/// The protocol is `~Copyable` because concrete writers wrap the server's move-only response
+/// writer; the body-stream closure receives one as ``ResponseBodyStreamWriter``.
+public protocol ResponseBodyWriter: ~Copyable {
```
0xTim, line 14: If we make this `AnyObject` it can hold the reference to the non-copyable writer. That resolves the need to change it to mutating (cause of the current crash in CI) and means we can lose `ResponseBodyStreamWriter`

0xTim (changes requested):
Looking better! A couple of things we should add:

* Examples in routes.swift for chunks, an AsyncSequence, back pressure demo, manual file streaming using `req.fileio.streamFile` and manually writing out chunks from a file reader
* a test for backpressure
* testing bad stream length (there's a pipeline test that we can copy over from)

There's also `.collect(on:)` to update, but I think we can do that in a future PR (feel free to create an issue to track it!)

Sources/Development/routes.swift
```diff
@@ -98,6 +98,49 @@ func routes(_ app: Application) async throws {
         return "Done"
     }
 
+    app.get("stream", "chunks") { _ -> Response in
```
0xTim, line 101: Can we add one more method that adds a `Task.sleep` after each chunk to prove suspension

Sources/Vapor/HTTPNewNew/HTTPServerHandler.swift
```diff
@@ -79,12 +79,53 @@ struct VaporHTTPServerHandler: HTTPServerRequestHandler {
...
 }
+
+/// Bridges Vapor's ``ResponseBodyWriter`` onto the server's move-only response writer.
+///
+/// Each chunk is copied into a `UniqueArray<UInt8>` and forwarded with `await`, so the transport's
+/// backpressure (the socket/HTTP-2 flow-control window) propagates straight to the body-stream
+/// closure — a fast producer suspends while a slow client catches up. The underlying writer is
+/// move-only (`~Copyable`) and ``finish(_:)`` consumes it, so it's stored in an `Optional`: a class
+/// can't move a stored property out in place, and `Optional.take()` is how ``finish(_:)`` moves it out.
+class NIOResponseBodyWriter: ResponseBodyWriter {
```
0xTim: This should be a final class

Sources/Vapor/Utilities/FileIO.swift
```diff
@@ -266,23 +266,35 @@ public struct FileIO: Sendable {
...
-                    do {
-                        for try await chunk in chunks {
-                            try await stream.writeBuffer(chunk)
-                        }
-                    } catch {
-                        throw error
-                    }
-                    try await stream.write(.end)
-                    try await onCompleted(.success(()))
+                handle = try await fileSystem.openFile(forReadingAt: FilePath(path), options: .init())
```
0xTim, line 277: I'd prefer to keep the `readFile` approach to avoid the need to manually close the handle in case it gets removed accidentally. Unless there's a reason why you picked this way

Sources/Vapor/Response/Response+Body.swift
```diff
@@ -229,50 +188,28 @@ extension Response {
...
+}
 
-    init(eventLoop: any EventLoop, byteBufferAllocator: ByteBufferAllocator) {
-        self.buffer = byteBufferAllocator.buffer(capacity: 0)
-        self.eventLoop = eventLoop
-        self.promise = eventLoop.makePromise(of: ByteBuffer.self)
-    }
+/// A ``ResponseBodyWriter`` that accumulates everything written into a `ByteBuffer`, used to
+/// eagerly collect a streaming body instead of forwarding it to the connection.
+private class CollectingBodyWriter: ResponseBodyWriter {
```
0xTim: This needs to be final as well

Sources/Vapor/HTTP/Server/ResponseBodyWriter.swift
```diff
@@ -4,20 +4,25 @@ public import HTTPTypes
...
+public protocol ResponseBodyWriter: AnyObject {
     /// Write a single ByteBuffer.
-    mutating func write(_ buffer: ByteBuffer) async throws
+    func write(_ buffer: ByteBuffer) async throws
     /// Write a sequence of ByteBuffers.
-    mutating func write(contentsOf buffers: some Sequence<ByteBuffer>) async throws
+    func write(contentsOf buffers: some Sequence<ByteBuffer>) async throws
     /// Finish writing the body with optional trailing headers.
-    consuming func finish(_ trailingHeaders: HTTPFields?) async throws
+    func finish(_ trailingHeaders: HTTPFields?) async throws
```
0xTim: I don't think this should be public to avoid users ever having the option to end the stream. We could consider a `setTrailers` function to set them if we want to expose that. That might require more work though and is probably worth a separate issue. So I think we remove this from the public API and keep it internal

Sources/Vapor/Response/Response+Body.swift
```diff
@@ -229,50 +188,28 @@ extension Response {
     }
 }
 
-// Since all buffer mutation is done on the event loop, we can be unchecked here.
-// This removes the need for a lock and performance hits from that
-// Any changes to this type need to be carefully considered
-private final class ResponseBodyCollector: BodyStreamWriter, AsyncBodyStreamWriter, @unchecked Sendable {
+/// Accumulates streamed body chunks into a single `ByteBuffer` for ``Response/Body/collect()``.
+/// Reference-typed so the collecting writer can append to the shared buffer while the stream closure runs.
+private final class ByteBufferAccumulator: @unchecked Sendable {
```
0xTim: Does this need to be unchecked Senable? Where do we mutate the buffer?

0xTim: In fact I don't think we even need this anymore, I think we can just store the buffer directly on the `CollectingBodyWriter`

0xTim (changes requested):
Looking really good! Just a few last tweaks then we're there I think

Sources/Vapor/Utilities/FileIO.swift
mob-connection: > I'd prefer to keep the `readFile` approach to avoid the need to manually close the handle in case it gets removed accidentally. Unless there's a reason why you picked this way

On readFile: there is a reason — it doesn't compile with the streaming writer. readFile's chunk closure is `@Sendable` (NIOFS withFileHandle needs a sending closure), but ResponseBodyWriter is non-Sendable (it wraps the move-only writer), so it can't be captured there. Kept the manual openFile + defer close and documented why — open to alternatives if you'd prefer.

0xTim: Ok makes sense, thanks!

Sources/Vapor/Response/Response+Body.swift
```diff
@@ -158,61 +146,31 @@ extension Response {
...
-        ///   - byteBufferAllocator: The allocator that is preferred when writing data to SwiftNIO
-        public init(asyncStream: @escaping @Sendable (any AsyncBodyStreamWriter) async throws -> (), byteBufferAllocator: ByteBufferAllocator = ByteBufferAllocator()) {
-            self.init(asyncStream: asyncStream, count: -1, byteBufferAllocator: byteBufferAllocator)
-        }
-
-        /// Creates a _managed_ chunked HTTP ``Response`` steam using ``AsyncBodyStreamWriter`` that automtically closes or fails based if the closure throws an error or returns.
+        /// The closure receives a ``ResponseBodyWriter`` and writes chunks to it with `await`. Writes are
+        /// backpressured by the transport, so the closure naturally throttles to the speed of the client.
+        /// It **must not** end the stream — the server appends the final chunk once the closure returns;
+        /// throwing from the closure fails the response.
```
0xTim:
```suggestion
        /// backpressured by the transport, so the closure naturally throttles to the speed of the client.
        /// Throwing from the closure fails the response.
```

PR #3503, opened 2026-08-08
Emit Accept-Ranges for partial file responses

### LINKED ISSUE #3408: It appears a head is missing in 206 Partial Content.
in Vapor (version: 4.119.2):
`vapor/Utilities/FileIO.swift:185`
`vapor/Utilities/FileIO.swift:544`

Codes: 
```swift
// line 185:
box.headers.add(name: .accept, value: contentRange.unit.serialize())

// line 544:
response.headers.add(name: .accept, value: contentRange.unit.serialize())
```

Description:
Only `Accept` header.
The `Accept-Ranges` header is missing.

When I use `Cloudflare` to proxy my `Vapor HTTP service`, without `Accept-Ranges`, Cloudflare cannot recognize it as a 206 response and always returns a 200 response, meaning it does not support resuming interrupted downloads.

I tried adding Accept-Ranges to Nginx, and then Cloudflare returned a 206 response correctly.

I'm not sure if this is a bug; I don't know much about HTTP headers.

I'm just pointing out this issue.

aryansk:
Fixes #3408

### What changed

The partial file-response path now emits the `Accept-Ranges` response header rather than assigning the range unit to the request-oriented `Accept` header. The existing file-header test now asserts `Accept-Ranges: bytes`.

### Validation

- `git diff --check` passed.
- `swift test --filter FileTests` is blocked before compilation in this checkout because the resolved `swift-http-server` dependency requires Swift tools 6.4.0 while the installed toolchain is Swift 6.3.3. No passing test result is claimed.

### DIFF
--- a/Sources/Vapor/Utilities/FileIO.swift
+++ b/Sources/Vapor/Utilities/FileIO.swift
@@ -244,7 +244,7 @@ public struct FileIO: Sendable {
         let byteCount: Int
         if let contentRange = contentRange {
             response.status = .partialContent
-            response.headers[.accept] = contentRange.unit.serialize()
+            response.headers[.acceptRanges] = contentRange.unit.serialize()
             if let firstRange = contentRange.ranges.first {
                 do {
                     let range = try firstRange.asResponseContentRange(limit: Int(fileInfo.size))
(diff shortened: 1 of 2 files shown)


0xTim:
This is resolved in #3554 which has the `bytes` fix (which this PR would still need). I'll close this but feel free if you'd like to fix this in the `vapor4` branch!

PR #3504, opened 2026-08-10
Add Initial Benchmarks

0xTim:
Adds some initial benchmarks to help us measure everything and ensure future changes are performant


Benchmarks/VaporBenchmarks/AuthenticationBenchmarks.swift
```diff
@@ -0,0 +1,137 @@
...
+func authenticationBenchmarks() {
+    Benchmark("auth/login") { benchmark in
+        let request = Request(application: app)
+        let user = BenchUser(id: 1, name: "Vapor", email: "vapor@vapor.codes")
+        for _ in benchmark.scaledIterations {
+            request.auth.login(user)
+        }
+        blackHole(request)
+    } setup: {
+        try await setUpApplication { _ in }
```
MahdiBM, line 14: This caught me off-guard once, so just FYI, these setup/teardown hooks are once-per-benchmark or something. They don't run before each run, for example.

Benchmarks/VaporBenchmarks/Benchmarks.swift
```diff
@@ -0,0 +1,18 @@
+import Benchmark
+
+let benchmarks: @Sendable () -> Void = {
+    Benchmark.defaultConfiguration = .init(
+        metrics: [.instructions, .mallocCountTotal, .wallClock],
```
MahdiBM, line 5: This might be fine for a start, but optimally we can have dedicated malloc count benchmarks. Malloc count affects the others a decent amount. I think the actual optimal setup is to have each one of these as its own benchmark, IIRC even instructions+timing benchmarks affect each other.

Wallclock is questionable since we are not running these on dedicated-core or bare-metal machines. cpu-user is likely better even taking into account that it won't count the time the threads are sleeping (e.g. waiting on some lock).

Benchmarks/VaporBenchmarks/Benchmarks.swift
```diff
@@ -0,0 +1,18 @@
...
+        warmupIterations: 3,
+        scalingFactor: .kilo,
+        maxDuration: .seconds(3)
```
MahdiBM, line 8: This is fine but I personally would have done 5-10 warmup + 5 seconds max duration.

Sources/Vapor/Responder/DefaultResponder.swift
```diff
@@ -102,3 +106,22 @@ extension RouteNotFound: DebuggableError {
...
+extension Application {
+    /// The ``Responder`` that handles requests for this application.
+    ///
+    /// This is the entry point Vapor's HTTP server uses once a request has been parsed off the
+    /// wire, so calling it exercises the same path a live request takes — routing, the middleware
+    /// stack, handler dispatch and response encoding — without needing a socket.
+    ///
+    /// Useful for driving an application in-process: test harnesses, serverless adapters and
+    /// benchmarks.
+    public func makeResponder() -> any Responder {
```
MahdiBM, line 119: Can't tell but do we actually want this as public?

Benchmarks/Package.swift
```diff
@@ -0,0 +1,30 @@
...
+
+let package = Package(
+    name: "vapor-benchmarks",
+    platforms: [
+        .macOS("26.2"),
+    ],
+    dependencies: [
+        .package(name: "vapor", path: ".."),
+        .package(url: "https://github.com/ordo-one/benchmark", from: "1.9.2"),
+    ],
```
MahdiBM, line 12: This is also fine and likely we shouldn't change it, but so you know there are good alternatives, and for whatever reason I had around a year ago, I simply symlink the folders in Benchmarks/ and copy paste the Package.swift too, and just add the extra modifications at the end of Package.swift via `package.whatever = whatever`, so when the main Package.swift is updated, it's also easy to just copy the file over to Benchmarks/ again.

Sources/Vapor/Responder/DefaultResponder.swift
0xTim: We need to unfortunately, now SPI but we need a hook somewhere

PR #3510, opened 2026-08-11
Add HTTP/2 support to the server

### LINKED ISSUE #3378: Support HTTP/2

Add H/2 back

H/2 doesn't currently work in the latest main for Vapor 5, we need to support it

_No response_

mob-connection:
Adds HTTP/2 support to the new server backend, keeping the underlying `swift-http-server` types out of Vapor's public API (fixes #3378).

Enable HTTP/2 (requires TLS — negotiated over ALPN):

    app.serverConfiguration.tlsConfiguration = .pemFile(certificateChainPath: "...", privateKeyPath: "...")
    app.serverConfiguration.httpVersions = [.http1_1, .http2(config: .defaults)]

- New Vapor-owned `ServerConfiguration.HTTP2` and `ServerConfiguration.HTTPVersion`; the adapter translates them to NIOHTTPServer types internally so no server types leak.
- New `httpVersions` on `ServerConfiguration`, defaulting to `[.http1_1]`.
- Requesting HTTP/2 without a `tlsConfiguration` throws a clear error instead of failing obscurely.

### DIFF
--- a/Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
@@ -1,8 +1,10 @@
 extension ServerConfiguration {
     /// HTTP/2 specific configuration.
     ///
-    /// HTTP/2 is only negotiated over TLS (via ALPN). Adding an ``ServerConfiguration/HTTPVersion/http2(config:)``
-    /// version without also setting ``ServerConfiguration/tlsConfiguration`` will cause the server to fail to start.
+    /// HTTP/2 is only negotiated over TLS (via ALPN). If an ``ServerConfiguration/HTTPVersion/http2(config:)``
+    /// version is added to ``ServerConfiguration/httpVersions`` without also setting a
+    /// ``ServerConfiguration/tlsConfiguration``, the server throws an error when it starts (rather than
+    /// serving over plaintext), so callers can catch it and decide how to handle it.
     public struct HTTP2: Sendable, Hashable {
         /// The maximum frame size to be used in an HTTP/2 connection.
         public var maxFrameSize: Int
@@ -54,12 +56,16 @@ extension ServerConfiguration {
             self.gracefulShutdown = gracefulShutdown
         }
 
+        /// `2^14` (16 KiB) — the smallest, and default, `SETTINGS_MAX_FRAME_SIZE` defined by HTTP/2 (RFC 9113 §6.5.2).
         @inlinable
         static var defaultMaxFrameSize: Int { 1 << 14 }
 
+        /// `2^16 - 1` (65 535) — the default `SETTINGS_INITIAL_WINDOW_SIZE` (flow-control window) defined by
+        /// HTTP/2 (RFC 9113 §6.9.2).
         @inlinable
         static var defaultTargetWindowSize: Int { (1 << 16) - 1 }
 
+        /// `100` — the minimum concurrent streams RFC 9113 (§8.2.2) recommends endpoints allow; used as the default.
         @inlinable
         static var defaultMaxConcurrentStreams: Int { 100 }
 
@@ -110,8 +116,14 @@ extension ServerConfiguration {
             Self.init(version: .http2(config: config))
         }
 
-        /// Two values are equal if they represent the same protocol version, regardless of any differences in the
-        /// HTTP/2 configuration.
+        /// Equality is by protocol version only: two values are equal when they represent the same HTTP
+        /// version, ignoring any associated configuration.
+        ///
+        /// This is intentional, and the reason `==`/`hash` are written by hand instead of synthesised. It
+        /// makes a `Set<HTTPVersion>` behave as *the set of supported protocols*: it holds at most one entry
+        /// per version, so the same protocol can't be added twice with conflicting configuration. Synthesised
+        /// conformances would fold the associated config into identity and allow such duplicates. This mirrors
+        /// `NIOHTTPServerConfiguration.HTTPVersion`, which does the same.
         public static func == (lhs: Self, rhs: Self) -> Bool {
             switch (lhs.version, rhs.version) {
             case (.http1_1, .http1_1), (.http2, .http2):
@@ -122,7 +134,7 @@ extension ServerConfiguration {
             }
         }
 
-        /// Hashes by protocol version only, consistent with the `Equatable` conformance.
+        /// Hashes by protocol version only, consistent with the `Equatable` conformance above.
         public func hash(into hasher: inout Hasher) {
             switch self.version {
             case .http1_1:
--- a/Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
@@ -1,0 +1,136 @@
+extension ServerConfiguration {
+    /// HTTP/2 specific configuration.
+    ///
+    /// HTTP/2 is only negotiated over TLS (via ALPN). Adding an ``ServerConfiguration/HTTPVersion/http2(config:)``
+    /// version without also setting ``ServerConfiguration/tlsConfiguration`` will cause the server to fail to start.
+    public struct HTTP2: Sendable, Hashable {
+        /// The maximum frame size to be used in an HTTP/2 connection.
+        public var maxFrameSize: Int
+
+        /// The target window size for the connection.
+        ///
+        /// - Note: This is also used as the initial window size for the connection.
+        public var targetWindowSize: Int
+
+        /// The maximum number of concurrent streams permitted on the connection.
+        public var maxConcurrentStreams: Int
+
+        /// The graceful shutdown configuration.
+        public var gracefulShutdown: GracefulShutdownConfiguration
+
+        /// Configuration options for HTTP/2 graceful shutdown behavior.
+        public struct GracefulShutdownConfiguration: Sendable, Hashable {
+            /// The maximum amount of time that the connection has to close gracefully.
+            ///
+            /// When `nil`, no time limit is enforced on the graceful shutdown process.
+            public var maximumGracefulShutdownDuration: Duration?
+
+            /// Creates a graceful shutdown configuration.
+            ///
+            /// - Parameter maximumGracefulShutdownDuration: The maximum amount of time the connection has to close
+            ///   gracefully. When `nil`, no time limit is enforced.
+            public init(maximumGracefulShutdownDuration: Duration? = nil) {
+                self.maximumGracefulShutdownDuration = maximumGracefulShutdownDuration
+            }
+        }
+
+        /// Creates a new HTTP/2 configuration.
+        ///
+        /// - Parameters:
+        ///   - maxFrameSize: The maximum frame size to be used in a connection. Defaults to `2^14`.
+        ///   - targetWindowSize: The target window size for a connection, also used as the initial window size.
+        ///     Defaults to `2^16 - 1`.
+        ///   - maxConcurrentStreams: The maximum number of concurrent streams permitted. Defaults to `100`.
+        ///   - gracefulShutdown: The graceful shutdown configuration.
+        public init(
+            maxFrameSize: Int = Self.defaultMaxFrameSize,
+            targetWindowSize: Int = Self.defaultTargetWindowSize,
+            maxConcurrentStreams: Int = Self.defaultMaxConcurrentStreams,
+            gracefulShutdown: GracefulShutdownConfiguration = .init()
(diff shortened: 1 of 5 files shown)


Sources/Vapor/HTTPNewNew/NIOHTTPServerAdapter.swift
```diff
@@ -63,11 +67,37 @@ final class NIOHTTPServerAdapter: Server, Sendable {
...
+                        gracefulShutdown: .init(
+                            maximumGracefulShutdownDuration: config.gracefulShutdown.maximumGracefulShutdownDuration
+                        )
+                    )
+                ))
+            }
+        }
+
+        // HTTP/2 is negotiated via ALPN, which requires TLS. Over plaintext, only HTTP/1.1 is allowed.
+        if !self.application.serverConfiguration.isTLSEnabled {
```
0xTim: I think we need an `h2c` flag somewhere that will override this check. H2c (HTTP/2 over cleartext) is not allowed by most browsers but there are legitimate use cases for it so I think we should add a way to enable it

Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
```diff
@@ -0,0 +1,136 @@
+extension ServerConfiguration {
+    /// HTTP/2 specific configuration.
+    ///
+    /// HTTP/2 is only negotiated over TLS (via ALPN). Adding an ``ServerConfiguration/HTTPVersion/http2(config:)``
+    /// version without also setting ``ServerConfiguration/tlsConfiguration`` will cause the server to fail to start.
```
0xTim: We should clarify that this throws an error, the user might handle that in different ways

Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
```diff
@@ -0,0 +1,136 @@
...
+            self.maxConcurrentStreams = maxConcurrentStreams
+            self.gracefulShutdown = gracefulShutdown
+        }
+
+        @inlinable
+        static var defaultMaxFrameSize: Int { 1 << 14 }
+
+        @inlinable
+        static var defaultTargetWindowSize: Int { (1 << 16) - 1 }
+
```
0xTim, line 67: Let's explain why these magic numbers have been picked

Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
```diff
@@ -0,0 +1,136 @@
...
+        /// Creates a new HTTP/2 configuration.
+        ///
+        /// - Parameters:
+        ///   - maxFrameSize: The maximum frame size to be used in a connection. Defaults to `2^14`.
+        ///   - targetWindowSize: The target window size for a connection, also used as the initial window size.
+        ///     Defaults to `2^16 - 1`.
+        ///   - maxConcurrentStreams: The maximum number of concurrent streams permitted. Defaults to `100`.
+        ///   - gracefulShutdown: The graceful shutdown configuration.
+        public init(
```
0xTim, line 47: This is probably a good place to enable h2c

Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        /// The HTTP/2 protocol version.
+        ///
+        /// - Parameter config: The configuration to use for HTTP/2 connections.
+        public static func http2(config: HTTP2) -> Self {
+            Self.init(version: .http2(config: config))
+        }
+
+        /// Two values are equal if they represent the same protocol version, regardless of any differences in the
```
0xTim: Why are we ignoring configs for equality?

Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
```diff
@@ -0,0 +1,136 @@
...
+            case (.http1_1, .http1_1), (.http2, .http2):
+                return true
+
+            default:
+                return false
+            }
+        }
+
+        /// Hashes by protocol version only, consistent with the `Equatable` conformance.
+        public func hash(into hasher: inout Hasher) {
```
0xTim, line 129: Can this not be auto-synthesised?

Sources/Vapor/HTTPNewNew/NIOHTTPServerAdapter.swift
```diff
@@ -7,6 +7,10 @@ import Logging
 enum NIOHTTPServerAdapterError: Error {
     /// The underlying server reported that it was listening but exposed no addresses.
     case noListeningAddress
+
+    /// HTTP/2 was requested without TLS. HTTP/2 is only negotiated over TLS (via ALPN),
```
0xTim: See h2c comment above - we can keep this error for when it's not enabled but clarify the comment

Sources/Vapor/HTTPNewNew/NIOHTTPServerAdapter.swift
```diff
@@ -63,11 +67,37 @@ final class NIOHTTPServerAdapter: Server, Sendable {
             transportSecurity = .plaintext
         }
 
+        var supportedHTTPVersions = Set<NIOHTTPServerConfiguration.HTTPVersion>()
+        for httpVersion in self.application.serverConfiguration.httpVersions {
```
0xTim, line 75: We should check we have at least one version

Tests/VaporTests/HTTPServerConfigurationTests.swift
```diff
@@ -0,0 +1,94 @@
+@testable import Vapor
+import VaporTesting
+import ServiceLifecycle
```
0xTim, line 3: I think this is currently unused

0xTim (changes requested):
Thanks! Some comments inline

We also need to add a test for the H2 happy path where the server starts and we can call it with the client and assert the response is an H2 response. We also have some existing tests commented out that should cover H2 so worth looking at those to pull in or replace

Sources/Vapor/HTTPNewNew/NIOHTTPServerAdapter.swift
mob-connection: > I think we need an `h2c` flag somewhere that will override this check. H2c (HTTP/2 over cleartext) is not allowed by most browsers but there are legitimate use cases for it so I think we should add a way to enable it

Agreed h2c has valid uses, but it's blocked downstream: swift-http-server doesn't support HTTP/2 over plaintext yet — its config rejects it and `TransportSecurity` only has `.plaintext`/`.tls`/`.mTLS`. The maintainers confirm h2c is planned but not implemented (see the discussion in swift-server/swift-http-server#107: "…plaintext possible over HTTP/1.1 (and HTTP/2 once we add support for h2c)").

So I'll keep the pre-check for now (rewording the error/comment to say h2c isn't supported yet) and wire up the opt-in once it lands upstream.

Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
mob-connection: > We should clarify that this throws an error, the user might handle that in different ways

Done — reworded the doc to make it explicit that the server throws an error at startup (instead of serving over plaintext), so callers can catch it and decide how to handle it.

Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
mob-connection: > This is probably a good place to enable h2c

Same as the h2c thread — deferring until `swift-http-server` supports cleartext HTTP/2 (it currently rejects plaintext + HTTP/2; maintainers say it's planned, swift-server/swift-http-server#107). When it lands, this init is exactly where I'd add the opt-in.

Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
mob-connection: > Why are we ignoring configs for equality?

Kept it version-only on purpose (added a comment explaining why): the Set is meant to be "the set of supported protocols", so it should hold at most one entry per version and not allow the same protocol twice with conflicting config. Synthesised ==/hash would fold the associated config into identity and permit duplicates. This mirrors NIOHTTPServerConfiguration.HTTPVersion, which hand-writes the same version-only conformance.

Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
mob-connection: > Can this not be auto-synthesised?

No — auto-synthesis would include the associated config in identity, which breaks the version-only semantics (a Set could then hold two HTTP/2 entries with different configs). Same reason NIOHTTPServerConfiguration.HTTPVersion writes it by hand. I added a comment on == explaining this; hash just stays consistent with it.

Tests/VaporTests/HTTPServerConfigurationTests.swift
mob-connection: > I think this is currently unused

It's actually used now — `app.server.run()` in the preflight tests needs it (`run()` comes from the `Service` protocol in ServiceLifecycle).

Sources/Vapor/HTTP/Server/HTTPServerConfiguration+HTTP2Configuration.swift
```diff
@@ -0,0 +1,148 @@
...
+    /// An HTTP protocol version the server can advertise and serve.
+    ///
+    /// Use ``http1_1`` and ``http2(config:)`` to build the set passed to ``ServerConfiguration/httpVersions``.
+    public struct HTTPVersion: Sendable, Hashable {
+        /// The underlying protocol version, carrying the HTTP/2 configuration when applicable.
+        enum Version: Sendable, Hashable {
+            case http1_1
+            case http2(config: HTTP2)
+
+            var http2Config: HTTP2? {
```
0xTim: Does anything actually use this?

Sources/Vapor/HTTPNewNew/NIOHTTPServerAdapter.swift
```diff
@@ -63,11 +71,41 @@ final class NIOHTTPServerAdapter: Server, Sendable {
...
+                        maxConcurrentStreams: config.maxConcurrentStreams,
+                        gracefulShutdown: .init(
+                            maximumGracefulShutdownDuration: config.gracefulShutdown.maximumGracefulShutdownDuration
+                        )
+                    )
+                ))
+            }
+        }
+
+        if self.application.serverConfiguration.httpVersions.isEmpty {
```
0xTim: Minor stylistic change, can you invert this to a guard to make it clear this is an early exit path?

Sources/Vapor/HTTPNewNew/NIOHTTPServerAdapter.swift
```diff
@@ -63,11 +71,41 @@ final class NIOHTTPServerAdapter: Server, Sendable {
...
+            throw NIOHTTPServerAdapterError.noHTTPVersionsSpecified
+        }
+
+        // HTTP/2 is negotiated via ALPN, which requires TLS. Over plaintext, only HTTP/1.1 is allowed.
+        if !self.application.serverConfiguration.isTLSEnabled {
```
0xTim: Minor stylistic change, can you invert this to a guard as well. For now, I'd merge the two as well into one guard statement until we add H3

0xTim (changes requested):
A couple of stylistic changes then we're good to go, thanks!

0xTim:
A quick heads up - https://github.com/vapor/vapor/blob/main/Sources/Vapor/HTTPNewNew/HTTPServerHandler.swift#L62 will need updating as well (and we should add a test to check)

mob-connection:
> A quick heads up - https://github.com/vapor/vapor/blob/main/Sources/Vapor/HTTPNewNew/HTTPServerHandler.swift#L62 will need updating as well (and we should add a test to check)

The negotiated HTTP version isn't reachable from a request handler — it's on the internal `connectionContext`, and `RequestContext` only exposes addresses + peer cert. Opened an issue to expose it: https://github.com/swift-server/swift-http-server/issues/113

Once it's exposed upstream it's a one-line map onto `Request.version` + a test here. Happy to do that swift-http-server PR too if you're good with the approach.

0xTim (approved):
Thanks!

PR #3511, opened 2026-08-11
Turn on InternalImportsByDefault for Non-Vapor Targets

0xTim:
Turn on InternalImportsByDefault for everything but the main Vapor target as that's a bigger piece of work to be done separately

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -238,6 +238,7 @@ let package = Package(
 )
 
 var swiftSettings: [SwiftSetting] { [
+    .strictMemorySafety(),
     .enableUpcomingFeature("ExistentialAny"),
 //    .enableUpcomingFeature("InternalImportsByDefault"),
     .enableUpcomingFeature("MemberImportVisibility"),
--- a/Package.swift
+++ b/Package.swift
@@ -147,7 +147,7 @@ let package = Package(
                 .product(name: "HTTPTypes", package: "swift-http-types"),
                 .product(name: "Algorithms", package: "swift-algorithms")
             ],
-            swiftSettings: swiftSettings
+            swiftSettings: swiftSettings + [.enableUpcomingFeature("InternalImportsByDefault")]
         ),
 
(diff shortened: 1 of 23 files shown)


Package.swift
```diff
@@ -232,14 +232,15 @@ let package = Package(
...
+            swiftSettings: swiftSettings + [.enableUpcomingFeature("InternalImportsByDefault")]
         ),
     ]
 )
 
 var swiftSettings: [SwiftSetting] { [
+    .strictMemorySafety(),
     .enableUpcomingFeature("ExistentialAny"),
-    //.enableUpcomingFeature("InternalImportsByDefault"),
+//    .enableUpcomingFeature("InternalImportsByDefault"),
```
MahdiBM, line 243: should remove it I guess? or we can keep it too it's not that big of a deal.

Sources/VaporTesting/Utilities.swift
```diff
@@ -1,4 +1,5 @@
-import NIOCore
+#warning("We should remove this when we don't need ByteBuffer")
+public import NIOCore
```
MahdiBM, line 2: Hopefully these warnings don't make it to the users?

for this one specifically, you could also do `public import struct NIOCore.ByteBuffer` to assert that indeed only `ByteBuffer` is needed in this file.

I guess these all might need a cleanup in another PR?

MahdiBM (approved):

Sources/VaporTesting/Utilities.swift
0xTim: They will make it to the user but will be removed before we hit a future beta. Essentially we need to remove `ByteBuffer` from the API once the bag of bytes type finally comes into fruition

Package.swift
0xTim: I'll add it in a future PR when the `vapor` adopts it, then all the targets will be using it and we can add it back

PR #3516, opened 2026-08-15
Add support for reloading TLS certificates

### LINKED ISSUE #3505: Support Reloading TLS Certs

Add ability to support reloading TLS certs on the fly alongside file/in-memory options

#3498 Added support for TLS to the new server for loading certs from disk or loading them from memory. The HTTP server also supports a `CertificateReloader`, which we should offer the ability to wire up as well

_No response_

ayush-that:
Fixes #3505

Adds `ServerConfiguration.TLSConfiguration.reloading(_:)`, which takes a `CertificateReloader` and passes it to the HTTP server, so rotated certificates are picked up without restarting the server.

```swift
let reloader = try TimedCertificateReloader.makeReloaderValidatingSources(
    refreshInterval: .seconds(300),
    certificateSource: .init(location: .file(path: "/etc/ssl/cert.pem"), format: .pem),
    privateKeySource: .init(location: .file(path: "/etc/ssl/key.pem"), format: .pem)
)
app.serverConfiguration.tlsConfiguration = .reloading(reloader)
app.addService(reloader)
```

- Running the reloader is the caller's responsibility. Vapor wires it into the server but does not start its task.
- Startup fails when the reloader has no credentials.
- When a reload fails, the server keeps serving the previous certificate and logs the failure.


0xTim (changes requested):
Ok I apologise, I misread the HTTP Server APIs. Refresh interval can be hooked up like it is, but we should also offer the ability to pass a `CertificateReloader` to the config so people can pass their own reloader for whatever they need to do. Note the API should comment that it's the callers responsibility to run the task in that case

0xTim (changes requested):
Ok sorry for the churn! Two things:
* can you rebase onto the latest changes
* I think we should drop the `refreshInterval` changes - just keep the `CertificateReloader` API as the only one. That simplifies the code we need to support it

Other than that it all looks good!

Sources/Vapor/Server/ServerConfiguration+TLSConfiguration.swift
```diff
@@ -19,5 +21,17 @@ extension ServerConfiguration {
         public static func pemFile(certificateChainPath: String, privateKeyPath: String) -> Self {
             Self.init(source: .pemFile(certificateChainPath: certificateChainPath, privateKeyPath: privateKeyPath))
         }
+
+        /// TLS credentials supplied by a `CertificateReloader`.
+        ///
+        /// The caller is responsible for running the reloader. Vapor wires it into the HTTP server
+        /// but does not start its task. The reloader must already hold valid credentials when the
+        /// server starts or startup fails; for `TimedCertificateReloader`, create it with
+        /// `makeReloaderValidatingSources(...)` and call `run()`.
```
0xTim:
```suggestion
        /// `makeReloaderValidatingSources(...)` and call `run()` or use `app.addService(reloader)`
```

0xTim (approved):
Thanks!

PR #3538, opened 2026-08-26
Remove NIO from Response's Public API

0xTim:
Removes NIO from the `Response.Body` public API 🎉

This removes the `ByteBuffer` usage in favour of Data/String where it makes sense and streaming as a first class citizen with easier APIs to use so we no longer need it. Also removes `ByteBufferAllocator` since we're not creating `ByteBuffer`s anymore.

We still have an internal reprensentation which we need for the `ContentContainer` but I'll migrate that next in a separate PR as it touches Request as well.

Also provide `Span` helpers for consuming the streaming response body

### DIFF
--- a/Sources/Vapor/HTTP/Server/ResponseBodyWriter.swift
+++ b/Sources/Vapor/HTTP/Server/ResponseBodyWriter.swift
@@ -1,6 +1,3 @@
-#warning("Make this internal")
-public import NIOCore
-
 /// Protocol for writing HTTP response bodies.
 ///
 /// Implementations of this protocol are provided by the HTTP server layer and allow Vapor's
@@ -13,10 +10,6 @@ public import NIOCore
 /// is exposed: concluding the response is the server's responsibility, so a body-stream closure
 /// can never end the stream itself.
 public protocol ResponseBodyWriter: AnyObject {
-    /// Write a single ByteBuffer.
-    func write(_ buffer: ByteBuffer) async throws
-    /// Write a sequence of ByteBuffers.
-    func write(contentsOf buffers: some Sequence<ByteBuffer>) async throws
     /// Write a sindle chunk of bytes
     func write(_ bytes: RawSpan) async throws
 }
@@ -48,18 +41,6 @@ extension ResponseBodyWriter {
     public func write(contentsOf chunks: some Sequence<some Sequence<UInt8>>) async throws {
         for chunk in chunks {
             try await self.write(chunk)
-        }
-    }
-
-    // Internal helper to map to what NIOHTTPServer wants
-    func write(_ buffer: ByteBuffer) async throws {
-        try await self.write(buffer.readableBytesSpan)
-    }
-
-    #warning("Remove")
-    public func write(contentsOf buffers: some Sequence<ByteBuffer>) async throws {
-        for buffer in buffers {
-            try await write(buffer)
         }
     }
 }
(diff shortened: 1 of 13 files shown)


Sources/Vapor/HTTP/Server/ResponseBodyWriter.swift
```diff
@@ -13,17 +10,37 @@ public import NIOCore
...
+        try await self.write(string.utf8Span.span.bytes)
+    }
+
+    /// Write a sequence of bytes
+    /// This will copy the bytes into contiguous storage first. Use the `Span` APIs to avoid copies
+    @inlinable
+    public func write(_ bytes: some Sequence<UInt8>) async throws {
+        let contiguous = ContiguousArray(bytes)
+        try await self.write(contiguous.span.bytes)
+    }
```
MahdiBM, line 44: Hmm there is `seq.withContiguousStorageIfAvailable { $0.span }` but unfortunately it doesn't support async, so that's an issue for it to be used here.

Considering the API is used a lot, I would encourage our own (Vapor's) impl of the protocol to do actually use that if possible (the writer can do non-async). It's a simple but quite effective win.

Sources/Vapor/Response/Response+Body.swift
```diff
@@ -35,6 +35,71 @@ extension Response {
...
+            case .buffer(let buffer):
+                try await body(buffer.readableBytesSpan)
+            case .data(let data):
+                try await body(data.span.bytes)
+            case .string(let string):
+                try await body(string.utf8Span.span.bytes)
+            case .staticString(let staticString):
+                try await unsafe body(
+                    RawSpan(_unsafeStart: staticString.utf8Start, byteCount: staticString.utf8CodeUnitCount)
+                )
```
MahdiBM, line 69: Again there is `staticString.withUTF8Buffer { $0.span }` but it neither allows throwing nor async. The code here is correct though and doesn't copy or something so the code is fine ... I guess StaticString should just have a direct `.span`/`.utf8Span`, there is no reason why not. A code similar to yours.

MahdiBM (approved):

Sources/Vapor/Response/Response+Body.swift
0xTim: Yeah SF-0042 is supposed to bring Foundation/Span APIs but that's gonna take 7 months to land into Swift 😒

Sources/Vapor/HTTP/Server/ResponseBodyWriter.swift
0xTim: Good idea, will add it

Sources/Vapor/Response/Response+Body.swift
```diff
@@ -267,4 +267,15 @@ private final class CollectingBodyWriter: ResponseBodyWriter {
...
+    /// it into a `ContiguousArray` first, as the protocol's default implementation must.
+    func write(_ bytes: some Sequence<UInt8>) async throws {
+        let borrowed: Void? = bytes.withContiguousStorageIfAvailable { buffer in
+            unsafe self.data.append(contentsOf: buffer)
+        }
+        if borrowed == nil {
+            self.data.append(contentsOf: bytes)
+        }
+    }
 }
```
MahdiBM, line 277:
```suggestion
    func write(_ bytes: some Sequence<UInt8>) async throws {
        self.data.append(contentsOf: bytes)
    }
```

`Data` already knows to try `withContiguousStorageIfAvailable` first. 
https://github.com/swiftlang/swift-foundation/blob/1335fd1b9dbee513f4f2b06de1288968f78b7957/Sources/FoundationEssentials/Data/Data.swift#L723


Edit: not sure why GitHub shows the "suggestion" like this hmm.

MahdiBM (approved):
Nice.

Sources/Vapor/Response/Response+Body.swift
0xTim: Let's see if it works!

MahdiBM: It didn't but I've already fixed it :).

PR #3542, opened 2026-08-27
Add Basic wrk harness

0xTim:
Just a simple harness to get raw numbers, benchmarks will cover actual deterministic numbers

### DIFF
--- a/Performance/Sources/PerformanceServer/main.swift
+++ b/Performance/Sources/PerformanceServer/main.swift
@@ -2,12 +2,6 @@ import Foundation
 import NIOCore
 import Vapor
 
-// A minimal Vapor app serving fixed-size responses, for load testing with `wrk` (see run-wrk.sh).
-//
-// Every route returns a constant payload built once at start-up, so a run measures Vapor's
-// request/response path rather than whatever work a handler happens to do. Bind address is fixed
-// (not port 0) so the load generator knows where to connect.
-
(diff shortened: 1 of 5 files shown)


Performance/Sources/PerformanceServer/main.swift
```diff
@@ -0,0 +1,48 @@
...
+struct Payload: Content {
+    var id: Int
+    var name: String
+    var tags: [String]
+}
+
+// Static responses for performance
+let small = String(repeating: "x", count: 1024)
+let large = String(repeating: "x", count: 64 * 1024)
+let chunk = String(repeating: "y", count: 1024)
```
MahdiBM, line 18: These are not static for the record, they'll be generated at run time.
For better performance numbers you can just write them out. Or write something weird like:
```swift
@_transparent // or @inline(always) with 6.3+, or @inline(__always) with any Swift
public func x1024() -> String {
    /// a == 16 of x
    let a = "xxxxxxxxxxxxxxxx" 
    let b = a + a, c = b + b, d = c + c
    let e = d + d, f = e + e, g = f + f
    return g
}
```

MahdiBM (commented):
Looks ok although the CI is not doing anything as mentioned in it.

Performance/Sources/PerformanceServer/main.swift
MahdiBM: oppsie a little bit late.

PR #3543, opened 2026-08-27
Stream request bodies through the server handler

mob-connection:
Streams request bodies through the server handler instead of eagerly buffering, mirroring the response-side `ResponseBodyWriter`. The body is pulled off the connection on demand with backpressure.

## API
- `req.body.withReader { reader in reader.read { span, isEnd in … } }` — drive the read loop yourself, one chunk at a time.
- `req.body.forEachChunk { span in … }` — convenience for iterating every chunk.
- `collect(max:)` / `body.data` / `content.decode` still work: a stream is collected on demand and cached, so later reads see the same bytes.
- Bytes come as a borrowed `RawSpan`, valid only for the call — copy what you keep. Route strategy unchanged (`.collect` default, `.stream` reads directly).

## Ownership & Sendable
The move-only server reader lives behind a `Mutex` in a checked-`Sendable` holder (no `@unchecked`); `Request` stays `Sendable`. The body is single-consumer: reading it from two tasks at once throws `RequestBodyAlreadyBeingRead` (500), not a silent end-of-body. The non-escapable reader makes "read after the request moved on" a compile error.

## DoS hardening
- `ServerConfiguration.maxDrainBytes` (default 16 KB): an unread body is drained only up to this budget, else the connection closes (GET/HEAD budget 0).
- `collect(max:)` rejects an over-limit declared `Content-Length` with 413 before reading any bytes.

## Removed
- `Request.Body: AsyncSequence`, legacy EventLoop `Request.BodyStream` / `BodyStreamWriter`, and `PipelineTests` (superseded).

Tests cover streaming read/echo, empty & large bodies, `content.decode`, backpressure, keep-alive after partial/throwing reads, concurrent-read rejection, drain budget, and `Content-Length` 413. swift-http-server types stay `package`/internal.


Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
```diff
@@ -27,21 +27,9 @@ struct VaporHTTPServerHandler: HTTPServerRequestHandler {
...
-                    bodyBuffer.writeBytes(chunk.span.bytes)
-                }
-                if trailers != nil {
-                    reachedEndOfBody = true
-                }
-            }
-        }
+        // 1. Wrap the request body in a lazy pull-based stream (no eager collection).
+        let bodyStream = RequestBodyStream(reader: consume reader)
+        defer { try? await bodyStream.drain() }
```
0xTim: This jumps out at me - it seems we will consume the body of a request for all requests? Which means that even if we have a route we don't know about (404) we'll consume the entire body of the request. This would enable anyone to perform a DoS attack against the server pretty easily unless I'm missing something?

0xTim (commented):
Will give it a proper review in the morning but one initial question

0xTim (changes requested):
Ok I've been back and forth a lot on this and tried out some spikes. I think we should give up the async sequence for now. It provides a nice API, but we give up a lot of the safety and guarantees that users would expect and that the HTTP server tries to push. It also means we can push most of the common errors (like double reads) to a compile time error. So I think we should essentially mirror what the HTTP Server provides and just ferry the body through, which gives us all the guarantees and performance. We can ensure the `decode` function on content still consumes the body up to a limit to make it easy and the collected body works for non streaming cases. Then add the collect/string/data helpers that `Response.Body` has to make it easy to use.

So essentially we can define a `RequestBodyReader` that looks something like

```swift
public struct RequestBodyReader: ~Copyable, ~Escapable {
    public mutating func read<R>(_ body: (RawSpan, Bool) async throws -> R) async throws -> R
}
```

Then in the body we can do something like

```swift
extension Request {
    public struct Body: Sendable {
        /// Lends a reader for the duration of the closure
        public func withReader<R>(_ body: (inout RequestBodyReader) async throws -> R) async throws -> R
        /// Drives the loop for you. What most handlers want.
        public func withBytes(_ body: (RawSpan) async throws -> Void) async throws
        /// Buffers into memory and caches, so the body drains once.
        public func collect(max: Int? = 1 << 14) async throws -> ByteBuffer?
    }
}
```

This gives us the ownership guarantees and means me only copy the data from the span when we actually need it.

Then from an API point of view we can do things like:

```swift
// Stream a file
app.on(.post, "upload", body: .stream) { req -> String in
    try await FileSystem.shared.withFileHandle(
        forWritingAt: FilePath("/tmp/upload.bin"), options: .newFile(replaceExisting: true)
    ) { handle in
        var writer = handle.bufferedWriter()
        var total: Int64 = 0
        try await req.body.withBytes { bytes in
            total += try await writer.write(contentsOf: bytes.withUnsafeBytes { unsafe Array($0) })
        }
        try await writer.flush()
  }
}

// Read the body into `Data`
var data = Data()
try await req.body.withBytes { bytes in
    bytes.withUnsafeBytes { unsafe data.append(contentsOf: $0) }
}

// Use collect to collect into `ByteBuffer`
let buffer = try await req.body.collect(max: 10 << 20) ?? ByteBuffer()

// Decode with content
try await req.content.decode(Payload.self)

// Act as a proxy
Response(body: .init(stream: { writer in
    try await req.body.withBytes { try await writer.write($0) }
}))
```

I think if we can hit those APIs with the ownership guarantees that would be great.

Sources/Development/routes.swift
```diff
@@ -22,39 +22,18 @@ func routes(_ app: Application) async throws {
...
+        // Consume the streamed body slowly to demonstrate backpressure: sleeping between
+        // reads keeps memory flat because the server stops pulling more of the body until
+        // this loop asks for the next chunk.
+        var total = 0
+        try await req.body.forEachChunk { buffer in
+            try await Task.sleep(nanoseconds: 1_000_000_000)
+            total += buffer.byteCount
+        }
+        return total.description
+    }
```
0xTim, line 36: Nice

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
```diff
@@ -194,6 +189,151 @@ final class NIOResponseBodyWriterStorage {
...
 }
 
+/// Holds the server's move-only request `Reader` and ferries the body to user code — the request-side
+/// mirror of ``NIOResponseBodyWriterStorage``. It pulls one part at a time (so backpressure reaches the
+/// producer) and hands the bytes out as a borrowed ``RawSpan``, copying nothing until user code keeps it.
+///
+/// `@unchecked Sendable` because the move-only, non-Sendable `Reader` can't sit behind a lock: safety
+/// rests on the contract that **a body is read by a single task, once** — never concurrently. `Request`
+/// is `Sendable` and may cross tasks, but its body must not be read from two of them.
+package final class RequestBodyStream: @unchecked Sendable {
```
0xTim: We need to find a way for this to not be `@unchecked Sendable`. I don't think we need this if we get the ownership right

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
```diff
@@ -194,6 +189,151 @@ final class NIOResponseBodyWriterStorage {
...
+    init(_ buffer: ByteBuffer?) {
+        self.buffer = buffer
+    }
+}
+
+/// The server's concrete ``RequestBodyReader`` — a borrowed, non-escapable view onto the request body,
+/// the mirror of ``NIOResponseBodyWriter``. Lent only for a ``Request/Body/withReader(_:)`` closure;
+/// being `~Escapable` it can't be stored, so "read the body twice" is a compile-time error. Each
+/// ``read(_:)`` hands the next part out as a borrowed ``RawSpan``, copying nothing until user code keeps it.
+struct NIORequestBodyReader: RequestBodyReader, ~Escapable {
```
0xTim, line 376: I think this needs to be `~Copyable` so "read the body twice is a compile-time error" is actually true

0xTim (changes requested):
Definitely in the right direction! Some initial comments

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
mob-connection: > We need to find a way for this to not be `@unchecked Sendable`. I don't think we need this if we get the ownership right

I gave the ownership route a proper spike to see how far it goes. What I did: took the reader out of the Sendable Request and threaded it as an owned `RequestBody: ~Copyable` through `Responder.respond(to:body:)` / `Middleware.respond(to:body:chainingTo:)`, so `Request` only holds `.none/.collected`.

That part works — `@unchecked` is gone and single-consumer is compiler-enforced. But once the body no longer lives on `Request`, I hit these:

• Streaming proxy — Response(body: .init(stream: { writer in req.body… })) can't capture a move-only inout in the @`Sendable @escaping` response-body closure.
• req.content.decode / req.body / Validations on `.stream` routes — they read the body off `Request`, so they break (6 test failures, all this root cause). To keep them working I'd have to thread the body owner through the whole content/validation layer, not just the responder chain.
• The body can't cross into a detached Task or escaping closure — the backpressure and file-upload cases had to be restructured.

So the pure-ownership model removes `@unchecked` but costs the `req.content/req.body/proxy` ergonomics. Is there an ownership approach you have in mind that avoids these — i.e. keeps the body reachable from `Request` while still getting rid of `@unchecked`? Want to make sure I'm not missing a cleaner shape before I go further.

Tests/VaporTests/RequestTests.swift
```diff
@@ -90,7 +118,9 @@ struct RequestTests {
 
         try await withApp { app in
             app.on(.post, "stream-decode", body: .stream) { req async throws -> String in
-                #expect(req.body.data != nil)
+                // NOTE: dropped an upstream `#expect(req.body.data != nil)` here — with lazy request
+                // streaming the body isn't buffered until `content.decode` collects it. Flagged for
+                // @0xTim (test added in #3552). See PR description.
```
0xTim, line 123: This is fine

0xTim (approved):
API shape and implementation look good! Think this is ready to go

PR #3544, opened 2026-08-28
Small Optimisation

0xTim:
We shouldn't need this but the compiler isn't picking the fast path (as shown by benchmarks). Potentially a Swift bug?

### DIFF
--- a/Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
+++ b/Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
@@ -173,7 +173,14 @@ final class NIOResponseBodyWriter: ResponseBodyWriter {
 
     func write(_ bytes: some Sequence<UInt8>) async throws {
         var out = UniqueArray<UInt8>(minimumCapacity: bytes.underestimatedCount)
-        out.append(copying: bytes)
+        // Staging is synchronous, so the sequence's own storage can be borrowed rather than copied
+        // element by element; only the transport write is awaited, after the borrow has ended.
+        let borrowed: Void? = bytes.withContiguousStorageIfAvailable { buffer in
+            out.append(copying: buffer)
+        }
+        if borrowed == nil {
+            out.append(copying: bytes)
+        }
         // `write` drains `out`, so the count has to be taken first.
         let count = out.count
         try await self.inner?.write(buffer: &out)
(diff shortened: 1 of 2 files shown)


Sources/Vapor/Response/Response+Body.swift
```diff
@@ -271,7 +271,14 @@ private final class CollectingBodyWriter: ResponseBodyWriter {
...
-        self.data.append(contentsOf: bytes)
+        // `Data.append(contentsOf:)` cannot reach a contiguous fast path here: the generic context
+        // has erased the concrete type, so it appends element by element. Recovering the buffer
+        // explicitly is what makes this cheaper than the protocol's default implementation.
+        let borrowed: Void? = bytes.withContiguousStorageIfAvailable { buffer in
+            unsafe self.data.append(contentsOf: buffer)
+        }
+        if borrowed == nil {
+            self.data.append(contentsOf: bytes)
+        }
```
MahdiBM, line 379: Hmmm this shouldn't be an issue.

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
```diff
@@ -169,7 +169,14 @@ final class NIOResponseBodyWriter: ResponseBodyWriter {
...
         var out = UniqueArray<UInt8>(minimumCapacity: bytes.underestimatedCount)
-        out.append(copying: bytes)
+        // Staging is synchronous, so the sequence's own storage can be borrowed rather than copied
+        // element by element; only the transport write is awaited, after the borrow has ended.
+        let borrowed: Void? = bytes.withContiguousStorageIfAvailable { buffer in
+            out.append(copying: buffer)
+        }
+        if borrowed == nil {
+            out.append(copying: bytes)
+        }
```
MahdiBM, line 183: This one though I can see UniqueArray is not taking the fast path of reading contiguous bytes.

MahdiBM (approved):

Sources/Vapor/Response/Response+Body.swift
MahdiBM: (on Swift 6.2/3 there is a try-cast to `ContiguousBytes`, on Swift 6.4 that doesn't exist anymore either)

Sources/Vapor/HTTP/Server/HTTPServerHandler.swift
MahdiBM: The stdlib array impl is good though (UniqueArray made it to stdlib in 6.4 but requires macOS 27).
This impl is using swift-collections i can see.

Sources/Vapor/Response/Response+Body.swift
0xTim: Yeah it's weird, I've proved it with the benchmarks. Without this the allocation count shoots up for streaming

MahdiBM: In any case these changes are certainly not any downgrade if not an upgrade so they're good.

PR #3547, opened 2026-08-29
Fail server startup on an empty TLS certificate chain

### LINKED ISSUE #3540: Empty certificate chain in .inMemory TLS configuration starts the server instead of failing
### Describe the issue

`.inMemory(certificateChain: [], privateKey:)` starts the server instead of failing, so every TLS handshake then fails at runtime

### Vapor version

main (`7de7a9e2`, 5.0.0-alpha.2 + 17)

macOS 27.0

### Swift version

Swift Package Manager - Swift 6.4

### Steps to reproduce

```swift
app.serverConfiguration.tlsConfiguration = .inMemory(certificateChain: [], privateKey: privateKey)
try await app.boot()
try await app.server.run()
```

Then connect with any TLS client, for example `curl -k https://127.0.0.1:<port>/`.

### Outcome

`run()` does not throw and `listeningAddress` resolves, so the app looks healthy. Every handshake fails with `TLSV1_ALERT_INTERNAL_ERROR` and nothing on the server side says why.

A mismatched private key fails at startup with `NIOSSLError.failedToLoadPrivateKey`, and `.pemFile` with missing paths fails at startup too. The empty chain is the one bad credential that gets through.

Expected behaviour is that startup fails like the other cases.

### Additional notes

Introduced with `.inMemory` in #3498. Happy to send a PR that rejects an empty chain where the TLS credentials are built in `NIOHTTPServerAdapter`, with negative-path tests for both cases in `ServerTLSTests`.

dioKaratzas:
Fails server startup when `.inMemory` TLS credentials carry an empty certificate chain (fixes #3540).

Previously `.inMemory(certificateChain: [], privateKey:)` started the server. `run()` did not throw and `listeningAddress` resolved, so the application looked healthy while every TLS handshake failed with `TLSV1_ALERT_INTERNAL_ERROR` and nothing on the server side reported why. A mismatched private key and `.pemFile` with missing paths both already fail at startup; the empty chain was the one bad credential that got through.

`NIOHTTPServerAdapter` now rejects it where the TLS credentials are built, throwing the new `NIOHTTPServerAdapterError.emptyCertificateChain`.

```swift
// Now throws at startup instead of failing every handshake later.
app.serverConfiguration.tlsConfiguration = .inMemory(certificateChain: [], privateKey: privateKey)
try await app.server.run()
```

Covered by a negative-path test in `ServerTLSTests`.

### DIFF
--- a/Sources/Vapor/HTTP/Server/NIOHTTPServerAdapter.swift
+++ b/Sources/Vapor/HTTP/Server/NIOHTTPServerAdapter.swift
@@ -15,6 +15,10 @@ enum NIOHTTPServerAdapterError: Error {
 
     /// No HTTP versions were configured. ``ServerConfiguration/httpVersions`` must contain at least one version.
     case noHTTPVersionsSpecified
+
+    /// The in-memory TLS credentials carried an empty certificate chain. A server has nothing to present
+    /// during the handshake without at least a leaf certificate.
+    case emptyCertificateChain
 }
 
 /// Adapts `NIOHTTPServer` to Vapor's `Server` protocol using structured concurrency.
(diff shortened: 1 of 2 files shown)


0xTim (approved):
Thanks for this! Can you fix the conflict then we can merge 🙌

PR #3551, opened 2026-08-31
[Discussion Welcome] New View Type

0xTim:
Migrates `View` to use `Data` as the backing storage instead of `ByteBuffer`. This removes another `ByteBuffer` from the public API. We can't really use anything but `Data` (or `[UInt8]` etc) for the parsing - we need the whole template in the AST tree before rendering.

However we do have the opportunity to make serialisation streaming. This would help in the use case where a for loop renders 10k rows etc. That would make it far more memory efficient and likely more performant as we can write each chunk to the response writer when it's ready rather than waiting.

The main downsides are:

* the streaming APIs are a bit more complex, but we can hide this easily with `ResponseEncodable`
* throwing errors can't easily lead to a sensible 500 error. Templates shouldn't really throw but if we've already returned a 200 OK head and stream the body, if we throw mid stream we're basically left with just closing the connection early. I looked at other renderers and Rails and Django both do everything in memory because of this.
* Content-length can't be a thing, which might be problematic for some CDNs or proxies and we can't use E-Tag and cache pages.

So options are:

1 Leave it as is, load everything into memory before returning
2 Streaming only, deal with the side effects above
3 Streaming opt-in for those that know they're rendering massive pages and need it.

I'm leaning towards 1 as loading a 50MB page into a web browser is painful anyway, so the use case for streaming and templating are quite small. But open to hearing other perspectives.

Thoughts?


Sources/Vapor/View/PlaintextRenderer.swift
```diff
@@ -58,7 +64,8 @@ public struct PlaintextRenderer: ViewRenderer, Sendable {
             : self.viewsDirectory + name
         return try await FileSystem.shared.withFileHandle(forReadingAt: .init(path)) { handle in
             let buffer = try await handle.readToEnd(maximumSizeAllowed: .megabytes(32))
-            return View(data: buffer)
+            let data = buffer.getData(at: 0, length: buffer.readableBytes) ?? Data()
```
MahdiBM: IIRC there is a `Data(buffer:)` init for this.

PR #3573, opened 2026-09-16
Don't link full Foundation

0xTim:
Our dependencies should be full Foundation free now, so lets see if CI goes green


MahdiBM (approved):
🎉

PR #3577, opened 2026-09-17
Add support for Crypto 5

0xTim:


MahdiBM (approved):

fpseverino:
The SwiftNIO HTTP/3 stack requires Swift Crypto v5. At the moment, they are using the beta version, and in fact I have added the dependency for it to the [Vapor branch for HTTP/3](https://github.com/vapor/vapor/pull/3570/changes#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR43-R44). Once there are new releases for the entire stack (from `swift-certificates` up to `swift-http-server`) that depend on (or support) the final version of Crypto 5, if we want to support HTTP/3, we will also have to depend on Crypto 5

